PR #12, opened 2019-08-30
Feature/incentive calculation

hamzajaved:


containers/Incentives/IncentiveCalculator.js
```diff
@@ -0,0 +1,190 @@
...
+        const allProducts = data.allProducts.allProducts.nodes;
+        let totalCarbonIncentive = 0;
+        let incentiveSegments = [];
+
+        productsReported.forEach((product) => {
+           // Get bm/et details for the product from the Products table
+           const productDetails = allProducts.filter((p) => p.name === product.processingUnit);
+           const productQuantity = parseFloat(product.quantity)
+           const attributableFuelPercentage = parseFloat(product.attributableFuelPercentage)
+           const benchmark = productDetails[0].benchmarksByProductId.nodes[0].benchmark;
```
dleard: @hamzajaved This may be more of an issue with how benchmarks are created / managed. But getting a Product's benchmark nodes and always selecting nodes[0] doesn't necessarily get the current benchmark for that Product.

example: a benchmark is created 1999-2001 it is node[0], a new benchmark is created 2001-2003 it is node[1] etc
Should we add a TODO here to document this issue?

hamzajaved: Yes, I believe this is a relic from before we added dates. will add a todo

PR #19, opened 2019-09-10
Reformat all the code using xo and fix a few linter errors

matthieu-foucault:
- `yarn lint` runs the linter.
- `yarn format` runs the linter and reformats the code.
- Added a check for the code style (not the linter) in CircleCI. If `yarn format` introduces some changes, then the job fails. If there are some linter errors  (e.g. unused var) but the code formatting is correct, then it succeeds. We can make this stricter when we fix all our errors.


wenzowski:
SonarCloud is failing because of `alert()` calls @hamzajaved can fix after this is merged

wenzowski:
Need help debugging the `build_app` error? Not sure why router isn't defined offhand

PR #22, opened 2019-09-18
update application status

dleard:


app/containers/Applications/ApplicationStatus.js
```diff
@@ -0,0 +1,106 @@
...
+        applicationStatusCondition: {
+          formResultId
+        }
+      });
+
+      this.setState({
+        status: queryReturn.allApplicationStatuses.nodes[0].applicationStatus,
+        displayStatus: queryReturn.allApplicationStatuses.nodes[0].applicationStatus.replace(
+          /^\w/,
+          char => char.toUpperCase()
```
hamzajaved: Let's use text-transform: capitalize for this

hamzajaved: Lets use css text-transform: capitalize for this.

PR #23, opened 2019-09-24
asdf doesn't seem to be recursing the file tree

wenzowski:


dleard:
my .tool-versions is working properly. I think I had the same problem and possibly fixed it by re-starting VSCode.

wenzowski:
I had an old version of [`asdf`](https://asdf-vm.com).

Upgrading solved it:
```bash
asdf update
asdf plugin-update --all
```

PR #25, opened 2019-09-25
Add storybook

dleard:


app/.storybook/addons.js
```diff
@@ -0,0 +1 @@
+import '@storybook/addon-a11y/register';
```
matthieu-foucault: needs a newline at the end of the file

PR #33, opened 2019-09-26
added jest version change to package.json

dleard:
I did yarn upgrade in the last PR but forgot to make the change to package.json...


dleard:
nevermind!

PR #47, opened 2019-09-30
wip: example use of singleton environment and fragments

wenzowski:
this still needs a bunch of work but hopefully helps as an example of how to use fragments

the $RELAY_ENDPOINT config will need to be passed to the client side for this to ever be deployable


wenzowski:
mutations will also probably need some `@connection` https://www.prisma.io/blog/relay-moderns-connection-directive-1ecd8322f5c8

wenzowski:
should probably pull the layouts example from #24 into this

wenzowski:
as much of this as was useful was implemented by @dleard in the Great Relay Refactor of 2019

PR #49, opened 2019-10-02
add created_at and updated_at triggers

wenzowski:
this commit adds the `ggircs_portal.update_timestamps()` function and installs it as a trigger on all tables


wenzowski:
@dleard @hamzajaved I'll rebase this onto `develop` once you two have got that passing :)

PR #54, opened 2019-10-03
(WIP) Relay refactor

dleard:


app/components/Header.js
```diff
@@ -64,9 +64,10 @@ const Header = () => {
                 </a>
                 <a
                   style={{marginLeft: '20px'}}
-                  target="_blank"
-                  rel="noopener noreferrer"
-                  href="http://localhost:3000/question/66"
+                  // (Dylan) Why these next 2 lines?
+                  // target="_blank"
+                  // rel="noopener noreferrer"
```
wenzowski: @dleard yeah, you don't need both. You also only need to strip the opener property on a cross-origin request (like the original to Metabase) so if you're opening the same app it's fine to just `target="_blank"` 
https://developers.google.com/web/tools/lighthouse/audits/noopener

app/components/Header.js
```diff
@@ -64,9 +64,10 @@ const Header = () => {
...
+                  href="/admin"
```
wenzowski: ...and now that this goes to Admin, I don't see why you'd open a new window anyway. Looks good :)

app/pages/products.js
```diff
@@ -1,15 +1,25 @@
 import React, {Component} from 'react';
-import ProductList from '../components/Products/ProductList';
+import {graphql} from 'react-relay';
+import ProductListContainer from '../containers/Products/ProductListContainer';
+
+// (Dylan) Why do we have this page? Is it for people that can't change the benchmarks/products to view them?
```
wenzowski: aherm...@hamzajaved

PR #84, opened 2019-10-15
Storyshots

dleard:


wenzowski:
rebased on #83 for review

wenzowski:
@dleard rebased on `develop`

dleard:
@wenzowski I think this needs an ignore script update somewhere, the linter is touching the .storyshots file extensions. I just haven't gotten around to figuring out where

wenzowski:
Eventually, it would be really nice if we could get the entire Storybook layer into its own folder and decoupled from our app.

PR #86, opened 2019-10-16
Add CircleCI target to check that graphql schema is in sync

matthieu-foucault:


matthieu-foucault:
@wenzowski Unless there's a CircleCI feature that I can't find. Looks like we'll need to build a docker image that has both `sqitch` and `node` for this to work.

wenzowski:
There's work in #89 that installs all `.tool-versions` via asdf as we would in development. Maybe we could use the right perl running as a base image instead of using the node base image and rely on the asdf code to load (& cache) everything else we need?

wenzowski:
this should now be fully included in #89

PR #89, opened 2019-10-18
sample pre-commit config

wenzowski:
if this seems valuable to others & worth pursuing I'll add `pre-commit run --all-files` so CI ensures these checks run for all, even when the user doesn't have pre-commit installed


wenzowski:
I'm having to `source` the `~/.asdf/asdf.sh` for each command run, and I'm not sure why that's the case given I've appended a line to `~/.bashrc` that should be loading asdf...

wenzowski:
this probably needs a bit of trial-and-error committing squashed to ensure the history is understandable

wenzowski:
rebased on develop

wenzowski:
rebased on develop

documentation/developer-guide.md
```diff
@@ -18,10 +18,23 @@ The `linter-xo-2` VSCode add XO support. The following VSCode settings will enab
...
+## Pre-commit hooks
+
+The Yelp [pre-commit](https://pre-commit.com) framework is configured for this project
+to automatically dispatch multi-language pre-commit hooks. Wherever possible,
+non-destructive linting steps will automatically fix common errors in staged code
+either when explicitly running `pre-commit` or automatically via the
+[git hook](https://git-scm.com/docs/githooks).
+
+Installation:
+- `asdf install` (grab the pything version specified in `.tool-versions`...but any python should do)
```
matthieu-foucault:
```suggestion
- `asdf install` (grab the python version specified in `.tool-versions`...but any python should do)
```

.circleci/config.yml
```diff
@@ -4,59 +4,195 @@ orbs:
...
+            yarn format || true # TODO: need to clean up linter errors
       - run:
           name: Check if there is some unformatted code
           command: git diff --exit-code || (echo 'These files are not formatted correctly. Did you forget to \"yarn format\"?' && false)
+      - run:
+          name: Verify style with pre-commit
+          command: |
+            source ~/.asdf/asdf.sh
+            asdf reshim
+            pre-commit run --all-files
```
matthieu-foucault: This will fail until we fix all the linter errors. There's only 25 of them, might as well fix them now?

wenzowski: Right, of course! Yes, I'll fix...

.bin/format.sh
```diff
@@ -2,5 +2,6 @@
 
 (
 pushd app || exit 1
-yarn format "$@"
+files=("$@")
+yarn format "${files[@]/#app\//}" # remove "app/" from the file names
```
wenzowski: this'll work for now, but I think what we really want is to resolve these to be absolute paths

.circleci/config.yml
wenzowski: Looks like there are a few that will need a bit of a refactor to fix. Can we do that in a separate PR? Trying to avoid merge conflicts.

wenzowski:
@matthieu-foucault thanks for catching the linter path stuff...how would you feel about getting this merged and addressing enforcing linter in a separate PR?

matthieu-foucault:
@wenzowski Sounds good. I'm playing with typescript, and it looks like this will add a lot more linter errors :smile:

PR #95, opened 2019-10-22
Application wizard now pulls data from SWRS schema

matthieu-foucault:
- added foreign key from application to facility
- created `application_swrs*` functions that add computed fields to the graphql `Application` class, allowing to retrieve the swrs data associated with a given application
- In the user dashboard, clicking on `Apply` creates a new application and routes the user to the `ciip-application` page
- the `ciip-application` page retrieves the `applicationId` and the `formId` from the querystring
- the `ApplicationWizardStep` component initializes the survey with the swrs data
- added a `SurveyWrapper` class around SurveyJS

**Regression to be fixed in the next PR**: the `applicationStatus` is not created. Need to add a stored procedure to create the application, and the `WizardConfirmationPage` needs to add a new status when the user submits the application.


app/containers/Applications/ApplicationWizard.js
```diff
@@ -1,38 +1,82 @@
-import React, {useState} from 'react';
+import {format} from 'url';
+import React, {useEffect} from 'react';
+import {useRouter} from 'next/router';
 import {graphql, createFragmentContainer} from 'react-relay';
 import ApplicationWizardStep from './ApplicationWizardStep';
 import ApplicationWizardConfirmation from './ApplicationWizardConfirmation';
 
+const setRouterQueryParam = (router, key, value, replace = false) => {
+  const newUrl = format({
```
dleard: no need to use format here. pass the object to router.push

app/containers/Organisations/Facility.js
```diff
@@ -1,15 +1,75 @@
 import React from 'react';
-import {graphql, createFragmentContainer} from 'react-relay';
+import {graphql, commitMutation, createFragmentContainer} from 'react-relay';
 import {Button} from 'react-bootstrap';
+import {useRouter} from 'next/router';
+import Link from 'next/link';
+
+export const FacilityComponent = ({relay, facility}) => {
```
dleard: Will need a stored procedure in the future for chaining an application status

PR #106, opened 2019-10-25
Use pre-commit to verify commit message format

wenzowski:
[git_commit_2x](https://xkcd.com/1296/)

## Solution: use pre-commit to verify commit message format
**Developer action is required to enable this feature locally.**

The `commit-msg` hook must be installed to ensure commit messages
conform to community standards:

```bash
pre-commit install --hook-type commit-msg
```

The project [pre-commit] framework configuration
has been updated to include hooks for [gitlint] and a [`.gitlint`]
configuration file has been added to control its [rules].

[pre-commit]: https://pre-commit.com/
[gitlint]: http://jorisroovers.com/gitlint/
[`.gitlint`]: http://jorisroovers.com/gitlint/configuration/
[rules]: http://jorisroovers.com/gitlint/rules/

### DIFF
--- a/.czrc
+++ b/.czrc
@@ -1,0 +1,3 @@
+{
+  "path": "app/node_modules/cz-conventional-changelog"
+}
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,96 @@
+# This file controls the behaviour of the `gitlint` tool and by extension the commit-msg pre-commit hook.
+# @see http://jorisroovers.com/gitlint/configuration/
+
+# All these sections are optional, edit this file as you like.
+# [general]
+# Ignore certain rules, you can reference them by their id or by their full name
+# ignore=title-trailing-punctuation, T3
+
+# verbosity should be a value between 1 and 3, the commandline -v flags take precedence over this
+# verbosity = 2
+
+# By default gitlint will ignore merge commits. Set to 'false' to disable.
+# ignore-merge-commits=true
+
+# By default gitlint will ignore fixup commits. Set to 'false' to disable.
+# ignore-fixup-commits=true
+
+# By default gitlint will ignore squash commits. Set to 'false' to disable.
+# ignore-squash-commits=true
+
+# Ignore any data send to gitlint via stdin
+# ignore-stdin=true
+
+# Enable debug mode (prints more output). Disabled by default.
+# debug=true
+
+# Enable community contributed rules
+# See http://jorisroovers.github.io/gitlint/contrib_rules for details
+# contrib=contrib-title-conventional-commits,CC1
+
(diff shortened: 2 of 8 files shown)


wenzowski:
This probably needs [CI config](http://jorisroovers.com/gitlint/#using-gitlint-in-a-ci-environment) before merging...

matthieu-foucault:
This should be combined with using committizen to write our commit messages

wenzowski:
This might tie in with a [good contributing document](https://github.com/deis/deis/blob/master/CONTRIBUTING.md) too

wenzowski:
contributing doc underway in https://github.com/bcgov/cas-ciip-portal/pull/117

wenzowski:
@matthieu-foucault either way we're implementing https://www.conventionalcommits.org/ right? I'm not super familiar with [cz-cli](http://commitizen.github.io/cz-cli/) but I'm happy to take a stab at that unless you have opinions about config setup?

matthieu-foucault:
> @matthieu-foucault either way we're implementing https://www.conventionalcommits.org/ right? I'm not super familiar with [cz-cli](http://commitizen.github.io/cz-cli/) but I'm happy to take a stab at that unless you have opinions about config setup?

Yes we are. `cz-cli` isn't technically required, it just formats the commit messages for you.

wenzowski:
@matthieu-foucault I'd like to make sure these comments are cleaned up before I fix this branch. Lmk if we need to sync on that and tyvm for pitching in here.

@blueQcloud something we should probably put in our Contributing document: the team integrates only on the `develop` branch. This means only the person who created a `feature/*` branch should update it, unless otherwise instructed by the original author.

matthieu-foucault:
@wenzowski You're good to go. I took care of the comments

matthieu-foucault:
The repo still has to be made [comittizen-friendly](https://github.com/commitizen/cz-cli#making-your-repo-commitizen-friendly) if we want to use git-cz insead of git commit

wenzowski:
had to use a `.czrc` file to adapt the ugly monorepo, but `git-cz` should work

wenzowski:
> instruct all devs to use the pre-commit and/or comittizen
there's currently divergence between the python config (gitlint) and the node config (git-cz) which has me wondering if we shouldn't settle on either comittizen or pre-commit as a process (and not both)

matthieu-foucault:
Perhaps we should use https://github.com/conventional-changelog/commitlint instead of gitlint

wenzowski:
I'll look into that then

matthieu-foucault:
Continuing work in #539

matthieu-foucault:
This actually looks closer to what I was going to do than I thought, not sure why it wasn't merged in the first place.

PR #109, opened 2019-10-25
Re-implement application status

dleard:
Added a stored procedure to cover createApplication -> createApplicationStatus mutation chain, renders a status in facility view && conditionally renders the apply/resume application button

### DIFF
--- a/app/containers/Organisations/Facility.js
+++ b/app/containers/Organisations/Facility.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import {graphql, createFragmentContainer} from 'react-relay';
-import {Button, Card, ListGroup, ListGroupItem} from 'react-bootstrap';
+import {Button, Badge, Card, ListGroup, ListGroupItem} from 'react-bootstrap';
 import {useRouter} from 'next/router';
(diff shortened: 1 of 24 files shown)


app/containers/Organisations/Facility.js
```diff
@@ -102,6 +137,13 @@ export default createFragmentContainer(FacilityComponent, {
         edges {
           node {
             id
+            applicationStatusesByApplicationId {
```
matthieu-foucault: order by creation date desc

app/containers/Organisations/Organisations.js
```diff
@@ -4,7 +4,7 @@ import {Table, Dropdown, Button} from 'react-bootstrap';
 import Organisation from './Organisation';
 import UserOrganisation from './UserOrganisation';
 
-const Organisations = props => {
+export const Organisations = props => {
```
matthieu-foucault: no named export with the same name as the file

app/containers/Organisations/Facility.js
dleard: good idea

PR #118, opened 2019-10-28
bump relay to v7

wenzowski:
closes a whole bunch of dependabot PRs if stable


app/pages/form-builder.js
```diff
@@ -1,6 +1,6 @@
 import React, {Component} from 'react';
 import {Container, Row, Col} from 'react-bootstrap';
-import Header from '../components/Header';
+// Import Header from '../components/Header';
```
matthieu-foucault: Remove this page? It's not used, and broken

.circleci/config.yml
```diff
@@ -194,6 +194,19 @@ jobs:
...
 
+  compile:
+    machine:
+      image: ubuntu-1604:201903-01
+    working_directory: ~/cas-ciip-portal/app
+    steps:
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: Audit Dependencies
```
matthieu-foucault:
```suggestion
          name: Make a production build
```

.circleci/config.yml
```diff
@@ -194,6 +194,19 @@ jobs:
...
+          command: |
+            source ~/.asdf/asdf.sh
+            yarn build # ensure next.js can compile a production build
```
matthieu-foucault:
```suggestion
            yarn build:relay
            yarn build:next # ensure next.js can compile a production build
```
This is what we're doing on pathfinder. See the s2i assemble script

matthieu-foucault:
You'll need to update the snapshots, looks like relay 7 adds more forwardrefs

.circleci/config.yml
wenzowski: yessir

.circleci/config.yml
wenzowski: good point, the schema is checked in a separate, isolated step

app/pages/form-builder.js
wenzowski: done

wenzowski:
rebased on develop

PR #121, opened 2019-10-29
User Dashboard design update

hamzajaved:
based on an old branch of #118; needs those commits rebased out of history and to be based off `develop` instead


matthieu-foucault:
My comment still stands. 1f6522f Updates the `.pipeline` submodule. Looks like you need an interactive rebase that drops that commit.

hamzajaved:
> My comment still stands. [1f6522f](https://github.com/bcgov/cas-ciip-portal/commit/1f6522f0c410d96f94d897716a9a1ed6ed805925) Updates the `.pipeline` submodule. Looks like you need an interactive rebase that drops that commit.

Pushed an update - hopefully, it should be fixed. Otherwise, I might bother you for a quick pair.

PR #123, opened 2019-10-29
fix: Add isomorphic-fetch as a dependency

matthieu-foucault:
It is needed by the relay server-side rendering

### DIFF
--- a/app/lib/relay-environment/index.js
+++ b/app/lib/relay-environment/index.js
@@ -1,4 +1,4 @@
-// TODO(Dylan): Do we need this import?
+// The relay SSR uses `fetch` under the hood to request data.
 import 'isomorphic-fetch';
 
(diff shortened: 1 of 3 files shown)


wenzowski:
I suspect the intent was actually to use `isomorphic-unfetch` instead, since that's already included as a dependency and lighter weight. Some inline-documentation is necessary here either way since we're global polyfilling rather than named-importing.

matthieu-foucault:
- `isomorphic-fetch` is already in our transitive dependencies.
- `isomorphic-unfetch` doesn't work as it defines `window.fetch` (and `window`), and we look at whether `window` is defined to switch between client and server. 
- Relay also has options to use graphql directly instead of fetch, not sure what would be the benefit of that

wenzowski:
I came to the same conclusion...but you beat me to it

PR #124, opened 2019-10-29
fix (relay-environment): do not throw an error with undefined relayData

matthieu-foucault:
This removes the "cannot get function 'find' of undefined" error that flashes before rendering a page

### DIFF
--- a/app/lib/relay-environment/server.js
+++ b/app/lib/relay-environment/server.js
@@ -19,7 +19,7 @@ export default {
           urlMiddleware({
             // TODO: set $RELAY_ENDPOINT
             // url: req => process.env.RELAY_ENDPOINT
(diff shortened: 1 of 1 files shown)


app/lib/relay-environment/server.js
```diff
@@ -32,9 +32,12 @@ export default {
...
       store,
-      network: Network.create(
-        () => relayData.find(([dataKey]) => dataKey === key)[1]
-      )
+      network: Network.create(() => {
+        if (!relayData) return;
+        const data = relayData.find(([dataKey]) => dataKey === key);
+        if (!data) return;
+        return data[1];
+      })
```
wenzowski: excellent :)

Now the question is why we don't have any `relayData`...

PR #134, opened 2019-10-31
Feature/faciliy no overfetch

blueQcloud:


schema/deploy/function_application_application_status.sql
```diff
@@ -0,0 +1,23 @@
...
+begin;
+
+create or replace function ggircs_portal.application_application_status(application ggircs_portal.application)
+    returns ggircs_portal.application_status
+as
+$function$
+declare
+begin
+    return (
+        select row(_application_status.*)
```
matthieu-foucault: Do you need `row()` ?

blueQcloud: Yes, because the query must return a single column

PR #147, opened 2019-11-05
Remove relay SSR

dleard:
This PR has removed SSR from relay, which fixes the react navigation tabs.
Relay SSR appears to be somewhat broken at the moment with active Issues and PRs:

https://github.com/facebook/relay/issues/1881
https://github.com/facebook/relay/pull/2883

### DIFF
--- a/app/lib/relay-environment/server.ts
+++ b/app/lib/relay-environment/server.ts
@@ -2,42 +2,35 @@ import {
   RelayNetworkLayer,
   urlMiddleware
 } from 'react-relay-network-modern/node8';
-import RelaySSR from 'react-relay-network-modern-ssr/node8/server';
 import {Network, Environment, RecordSource, Store} from 'relay-runtime';
 
 export default {
   initEnvironment: () => {
     const source = new RecordSource();
     const store = new Store(source);
-    const relaySSR = new RelaySSR();
 
     return {
-      relaySSR,
       environment: new Environment({
         store,
         // @ts-ignore
         network: new RelayNetworkLayer([
           urlMiddleware({
-            // TODO: set $RELAY_ENDPOINT
+            // TODOx: set $RELAY_ENDPOINT
             // url: req => process.env.RELAY_ENDPOINT
             url: _ => 'http://localhost:3004/graphql'
-          }),
-          relaySSR.getMiddleware()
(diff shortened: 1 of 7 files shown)


app/lib/relay-environment/server.ts
```diff
@@ -2,42 +2,35 @@ import {
...
             // url: req => process.env.RELAY_ENDPOINT
             url: _ => 'http://localhost:3004/graphql'
-          }),
-          relaySSR.getMiddleware()
+          })
         ])
       })
     };
   },
-  createEnvironment: (relayData, key) => {
```
matthieu-foucault: we're not passing any `relayData` anymore when creating the server environment

app/pages/_app.tsx
```diff
@@ -47,19 +49,20 @@ export default class App extends NextApp<AppProps> {
...
+      pageProps: {variables = {}}
     } = this.props;
     const environment = createEnvironment(
-      relayData,
       JSON.stringify({
         queryID: Component.query
           ? getRequest(Component.query).params.name
           : undefined,
-        variables
+        variables: {...variables, ...router.query}
```
matthieu-foucault: Should this parameter be removed as well?

app/pages/_app.tsx
```diff
@@ -47,19 +49,20 @@ export default class App extends NextApp<AppProps> {
...
       })
     );
     return (
       <ErrorBoundary>
+        {/*
+          // @ts-ignore */}
```
matthieu-foucault: Do we need this?

app/pages/_app.tsx
dleard: need this one

app/lib/relay-environment/server.ts
dleard: right! I've removed it

app/pages/_app.tsx
dleard: nope!

app/pages/_app.tsx
dleard: It's used in the client.ts fetchQuery function to determine whether the same query with the same variables is in the cache and to just use the cache in that case. This is also why mutations clear the cache.

PR #152, opened 2019-11-07
Migrate heat and gas form to json-schema form

dleard:
Changes json file to work with json-schema
Adds props & custom submit button to JsonSchemaForm

### DIFF
--- a/schema/data/portal/form_json/electricity_and_heat.json
+++ b/schema/data/portal/form_json/electricity_and_heat.json
@@ -5,48 +5,15 @@
     "properties": {
       "electricity": {
         "title": "Electricity",
-        "type": "object",
-        "properties": {
-          "purchased": {
-            "type": "number",
-            "title": "Purchased"
-          },
-          "generatedOnSite": {
-            "type": "number",
-            "title": "Generated On-Site"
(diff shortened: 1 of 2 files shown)


schema/data/portal/form_json/electricity_and_heat.json
```diff
@@ -1,113 +1,145 @@
...
+              }
+            ]
+          }
         }
-      ]
+      },
+      "heat": {
+        "title": "Heat",
+        "type": "object",
+        "properties": {
```
matthieu-foucault: Can you use a definition (see address in the admin page) instead of repeating the fields?

schema/data/portal/form_json/electricity_and_heat.json
```diff
@@ -1,113 +1,145 @@
...
+        "sold",
+        "consumedOnSite",
+        "generatedOnSite",
+        "*"
+      ],
+      "purchased": {
+        "ui:col-md": 4
+      },
+      "generatedOnSite": {
+        "ui:col-md": 6
```
matthieu-foucault: md-6 is the default, so you can remove those

PR #172, opened 2019-11-12
added organisations requests page with filtrable list of all appliations

blueQcloud:


app/containers/Admin/OrganisationRequestsTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+      <Table
+        striped
+        hover
+        style={{textAlign: 'center', border: '1px solid #f5f5f5'}}
+      >
+        <thead style={{backgroundColor: '#036', color: 'white'}}>
+          <tr>
+            <th>#</th>
+            <th>Name</th>
+            <th>Lastname</th>
```
matthieu-foucault: We should have the user email in this table. Name and last name are not enough to identify a user

app/containers/Admin/OrganisationRequestsTableRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {graphql, createFragmentContainer} from 'react-relay';
+import {OrganisationRequestsTableRow_userOrganisation} from '__generated__/OrganisationRequestsTableRow_userOrganisation.graphql';
+interface Props {
+  userOrganisation: OrganisationRequestsTableRow_userOrganisation;
+  key: string;
+}
+export const OrganisationRequestsTableRowComponent: React.FunctionComponent<
+  Props
+> = props => {
+  console.log(props);
```
matthieu-foucault:
```suggestion
```

app/containers/Organisations/UserOrganisation.tsx
```diff
@@ -11,10 +11,10 @@ export const UserOrganisationComponent = props => {
...
-    attention: 'warning',
+    rejected: 'warning',
     pending: 'info',
-    expired: 'danger',
-    active: 'success'
+    // Expired: 'danger',
+    approved: 'success'
   };
 
   /** TODO: INTENDED FOR DEMO ONLY. To be removed */
```
matthieu-foucault: To properly the new approval interface, while keeping this behavior for development, I suggest that we create two buttons: one that sets directly sets the status as active, only visible in development (`process.env.NODE_ENV !== 'production'`) and one that creates a pending application.

app/server/schema.graphql
```diff
@@ -6918,9 +6947,7 @@ input UserOrganisationCondition {
 input UserOrganisationInput {
   """foreign key that references the organisation table"""
   organisationId: Int!
-
-  """the status of user access to the organisation"""
-  status: String!
+  status: Statuses
```
matthieu-foucault: :heart_eyes:

schema/deploy/type_statuses.sql
```diff
@@ -0,0 +1,8 @@
+-- Deploy ggircs-portal:type_statuses to pg
+-- requires: schema_ggircs_portal
+
+begin;
+
+create type ggircs_portal.statuses as enum ('pending', 'approved', 'rejected');
```
matthieu-foucault: The name of this type should be more precise. `user_organisation_status` is better, as it tells us in which context the type is used

app/containers/Organisations/UserOrganisation.tsx
dleard: good idea

schema/deploy/type_statuses.sql
dleard: :+1:

PR #173, opened 2019-11-12
Update all dependencies assumed to be compatible

wenzowski:
This leaves four available dependencies unmerged as they have breaking
changes (indicated by semver):

```
Package                     Current Wanted Latest        Package Type    URL
bootstrap                   3.4.1   3.4.1  4.3.1         dependencies    https://getbootstrap.com/
eslint-config-xo-typescript 0.19.0  0.19.0 0.21.0        devDependencies https://github.com/xojs/eslint-config-xo-typescript#readme
nock                        9.6.1   9.6.1  11.7.0        devDependencies https://github.com/nock/nock#readme
react-jsonschema-form       1.8.0   1.8.0  2.0.0-alpha.1 dependencies    https://github.com/mozilla-services/react-jsonschema-form#readme
```
^ use `yarn outdated` to verify

This closes a bunch of dependabot PRs.

### DIFF
--- a/app/next.config.js
+++ b/app/next.config.js
@@ -1,8 +1,8 @@
-import path from 'path';
-import dotenv from 'dotenv';
+const path = require('path');
+const withCSS = require('@zeit/next-css');
+const Dotenv = require('dotenv-webpack');
+const dotenv = require('dotenv');
 dotenv.config();
-import Dotenv from 'dotenv-webpack';
-import withCSS from '@zeit/next-css';
 
 module.exports = withCSS({
   cssModules: true,
--- a/app/package.json
+++ b/app/package.json
@@ -171,7 +171,7 @@
     "relay-compiler-language-typescript": "^10.1.0",
     "relay-test-utils": "^7.0.0",
     "require-context.macro": "^1.2.2",
-    "typescript": "^3.6.4",
+    "typescript": "3.6.4",
     "webpack": "^4.41.0",
     "xo": "^0.25.0"
   },
--- a/app/yarn.lock
+++ b/app/yarn.lock
@@ -2,30 +2,36 @@
 # yarn lockfile v1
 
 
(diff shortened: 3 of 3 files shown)


wenzowski:
@matthieu-foucault this fails due to https://github.com/zeit/next.js/issues/5318

matthieu-foucault:
`'next.config.ts' is not supported. Please replace the file with 'next.config.js'.` Looks like I went too far with my types. Given that it was silently ignoring the ts config before the update, do we even need the next config?

matthieu-foucault:
Either way, renaming it to `js` will solve the issue

matthieu-foucault:
I'll update `eslint-config-xo-typescript` in a separate PR, The default config is now disabling some rules which I disabled as well, so I'll clean that up

wenzowski:
I think we want the next config for dotenv support? Would be nice to remove cssmodules support if we're not using them now–they're super buggy (as we discovered) and don't play nice with `styled-jsx` which we are definitely still using.

matthieu-foucault:
Maybe just rename the config for now then.

wenzowski:
@matthieu-foucault looks like there type errors caused by adding `config.next.js` https://circleci.com/gh/bcgov/cas-ciip-portal/4927

Are you good to take that or shall I?

matthieu-foucault:
I can take a look. Looks like next is ignoring our `@ts-ignore` annotations now...

wenzowski:
Looks like we were using an unsupported feature with the 
```tsx
{ /*
  // @ts-ignore something
  */ }
```
pattern, since there's still an [open issue](https://github.com/microsoft/TypeScript/issues/31147) about it

matthieu-foucault:
Sound good. I can see about fixing those errors and upgrading TypeScript in another PR.

PR #179, opened 2019-11-13
Throttle mutation calls

dleard:
Adds throttle to the edit user (/registration) page:
  - updates the local store on every keystroke
  - sends a commitMutation call after 1 second of handleChange() inactivity


dleard:
Closing this since it is way out of date. It is a good example, but should be reimplemented on the refactored code.

PR #200, opened 2019-11-18
Add footer links

dimak1:
Link styles were applying, as per this [issue](https://github.com/zeit/next.js/issues/610), so had to fix by adding :global(a).

### DIFF
--- a/app/components/Layout/Footer.tsx
+++ b/app/components/Layout/Footer.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import Link from 'next/link';
 
 const Footer = () => {
   return (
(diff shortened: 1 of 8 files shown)


dimak1:
> Can you add a page for each of the three links that just has the `DefaultLayout` and some `TODO` text?

dimak1:
Closed by mistake

dimak1:
@matthieu-foucault for these pages (Disclaimer, Privacy, Accessibility and Copyright) we don't need a session, right?

matthieu-foucault:
@dimak1 Yes, those pages do not need a session

dimak1:
@matthieu-foucault DefaultLayout requires session

matthieu-foucault:
Yes. It does need a query with a session, as it will change what is rendered in the header, but if you set `needsSession` and `needsUser` to `false`, it will not redirect the user if there is no session (see the index page for an example)

app/components/Layout/Footer.tsx
```diff
@@ -7,10 +8,16 @@ const Footer = () => {
         <div className="container">
           <ul>
             <li>
-              <a href=".">Home</a>
+              <Link href="/resources/disclaimer">Disclaimer</Link>
             </li>
             <li>
-              <a href=".">Contact Us</a>
```
matthieu-foucault: I don't think those should be removed. `Home` should be a link to `/`, and you can add another page for `Contact Us`

dimak1: Added links and Contact Us page.

PR #204, opened 2019-11-18
Fix ciip views

dleard:
fixed electricity & heat, production, fuel & applicaiton views after change to jsonSchema

### DIFF
--- a/schema/deploy/view_ciip_application.sql
+++ b/schema/deploy/view_ciip_application.sql
@@ -8,7 +8,7 @@ create view ggircs_portal.ciip_application as (
     with x as (
       select
         form_result.application_id as id,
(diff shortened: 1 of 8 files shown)


schema/deploy/view_ciip_application.sql
```diff
@@ -22,7 +22,7 @@ create view ggircs_portal.ciip_application as (
     y as (
       select
         form_result.application_id as id,
-        json_array_elements((form_result -> 'reportingOperationInformation')::json) as operator_data
+        (form_result -> 'reportingOperationInformation')::json as operator_data
```
matthieu-foucault: `reportingOperationInformation` is not a thing anymore

PR #212, opened 2019-11-20
fix(authentication): ensure malformed sessions are recoverable by logout

wenzowski:
there's currently no way to abort registration if we signed in with the wrong account; fixed here

### DIFF
--- a/app/components/Layout/Header.tsx
+++ b/app/components/Layout/Header.tsx
@@ -3,7 +3,7 @@ import {Row, Col, Form, Button} from 'react-bootstrap';
 import Link from 'next/link';
 import LoginButton from '../LoginButton';
 
-const HeaderLayout = ({isLoggedIn = false}) => (
+const HeaderLayout = ({isLoggedIn = false, isRegistered = false}) => (
   <header>
     <div className="container">
       <div className="banner">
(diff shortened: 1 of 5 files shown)


app/yarn.lock
```diff
@@ -9198,7 +9198,7 @@ lodash.zip@^4.2.0:
   resolved "https://registry.yarnpkg.com/lodash.zip/-/lodash.zip-4.2.0.tgz#ec6662e4896408ed4ab6c542a3990b72cc080020"
   integrity sha1-7GZi5IlkCO1KtsVCo5kLcswIACA=
 
-lodash@4.17.15, "lodash@>=4 <5", lodash@^4.13.1, lodash@^4.15.0, lodash@^4.17.10, lodash@^4.17.11, lodash@^4.17.12, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.4, lodash@^4.17.5:
+lodash@4.17.15, "lodash@>=4 <5", lodash@^4.11.1, lodash@^4.13.1, lodash@^4.15.0, lodash@^4.17.10, lodash@^4.17.11, lodash@^4.17.12, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.4, lodash@^4.17.5:
```
wenzowski: this shouldn't really be in this commit, but the lockfile in develop is out of sync and this doesn't affect actual resolution in any way

PR #205, opened 2019-11-19
test (form): use a test QueryRenderer and test the entire fragment tree

matthieu-foucault:

PR #214, opened 2019-11-20
Feature: Page to View Submitted Application

dimak1:


app/containers/Organisations/Facility.tsx
```diff
@@ -44,7 +44,11 @@ export const FacilityComponent: React.FunctionComponent<Props> = ({
   const applyButton = () => {
     if (!applicationId) {
       return (
-        <Button variant="primary" onClick={startApplication}>
+        <Button
+          variant="primary"
+          data-action="apply"
```
matthieu-foucault: I'm not convinced about adding a property for testing purposes. @wenzowski  any opinion on that?

dimak1: We can make it `data-test-` or `test-data-` and remove these attr when building prod version

dimak1:
Here is a thread about removing attributes for production build: https://stackoverflow.com/questions/43934210/remove-html-attribute-in-production-build
We can maybe use [babel-plugin-remove-attribute](https://www.npmjs.com/package/babel-plugin-remove-attribute).

dimak1:
Fixed conflicts.

app/pages/view-application.tsx
```diff
@@ -0,0 +1,59 @@
+import React, {Component} from 'react';
```
matthieu-foucault: Looks like this is now a duplicate of the `applications-details` page?

matthieu-foucault: Except for the "Back" button maybe. Which reminds me that we should probably use the `DefaultLayout`'s `showSubheader` prop instead

app/containers/Organisations/Facility.tsx
matthieu-foucault: If @wenzowski approves, then I would go for `test-data-action`. That should also be documented in our developer guide. I started a section on tests in #205

wenzowski: Let's please not add production-path code to support testing.

Why is the `data-action` necessary?  The button had an `onClick`...why can't we bind an event handler?

wenzowski:
> removing attributes for production build

This way lies pain and heisenbugs. By changing the codepath to produce a production build we're essentially testing a different artifact from the one that's actually deployable. Let's do our best to isolate our test constructs from the production code we're trying to test. Happy to discuss approaches to this

app/containers/Organisations/Facility.tsx
```diff
@@ -44,7 +44,11 @@ export const FacilityComponent: React.FunctionComponent<Props> = ({
...
+          onClick={startApplication}
```
wenzowski: style: an `onEvent` should always be bound to a `handleEvent` callback

app/tests/containers/Organisation/Facility.test.tsx
```diff
@@ -40,12 +40,13 @@ describe('Facility', () => {
...
 
-  it('should not render a Button if an applicationID exists and is not in draft status', async () => {
+  it('should render a View Submitted App button if an applicationID exists and is not in draft status', async () => {
     const r = shallow(
       <FacilityComponent relay={{environment: {}}} facility={facility} />
     );
     expect(r).toMatchSnapshot();
-    expect(r.exists('Button')).toEqual(false);
+    expect(r.exists('Button')).toEqual(true);
+    expect(r.find('Button').prop('data-action')).toBe('view');
```
wenzowski: if we bind and event handler instead, we can use enzyme to dispatch a synthetic event and test that code path

wenzowski: something along the lines of...
```
const button = r.find('Button');
button.simulate('click', new Event('click'));
```

app/containers/Organisations/Facility.tsx
dimak1: @wenzowski this is attribute for testing, not for action handling.

app/pages/view-application.tsx
dimak1: It is similar to Summary page (`ApplicationWizardConfirmation`) but doesn't have Continue button.

matthieu-foucault: The `ApplicationWizardConfirmation` got refactored into an `ApplicationDetails` container, which doesn't have that button. That's being rendered by the `application-details` page at the moment, so I'm not sure the `view-application` page is needed anymore

dimak1: Ok let me see if i can reuse `application-details` page instead.

app/containers/Organisations/Facility.tsx
dimak1: I only added `data-action` attribute here. Im happy to refactor onClick event id necessary.

app/tests/containers/Organisation/Facility.test.tsx
dimak1: Here I'm testing that the button is a `view` application button. I'm not testing events. Just to see that right button is showing up.

matthieu-foucault: I think that what @wenzowski  means is that instead of testing for the button text, or a `data-` attribute, you can test that clicking the button will trigger the correct handler (i.e. the "Start Application" button invokes `handleStartApplication`). In the case of the "view" button, I suppose you want to check the presence of a next.js `Link`

matthieu-foucault:
```
const spy = jest.spyOn(wrapper.instance(), 'handleStartApplication'); // replace function
wrapper.update(); // not sure this is needed
const button = r.find('Button');
button.simulate('click');
expect(spy).toBeCalled(); 
```

app/pages/view-application.tsx
dimak1: I checked, and `application-details` cannot be reused. It's a different page.

matthieu-foucault: Right, there's probably some naming issues here. Can you at least use the `ApplicationDetails` component instead of iterating on the `formResultsByApplicationId` (`ApplicationDetails` should be doing that for you)

app/containers/Organisations/Facility.tsx
wenzowski: @dimak1 what are you testing then, if not an action?

app/tests/containers/Organisation/Facility.test.tsx
wenzowski: Exactly...with the caveat that `button.simulate('click');` may need an event passed in as well when you're shallow rendering with Enzyme.

wenzowski: If the button is always contained in a `<Link><Button /><Link>` and that's where the click handler resides, maybe you're trying to assert that 
`r.find('Button').parent().is('Link')` and 
`r.find('Button').parent().prop('href').is({"pathname": "/view-application", "query": {"applicationId": "id"})`
...or something like that?

dimak1: I was trying to assert that View Submitted Application button exits for `pending` app, since we didnt have any button before. Let me add that and clean-up...

app/pages/view-application.tsx
dimak1: Yeap, I added `ApplicationDetails` component. Updated tests.

app/tests/containers/Organisation/Facility.test.tsx
dimak1: Updated Facility tests.

app/containers/Organisations/Facility.tsx
dimak1: Removed `data-test` attr.

app/containers/Organisations/Facility.tsx
dimak1: Testing dynamic button. If application is completed it should show `View` button, otherwise, `Resume`. Updated tests.

app/tests/pages/view-application.test.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import {shallow} from 'enzyme';
+import ViewApplication from '../../pages/view-application';
```
matthieu-foucault:
```suggestion
import ViewApplication from 'pages/view-application';
```
No need to add `..` anymore, you can write relative imports starting form the `app` directory now

app/tests/pages/view-application.test.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
matthieu-foucault: Can you move this test to the `tests/unit` folder?

app/pages/view-application.tsx
```diff
@@ -0,0 +1,49 @@
...
+import {Button} from 'react-bootstrap';
+import {graphql} from 'react-relay';
+import Link from 'next/link';
+import {viewApplicationQueryResponse} from 'viewApplicationQuery.graphql';
+import DefaultLayout from '../layouts/default-layout';
+import ApplicationDetails from '../containers/Applications/ApplicationDetailsContainer';
```
matthieu-foucault:
```suggestion
import ApplicationDetails from 'containers/Applications/ApplicationDetailsContainer';
```

app/pages/view-application.tsx
```diff
@@ -0,0 +1,49 @@
+import React, {Component} from 'react';
+import {Button} from 'react-bootstrap';
+import {graphql} from 'react-relay';
+import Link from 'next/link';
+import {viewApplicationQueryResponse} from 'viewApplicationQuery.graphql';
+import DefaultLayout from '../layouts/default-layout';
```
matthieu-foucault:
```suggestion
import DefaultLayout from 'layouts/default-layout';
```

app/tests/pages/view-application.test.tsx
dimak1: Updated.

app/pages/view-application.tsx
dimak1: Updated.

app/tests/pages/view-application.test.tsx
dimak1: Moved to unit test folder.

app/pages/view-application.tsx
dimak1: Updated.

PR #215, opened 2019-11-20
fix: enable rsjf's safeRenderCompletion hack to prevent cursor jumps

matthieu-foucault:
See: https://github.com/rjsf-team/react-jsonschema-form/issues/1197


matthieu-foucault:
> Seems like we are waiting on RJSF to fix something and this is a hack in the meantime?
> [rjsf-team/react-jsonschema-form#1197](https://github.com/rjsf-team/react-jsonschema-form/issues/1197)

Yup. This is only needed if the form contents are stored in the relay store.

PR #217, opened 2019-11-21
WIP: relay dedupe/debounce

matthieu-foucault:
DO NOT MERGE

This is making the batch middleware very unhappy, but it works.
This needs to be changed to handle debouncing before the batch middleware instead of modifying the request it creates


wenzowski:
This looks really cool and it's awesome that this works. In practice, this currently violates the `Disposable` interface exposed by [`commitMutation`](https://github.com/facebook/relay/blob/master/packages/relay-runtime/mutations/commitMutation.js#L80) since the [`Disposable type`](https://github.com/facebook/relay/blob/master/packages/relay-runtime/util/RelayRuntimeTypes.js#L24) doesn't allow registration of a continuation to inform the caller when the Disposable has been `dispose()`'d. As a result, we really want to split this into two parts: 
- a debounce/throttle-based "delay" middleware at the network layer that releases all queued mutations no more frequently than at a specific interval but never interferes with their being dispatched, and
- a singleton mixin for use in mutation modules that automatically preserves a reference to the last mutation and calls `dispose()` on it if it has not yet resolved via a `onCompleted` or `onError` continuation.
This way, each mutation is responsible for managing its own cancellation state and unnecessary mutations are given an opportunity to be cancelled by way of the delay introduced at the network level, without potentially interfering with any mutations that should not be `dispose()`'d of.

matthieu-foucault:
@wenzowski Looking at your comment above, I think that one small adjustment that needs to be made is that the network layer middleware does need to interfere with the mutations being dispatched, since `dispose()` will not cancel the network request.
As per the only docs I could find on it (https://relay.dev/docs/en/api-reference#return-value-6)
> Calling disposable.dispose() will revert the optimistic update, and Relay won’t update the store or call any success/error callback, but the network request is not guaranteed to be cancelled.

app/cypress/integration/email.spec.js
```diff
@@ -41,13 +41,12 @@ function requestAccessToOrg() {
   cy.visit('/reporter');
   cy.get('#page-content');
   cy.get('button#org-dropdown').click();
-  cy.wait(500);
   cy.get('.dropdown-menu.show a.dropdown-item')
     .contains('MacDonalds Agriculture, Ltd.')
     .click();
-  cy.wait(500);
+  cy.get('.card h4').contains('MacDonalds Agriculture, Ltd.');
```
wenzowski, line 47: would be awesome if in refactoring we got some semantic css classes in...

app/mutations/BaseMutation.ts
```diff
@@ -53,10 +53,25 @@ export default class BaseMutation<T extends BaseMutationType = never> {
       }
     ) {
       return new Promise<T['response']>((resolve, reject) => {
-        commitMutationDefault<T>(environment, {
+        // Debounced mutations should be commited immediately to perform the optimisticUpdate
+        // The actual request will be cancelled in the network layer
```
wenzowski, line 57: it's only getting deferred in the network layer, right? Up to 250 ms?

app/lib/relay-environment/debounce-mutations.ts
```diff
@@ -0,0 +1,64 @@
...
+} from 'react-relay-network-modern/node8';
+import {CacheConfigWithDebounce} from 'next-env';
+
+let debouncedMutation: {
+  debounceKey: string;
+  debouncedFn: () => void;
+  timeoutId: number;
+} = null;
+
+const debounceMutationMiddleware = (timeout = 250): Middleware => {
```
wenzowski, line 14: love `timeout` as a good default

app/mutations/BaseMutation.ts
```diff
@@ -53,10 +53,25 @@ export default class BaseMutation<T extends BaseMutationType = never> {
...
+        // Here we either dispose of a debounced mutation, or remove it from the map when it errors/completes
+        if (debounceKey) {
+          const previousMutation = debouncedMutationMap.get(debounceKey);
+          if (previousMutation) {
+            previousMutation.dispose();
```
wenzowski, line 62: perfect! We talked about some hiccups with disposables not getting fully disposed...did we resolve the edge case of disposing network-requests that are already in flight?

app/tests/unit/mutations/application/createApplicationMutation.test.ts
```diff
@@ -12,7 +12,6 @@ const mutation = `
     $input: CreateApplicationMutationChainInput!
   ) {
     createApplicationMutationChain(input: $input) {
-      clientMutationId
```
wenzowski, line 15: Is clientMutationId no longer needed for relay's optimism? It's just an echoback field: it's a no-op server-side.

docs/stack.md
```diff
@@ -100,6 +100,33 @@ Other tools such as alembic ([https://alembic.sqlalchemy.org/](https://alembic.s
 
 #### Debouncing Mutations
 
+If we expect a mutation to be dispatched several times in quick succession, and if the last instance of the mutation is the only one required to generate the correct state, then only the final mutation should be sent to the back-end, to reduce the load and network traffic.
+This is the case for instance for mutations that are sent on every key press in an input.
```
wenzowski, line 104: Only idempotent mutations are candidates for debounce.

wenzowski: Some mutations that need to be dispatched in quick succession are a better fit for the batch dispatcher, eg. a counter mutation.

app/tests/unit/mutations/application/createApplicationMutation.test.ts
matthieu-foucault: It's not needed since https://github.com/facebook/relay/releases/tag/v8.0.0

docs/stack.md
matthieu-foucault: Will update the docs for completeness. Regarding the batch dispatcher, should it be opt-in? All mutations are batched (with a 10ms window)

app/mutations/BaseMutation.ts
matthieu-foucault: Right. If the mutation made it past the `debounceMiddleware` then we may dispose of a mutation that will make its way to the serve. I guess the milddeware should remove the mutation from the `debouncedMutationMap` before calling `next()`. Will update that

PR #218, opened 2019-11-22
fixed get reporting year function

blueQcloud:


app/pages/index.tsx
```diff
@@ -30,12 +27,32 @@ export default class Index extends Component<Props> {
...
+      'Apr',
+      'May',
+      'Jun',
+      'Jul',
+      'Aug',
+      'Sep',
+      'Oct',
+      'Nov',
+      'Dec'
+    ];
```
matthieu-foucault: You don't need to do this manually. There is a `DateTimeFormat` object in the standard library: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/format

something like this should work:
```js
const dateTimeFormat = new Intl.DateTimeFormat('en-CA', { year: 'numeric', month: 'short', day: 'numeric' });
dateTimeFormat.format(startDate)

```

schema/deploy/function_get_reporting_year.sql
```diff
@@ -6,12 +6,17 @@ begin;
...
+            where (date_part('year', CURRENT_DATE)) = reporting_year::numeric
+            );
+        end;
     $function$
-        language sql stable;
+        language plpgsql stable;
 
 commit;
+
+--returns current reporting year and related dates
```
matthieu-foucault: By comment, I meant a database comment (`comment on function ggircs_portal.get_reporting_year is`), like the ones that are on our tables. This will make the comment visible on the graphql API too. You'll have to re-generate the graphql schema

schema/deploy/function_get_reporting_year.sql
```diff
@@ -6,12 +6,17 @@ begin;
...
+            select row (_reporting_year.*)
+            from ggircs_portal.reporting_year as _reporting_year
+            where (date_part('year', CURRENT_DATE)) = reporting_year::numeric
+            );
+        end;
     $function$
-        language sql stable;
+        language plpgsql stable;
+
+comment on function ggircs_portal.get_reporting_year is 'Function returning current reporting year and related dates'
```
matthieu-foucault: looks like you're missing a semicolon here

app/pages/index.tsx
```diff
@@ -30,12 +27,13 @@ export default class Index extends Component<Props> {
   render() {
     const {query} = this.props;
     const {session} = query || {};
+    const options = {month: 'short', day: 'numeric', year: 'numeric'};
     const date1: any = query.getReportingYear.applicationOpenDate;
     const startDate: Date = new Date(Date.parse(date1));
     const date2: any = query.getReportingYear.applicationEndDate;
```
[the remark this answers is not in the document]
blueQcloud: i think so, if i don't use it, it gives me a type error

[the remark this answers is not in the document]
matthieu-foucault: Alright, I've figured it out. The type of this field in the graphql API is a `Datetime`, which is a postgraphile type. The relay typescript plugin doesn't know how to deal with this, which is why it was typed as `unknown` (and why you had to cast it as `any`). There is a `customScalars` option, in the newly added `relay.config.js` file, where we can define that this `Datetime` format is actually a string

app/pages/index.tsx
```diff
@@ -30,12 +29,11 @@ export default class Index extends Component<Props> {
...
-    const startDate: Date = new Date(Date.parse(date1));
-    const date2: any = query.getReportingYear.applicationEndDate;
-    const endDate: Date = new Date(Date.parse(date2));
-    const parsedStartDate = startDate.toDateString();
-    const parsedEndDate = endDate.toDateString();
+
+    const date1 = query.getReportingYear.applicationOpenDate;
+    const startDate = new Date(Date.parse(date1));
+    const date2 = query.getReportingYear.applicationEndDate;
+    const endDate = new Date(Date.parse(date2));
```
matthieu-foucault: Now the issue is that using `Date.parse` with a string is actually discouraged.
From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date:
> Note: Parsing of date strings with the Date constructor (and Date.parse(), which works the same way) is strongly discouraged due to browser differences and inconsistencies. Support for RFC 2822 format strings is by convention only. Support for ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01") are treated as UTC, not local.

Looks like we should be using a library instead. I'm sure @wenzowski has an opinion on that :smile:  https://momentjs.com/ perhaps?

matthieu-foucault:
@blueQcloud I'm going to go ahead and add moment (it's being used in other PRs already) and merge this for you

PR #220, opened 2019-11-22
Fix products & benchmarks

dleard:
Products and benchmarks got borked with some of the new updates.
 - Fix create/edit mutation implementations
 - Update UI for products & benchmarks
 - Refactor re-usable search components


app/containers/Products/ProductCreatorContainer.tsx
```diff
@@ -1,51 +1,79 @@
...
-
-    const {environment} = props.relay;
+    resetForm();
+    const {environment} = relay;
     await createProductMutation(environment, variables);
-    createProductFromRef.current.reset();
+  };
+
+  // Schema for JsonSchemaForm component
+  const createProductSchema: JSONSchema6 = {
```
matthieu-foucault: You can move this variable outside of the component, so that the schema prop doesn't change for every render

app/containers/Products/FutureBenchmarks.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const formData = {
+    benchmark: benchmark.benchmark,
+    eligibilityThreshold: benchmark.eligibilityThreshold,
+    startDate: moment(benchmark.startDate).format('DD-MM-YYYY'),
+    endDate: moment(benchmark.endDate).format('DD-MM-YYYY')
+  };
+
+  return (
+    <Container>
```
matthieu-foucault: Do you need the `Container`?

app/containers/Products/ProductListContainer.tsx
```diff
@@ -1,39 +1,104 @@
-import React from 'react';
-import {graphql, createFragmentContainer} from 'react-relay';
+import React, {useEffect} from 'react';
+import {Table, Col, Container, Row} from 'react-bootstrap';
+import {graphql, createRefetchContainer} from 'react-relay';
 import {ProductListContainer_query} from 'ProductListContainer_query.graphql';
+// Import {RelayNetworkLayer} from 'react-relay-network-modern/node8';
```
matthieu-foucault:
```suggestion
```

app/containers/Products/ProductRowItemContainer.tsx
```diff
@@ -36,430 +80,369 @@ export const ProductRowItemComponent = props => {
...
-        ),
-        eligibilityThresholdInput: parseFloat(
-          event.nativeEvent.target[eligibilityThresholdTargetIndex].value
-        ),
-        startDateInput: startDate,
-        prevBenchmarkIdInput: currentBenchmark ? currentBenchmark.rowId : null
+        id: currentBenchmark.id,
+        benchmarkPatch: {
+          deletedAt: moment(),
+          deletedBy: 'Admin'
```
matthieu-foucault: This should probably be the current user's `rowId` instead.

app/containers/Products/ProductRowItemContainer.tsx
```diff
@@ -36,430 +80,369 @@ export const ProductRowItemComponent = props => {
...
-  const background = props.product.state === 'archived' ? 'lightGrey' : '';
-  const buttonVariant =
-    props.product.state === 'archived' ? 'success' : 'warning';
-  const archiveRestore =
-    props.product.state === 'archived' ? 'Restore' : 'Archive';
+  const currentBenchmarkFormData = {
+    benchmark:
+      currentBenchmark && currentBenchmark.benchmark
+        ? currentBenchmark.benchmark
+        : null,
```
matthieu-foucault: This can now be written this way: `benchmark: currentBenchmark?.benchmark ?? null`

app/tests/unit/containers/Products/productCreator.test.tsx
```diff
@@ -1,13 +1,20 @@
 import React from 'react';
-import {shallow} from 'enzyme';
+import {mount} from 'enzyme';
```
matthieu-foucault: It's not a unit test if you use `mount`

app/containers/Products/FutureBenchmarks.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface Props {
+  benchmark: any;
+  environment: RelayModernEnvironment;
+  deleteBenchmark: (...args: any[]) => void;
+}
+
+const FutureBenchmarksComponent: React.FunctionComponent<Props> = ({
+  benchmark,
+  environment,
```
matthieu-foucault: Not a big fan of passing the relay environment through props. Either the update should be done in the parent component via an event handler, or this should be a fragment container

schema/data/portal/application.sql
```diff
@@ -3,7 +3,7 @@ begin;
 with rows as (
 insert into ggircs_portal.application(id, facility_id)
 overriding system value
-values (1,3)
+values (1,3), (2,2)
```
matthieu-foucault: We need to add some comments in these files, make explicit what scenarios/UIs the rows we're inserting are here for.

app/components/SearchBox.tsx
```diff
@@ -0,0 +1,89 @@
...
+    handleEvent('applySearchField', undefined);
+    handleEvent('applySearchValue', undefined);
+  };
+
+  const schema: JSONSchema6 = {
+    type: 'object',
+    properties: {
+      searchField: {
+        type: 'string',
+        title: '  ',
```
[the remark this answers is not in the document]
dleard: I added a classname and hid it with css

app/containers/Products/FutureBenchmarks.tsx
dleard: nope!

app/containers/Products/FutureBenchmarks.tsx
dleard: fixed with a handler function

app/containers/Admin/OrganisationRequestsTable.tsx
```diff
@@ -18,21 +19,40 @@ interface Props {
...
-    {eventKey: 'operator_name', title: 'Operator'}
+    'User ID',
+    'First Name',
+    'Last Name',
+    'Email',
+    'Status',
+    'Operator'
   ];
+
+  const displayNameToColumnNameMap = {
```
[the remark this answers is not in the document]
dleard: Will be fixed by introducing enums

app/containers/Admin/OrganisationRequestsTable.tsx
```diff
@@ -18,21 +19,40 @@ interface Props {
...
     orderByField,
-    orderByDisplay,
     searchField,
     searchValue,
     searchDisplay,
     direction,
     handleEvent
   } = props;
+
+  const tableHeaders = [
```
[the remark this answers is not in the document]
dleard: will fixed by introducing enums in a future PR

schema/data/portal/application.sql
dleard: I added a note in youtrack

PR #224, opened 2019-11-25
fixed ciip applications approve or reject function

blueQcloud:


hamzajaved:
> I'm confused as to what we're fixing in this PR.
> Also, why is there an `updateApplicationStatusByRowIdMutation` in the first place? The application status rows should not be updated, we should be adding more statuses, in order to keep the history of an application status

This seems like a good discussion for a quick parking lot. We probably want to do this for every status

app/containers/Applications/ApplicationRowItemContainer.tsx
```diff
@@ -6,10 +6,10 @@ import Link from 'next/link';
 export const ApplicationRowItem = props => {
   const {ciipApplication = {}} = props;
   const statusBadgeColor = {
```
matthieu-foucault: This object should be defined outside of the component (at the top of this file is fine), and should probably have a `Record` type

app/containers/Applications/ApplicationStatusContainer.tsx
```diff
@@ -26,51 +28,55 @@ export const ApplicationStatus = props => {
...
+            >
+              {props.applicationStatus.applicationStatus}
+            </Dropdown.Toggle>
+            <Dropdown.Menu style={{width: '100%'}}>
+              {Object.keys(statusBadgeColor).map(status => (
+                <DropdownMenuItemComponent
+                  key={status}
+                  itemEventKey={status}
+                  itemFunc={setApplicationStatus}
+                  itemTitle={status}
```
matthieu-foucault: Shouldn't this be `statusBadgeColor[status]`?

app/containers/Applications/ApplicationWizardStep.tsx
```diff
@@ -82,10 +57,11 @@ const ApplicationWizardStep: React.FunctionComponent<Props> = ({
...
 
   if (!application) return null;
-
+  // Save Application status to database
   return (
-    <Form
-      query={query}
+    // Save Application status to database
+    <Form // Save Application status to database
+      query={query} // Save Application status to database
```
matthieu-foucault: What's going on with those comments here?

app/mutations/application/createApplicationStatusMutation.ts
```diff
@@ -0,0 +1,32 @@
...
+import BaseMutation from '../BaseMutation';
+
+const mutation = graphql`
+  mutation createApplicationStatusMutation(
+    $input: CreateApplicationStatusInput!
+  ) {
+    createApplicationStatus(input: $input) {
+      clientMutationId
+      applicationStatus {
+        id
```
matthieu-foucault: As discovered this week, the reason the application status isn't updating in the page is because this mutation response is incomplete. The following needs to be added here:
```
applicationByApplicationId {
  id
  applicationStatus {
    id
    applicationStatus
  }
}
```
This will update the `applicationStatus` field in the `application` with the newly created `applicationStatus`, and should update what is rendered without requiring a refresh

app/containers/Applications/ApplicationWizardConfirmation.tsx
```diff
@@ -16,22 +18,33 @@ interface Props {
...
       environment,
       variables
     );
     console.log(response);
+    // TODO: check response
+    const newUrl = {
+      pathname: '/complete-submit'
+    };
+    router.push(newUrl, newUrl, {shallow: true});
+    // TODO: do we need shallow?
```
matthieu-foucault: we don't need shallow, this can just be `router.push('/complete-submit')`
Shallow routing is for the times where you don't actually change pages, but you want to update the query string, for instance when navigating between application steps

schema/deploy/type_ciip_application_status.sql
```diff
@@ -0,0 +1,8 @@
+-- Deploy ggircs-portal:type_ciip_application_status to pg
+-- requires: schema_ggircs_portal
+
+begin;
+
+create type ggircs_portal.ciip_application_status as enum ('pending', 'approved', 'rejected', 'draft');
```
matthieu-foucault: Maybe we need to add a "requested changes" and "changes submitted" values here, to match what is in the review status type.

schema/sqitch.plan
```diff
@@ -5,6 +5,7 @@
 schema_ggircs_portal 2019-07-11T18:30:51Z Hamza Javed <hamza@button.is> # Add a schema for the GGIRCS Industry Portal (GIP)
 function_update_timestamps [schema_ggircs_portal] 2019-10-01T21:42:55Z Alec Wenzowski <alec@button.is> # A function to be used as a trigger to update the created_at and updated column values
 type_ciip_user_organisation_status [schema_ggircs_portal] 2019-11-06T21:07:19Z Helen Liu <helen@button.is> # added new type file ciip_user_organisation_status
+type_ciip_application_status [schema_ggircs_portal] 2019-11-26T18:39:57Z Helen Liu <helen@button.is> # Added ciip application status type
```
matthieu-foucault: `type_ciip_application_status` is used in the `table_application_status`, so you should add it as a dependency there (you can just edit this file directly and add the new dependency

blueQcloud: do you mean just declare the type in the application status table?

matthieu-foucault: I mean add `type_ciip_application_status` between the square brackets on line 20

app/mutations/application/createApplicationStatusMutation.ts
matthieu-foucault: TODO after #237 gets merged

PR #237, opened 2019-12-03
Fix mutation chain response so relay can update the frontend in the Application review flow

hamzajaved:
Custom mutation chains using computed columns need to be told how to traverse the graphql graph explicitly so that relay can update the store

### DIFF
--- a/app/containers/Applications/ApplicationCommentsByForm.tsx
+++ b/app/containers/Applications/ApplicationCommentsByForm.tsx
@@ -1,0 +1,86 @@
+import React from 'react';
+import {createFragmentContainer, graphql, RelayProp} from 'react-relay';
+import moment from 'moment';
+
+/*
+ * The ApplicationComments renders all the comments on the various sections of the application
+ */
+
+interface Props {
+  applicationReview;
+  relay: RelayProp;
+}
+
+export const ApplicationCommentsByForm: React.FunctionComponent<Props> = props => {
+  const {applicationReview} = props;
+  const reviewComments =
+    applicationReview.reviewCommentsByApplicationReviewId.edges;
+  return (
+    <>
+      <div key={applicationReview.id} className="review-box">
+        <div>
+          Set to <strong>{applicationReview.reviewStatus}</strong>
+          &nbsp;
+          <small>
+            {moment(applicationReview.createdAt).format(
+              'MMM Do YYYY, h:mm:ss a'
+            )}
+          </small>
+        </div>
+        <ul className="comment-box">
+          {reviewComments.map(({node}) => (
+            <li key={node.id} className="comments">
+              <div>{node.description}</div>
+              <div>
+                <small>
+                  {moment(node.createdAt).format('MMM Do YYYY, h:mm:ss a')}
+                </small>
+              </div>
+            </li>
+          ))}
+        </ul>
+      </div>
+
+      <style jsx>{`
+        .form-result-box {
+          padding: 25px;
+          margin-bottom: 20px;
+          border: 1px solid #ff713a;
+          border-radius: 5px;
+          background-color: #fff4dcab;
+          font-size: 14px;
+        }
+
+        .form-result-box h5 {
+          border-bottom: 1px solid #888;
+          padding-bottom: 10px;
+          margin-bottom: 20px;
+        }
+
+        .form-result-box .review-box {
+          border-bottom: 1px solid #d2d2d2;
+          margin-bottom: 20px;
+        }
+      `}</style>
+    </>
+  );
+};
+
+export default createFragmentContainer(ApplicationCommentsByForm, {
+  applicationReview: graphql`
+    fragment ApplicationCommentsByForm_applicationReview on ApplicationReview {
+      id
+      reviewStatus
+      createdAt
+      reviewCommentsByApplicationReviewId {
+        edges {
+          node {
+            id
+            description
+            createdAt
+          }
+        }
+      }
+    }
+  `
+});
--- a/app/containers/Applications/ApplicationCommentsByForm.tsx
+++ b/app/containers/Applications/ApplicationCommentsByForm.tsx
@@ -60,7 +60,7 @@ export default createFragmentContainer(ApplicationCommentsByForm, {
       createdAt
       reviewCommentsByApplicationReviewId(first: 2147483647)
         @connection(
-          key: "ApplicationCommentsByFrom_reviewCommentsByApplicationReviewId"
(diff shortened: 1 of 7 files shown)


app/mutations/application/createApplicationReviewMutation.ts
```diff
@@ -15,6 +15,18 @@ const mutation = graphql`
...
+          applicationReviewsByFormResultId {
+            edges {
+              node {
+                ...ApplicationCommentsByForm_applicationReview
+              }
+            }
+          }
+          applicationReview {
+            ...ApplicationReviewContainer_applicationReview
+          }
```
wenzowski: Why refetch both the list of all `applicationReviewsByFormResultId` and the current `applicationReview`? Smells like an overfetch

app/containers/Applications/ApplicationCommentsByForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+  );
+};
+
+export default createFragmentContainer(ApplicationCommentsByForm, {
+  applicationReview: graphql`
+    fragment ApplicationCommentsByForm_applicationReview on ApplicationReview {
+      id
+      reviewStatus
+      createdAt
+      reviewCommentsByApplicationReviewId {
```
wenzowski: A list of edges should contain an [`@connection`](https://relay.dev/docs/en/pagination-container#connection) directive, to allow for effective updates to the store. (Yes, even though this isn't a`PaginationContainer`)

wenzowski: The `@connection` allows us to find this edge in the store and manipulate it locally (via mutations and subscriptions) without having to refetch the entire query.

app/mutations/application/createApplicationReviewMutation.ts
hamzajaved: applicationReviewsByFormResultId is for refetching details for the comments component
applicationReview is for updating the button component. 

If I remove either, their respective components don't update. 
Solutions?

app/containers/Applications/ApplicationCommentsByForm.tsx
hamzajaved: Cool - will look into it :P

app/mutations/application/createApplicationReviewMutation.ts
hamzajaved: applicationReviewsByFormResultId is for refetching details for the comments component
applicationReview is for updating the button component.
The two components have different render paths.
If I remove either, their respective components don't update.
Solutions?

wenzowski: Why not something like...
```diff
  mutation createApplicationReviewMutation(
    $input: CreateApplicationReviewMutationChainInput!
  ) {
    createApplicationReviewMutationChain(input: $input) {
      applicationReview {
-       id
-       formResultId
-       reviewStatus
+       ...ApplicationCommentsByForm_applicationReview
+       ...ApplicationReviewContainer_applicationReview
        formResultByFormResultId {
          applicationReviewsByFormResultId {
            edges {
              node {
-               ...ApplicationCommentsByForm_applicationReview
+               id
              }
            }
          }
          applicationReview {
-           ...ApplicationReviewContainer_applicationReview
+           id
          }
        }
      }
    }
  }
```
that would avoid re-fetching all the fields for all the connections you're trying to update

wenzowski: ^ still doesn't solve for the fact that `applicationReviewsByFormResultId` isn't using the `@connection` directive, but maybe we don't need it anymore for relay 7?

wenzowski: Ensuring the relay store has the expected state is something we need a test for...have @matthieu-foucault or @dleard already written one of those?

matthieu-foucault: As far as I understood `@connection` is only for pagination.
There are some mutation tests, but they need to be updated to actually import the mutation defined in the code. Currently, I believe that the mutation code was copy-pasted into the test file.

wenzowski: The `@connection` directive annotates any Connection in the store (an array of Edges) such that that specific Connection instance can be retrieved and manipulated using that key

wenzowski: This can be helpful if we have the same Connection appearing at multiple depths in the fragment tree; we can dedupe them using the `@connection` directive (or at least we're supposed to be able to.

hamzajaved: I see. I tried this, but without the nested Id, so it didn't work. This makes sense.

app/containers/Applications/ApplicationCommentsByForm.tsx
```diff
@@ -0,0 +1,75 @@
...
+export default createFragmentContainer(ApplicationCommentsByForm, {
+  applicationReview: graphql`
+    fragment ApplicationCommentsByForm_applicationReview on ApplicationReview {
+      id
+      reviewStatus
+      formResultId
+      createdAt
+      reviewCommentsByApplicationReviewId(first: 2147483647)
+        @connection(
+          key: "ApplicationCommentsByFrom_reviewCommentsByApplicationReviewId"
```
wenzowski:
```suggestion
          key: "ApplicationCommentsByForm_reviewCommentsByApplicationReviewId"
```

app/mutations/application/createApplicationReviewMutation.ts
```diff
@@ -12,9 +12,20 @@ const mutation = graphql`
   ) {
     createApplicationReviewMutationChain(input: $input) {
       applicationReview {
-        id
-        formResultId
-        reviewStatus
+        ...ApplicationReviewContainer_applicationReview
+        ...ApplicationCommentsByForm_applicationReview
+        formResultByFormResultId {
+          applicationReviewsByFormResultId {
```
wenzowski: I see above that `reviewCommentsByApplicationReviewId` has an `@connection` declared. Are we updating one here with this edge list? Note that by refetching the list of `id` fields only we're triggering a rebuild of the edges list. If we were to add a `@connection` handler, we would be able to insert just the new edge into the list without rebuilding it.

hamzajaved: What's should we do?

wenzowski: If this is working, let's leave as-is for now and optimize payload size in a subsequent PR

PR #238, opened 2019-12-03
Search dropdown component

dimak1:


dimak1:
@matthieu-foucault this is how it looks
search-dropdown

app/components/SearchDropdown.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import {Typeahead} from 'react-bootstrap-typeahead';
+
+interface Props {
+  placeholder?: string;
+  options: string[];
```
wenzowski: This attribute [should accept an array of objects](https://github.com/ericgio/react-bootstrap-typeahead/blob/master/docs/Data.md) as well.

app/containers/Forms/FuelFields.tsx
```diff
@@ -26,22 +27,17 @@ const FuelFields: React.FunctionComponent<Props> = ({
       <Col xs={12} md={6}>
         <Form.Group controlId="id.fuelType">
           <Form.Label>Fuel Type</Form.Label>
-          <Form.Control
-            as="select"
-            value={formData.fuelType}
-            onChange={e => {
+          <SearchDropdown
+            placeholder="Select fuel or type to filter..."
+            options={query.allFuels.edges.map(({node}) => node.name)}
```
wenzowski: Would be really nice if we could pass [a list of object options](https://github.com/ericgio/react-bootstrap-typeahead/blob/master/docs/Data.md#arrayobject-wdefault-labelkey) to set the `fuelId` in case there are ever multiple nodes with the same name.

dimak1: You mean options would contain items for list + id and potentially anything else we need to pass?

wenzowski: Yes, instead of a list of strings we'd pass a list of objects with both `id` and `name` attributes, that way selecting the `name` in the UI would update the `id` on the form (would need `labelKey="name"` prop sent to typeahead per docs).

app/components/SearchDropdown.tsx
dimak1: Updated.

app/containers/Forms/FuelFields.tsx
dimak1: Updated in last commit.

dimak1: I see. OK.

dimak1: Updated and added more props, i.e. `inputProps` and `defaultInputValue`.

dimak1:
@wenzowski i made changes, pls take a look when you get a chance. thanks!

PR #239, opened 2019-12-03
Certification signature

dleard:
Implements certifier signature workflow.
 - adds a certify page
 - uses the ApplicationDetailsCardItem component to render the application summary
 - adds a CertificationSignature component for the certifier to sign or upload a signature
 - adds a certification_url table for the created url details


app/pages/certification-redirect.tsx
```diff
@@ -0,0 +1,101 @@
+import React, {Component} from 'react';
+import {Row, Col, Button, Form} from 'react-bootstrap';
+import {graphql} from 'react-relay';
+import moment from 'moment';
+import {certificationRedirectQueryResponse} from 'certificationRedirectQuery.graphql';
+import DefaultLayout from '../layouts/default-layout';
+
+interface Props {
```
matthieu-foucault: This should extend the `CiipPageComponentProps`, which defines the type of `router` for you

app/pages/registration.tsx
```diff
@@ -7,6 +7,7 @@ import DefaultLayout from '../layouts/default-layout';
 
 interface Props extends CiipPageComponentProps {
   query: registrationQueryResponse['query'];
+  router: any;
```
matthieu-foucault: You don't need this, router is defined in the parent interface

app/pages/registration.tsx
```diff
@@ -33,6 +34,10 @@ class Registration extends Component<Props> {
       router
     } = this.props;
     const {ciipUserBySub, sub, givenName, familyName, email} = session || {};
+    let redirectURI = router.query.redirectTo;
+    // The redirectTo loses the query string information, which is needed when redirecting a certifier. The query string is still stored in pageProps
+    if (router?.components?.['/certify'])
+      redirectURI = `${router.query.redirectTo}?applicationId=${router.components['/certify'].props.pageProps.variables.applicationId}`;
```
matthieu-foucault: Looks like that's a bug in `default-layout`, where we need to use `router.asPath` instead of `router.route`

app/package.json
```diff
@@ -128,8 +128,10 @@
     "react-relay": "^7.0.0",
     "react-relay-network-modern": "^4.1.0",
     "react-relay-network-modern-ssr": "^1.2.4",
+    "react-signature-canvas": "^1.0.3",
     "relay-runtime": "^7.0.0",
-    "sinon": "^7.4.1"
+    "sinon": "^7.4.1",
```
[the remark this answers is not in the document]
wenzowski: wasn't added by this PR, but currently commented out in `lib/setupTests.ts`

app/containers/Forms/CertificationSignature.tsx
```diff
@@ -0,0 +1,152 @@
...
+          certifiedByUser: props.user
+        }
+      }
+    };
+    const response = await updateApplicationMutation(environment, variables);
+    console.log(response);
+  };
+
+  const router = useRouter();
+  // Change application status to 'pending' on application submit	  // Change application status to 'pending' on application submit
```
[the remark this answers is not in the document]
dleard: copy paste!

app/pages/certification-redirect.tsx
```diff
@@ -86,7 +86,7 @@ class CertificationRedirect extends Component<Props> {
                     action={`/login?redirectTo=${decodeURI(redirectURI)}`}
                     method="post"
                   >
-                    <Button type="submit">GOGOGO</Button>
+                    <Button type="submit">Continue</Button>
```
wenzowski: GOGO!

dleard: I like GOGOGO better, but figured I should change it

app/containers/Forms/CertificationSignature.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const submitApplication = async () => {
+    const {environment} = props.relay;
+    const variables = {
+      input: {
+        applicationStatus: {
+          applicationId: props.application.rowId,
+          applicationStatus: 'PENDING' as CiipApplicationStatus
```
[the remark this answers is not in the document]
matthieu-foucault: You have a button that submits the application and redirects to `complete-submit` in two different places, you should create a `SubmitApplicationButton` fragment container that you can reuse

app/pages/certification-redirect.tsx
```diff
@@ -0,0 +1,103 @@
...
+                  On behalf of <strong>{organisationName}</strong>, for facility{' '}
+                  <strong>{facilityName} </strong>
+                  please certify the information reported in the application is
+                  correct.
+                </p>
+                <p>
+                  Please log in or register in order to access this page.
+                  <br />
+                  You will be redirected to the certification page after doing
+                  so.
```
matthieu-foucault: What if they are already logged in? When rendering you should check `session?.ciipUserBySub` and redirect to the certification page if it is not `undefined`

schema/deploy/function_set_random_id.sql
```diff
@@ -0,0 +1,37 @@
...
+create extension if not exists pgcrypto;
+create function ggircs_portal.set_random_id()
+  returns trigger as $$
+
+declare
+  new_id varchar(1000);
+begin
+  while new_id in (select id from ggircs_portal.certification_url) or new_id is null loop
+    select gen_random_bytes(64) into new_id;
+  end loop;
```
wenzowski: This loop avoids a [birthday attack](https://en.wikipedia.org/wiki/Birthday_attack) by detecting and avoiding hash collisions by re-running our CSPRNG function until there is no collision. @Krocodial let me know if you want to run through this together?

@dleard this is looking really good. Can we add a collision warning? Something like `raise warning 'ggics_portal.set_random_id() experienced hash collision on % – rerunning hash function', new_id;` This way, if we run out of address space in our hash column, or our CSPRNG fails or is compromised we'll have some logging to let us know what's going on.

schema/deploy/function_set_random_id.sql
```diff
@@ -0,0 +1,37 @@
...
+begin;
+create extension if not exists pgcrypto;
+create function ggircs_portal.set_random_id()
+  returns trigger as $$
+
+declare
+  new_id varchar(1000);
+begin
+  while new_id in (select id from ggircs_portal.certification_url) or new_id is null loop
+    select gen_random_bytes(64) into new_id;
```
wenzowski: Thinking about this a bit more, we need to encode the `bytea` result from `gen_random_bytes` into a URL-safe id to be stored as `varchar(1000)`.

```suggestion
    -- Securely generate a pseudo-random hash and re-encode it as a base64 string.
    new_id := encode(gen_random_bytes(64), 'base64');

    -- Base64 encoding contains 2 URL unsafe characters by default.
    -- The URL-safe version has these replacements.
    new_id := replace(key, '/', '_'); -- url safe replacement
    new_id := replace(key, '+', '-'); -- url safe replacement
```

app/containers/Forms/CertificationSignature.tsx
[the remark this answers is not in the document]
matthieu-foucault: @hamzajaved  confirmed that we don't want the certifier to submit

schema/deploy/function_set_random_id.sql
dleard: I added this in to the deploy file, but replacing the '+' with gen_random_bytes(64) wasn't working. There was always a '+' generated in the id, but not picked up on the front end. Maybe a '+' is automatically added to the string at a certain length? To resolve this for now I dropped the bytes to 32.

app/package.json
```diff
@@ -128,8 +128,10 @@
...
+    "uuid": "^3.3.3"
```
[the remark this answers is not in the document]
matthieu-foucault: @dleard is this used or not?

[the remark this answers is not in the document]
dleard: uuid isn't used anymore, I used it in a very early iteration of this PR. I'll yarn remove it. sinon, I have no idea what it's part of or if its used in anything

app/pages/certification-redirect.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    <p>
+                      Please log in or register in order to access this page.
+                      <br />
+                      You will be redirected to the certification page after
+                      doing so.
+                    </p>
+
+                    <Col md={{span: 5, offset: 1}}>
+                      <Form
+                        action={`/login?redirectTo=${decodeURI(redirectURI)}`}
```
matthieu-foucault: The `decodeURI` doesn't seem to do much. Investigating this to figure it how it should be handled. There might be a bug in the existing login buttons too

matthieu-foucault: Yeah so this is wrong (although somehow handled nicely by the browser and/or express) https://github.com/bcgov/cas-ciip-portal/blob/18f1e3e1ebc7d17e3c1e2882d149fd5a215fd77e/app/components/LoginButton.tsx#L13

We shouldn't decode the URI. Can you please remove the use of `decodeURI` in `LoginButton`/

As for your code above, it should be using [`encodeURIComponent()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

dleard: done!

app/pages/certification-redirect.tsx
```diff
@@ -0,0 +1,123 @@
...
+      facility?.organisationByOrganisationId?.operatorName;
+    const applicationId = this.props.router.query.id;
+    const redirectURI = `/certify?applicationId=${applicationId}`;
+
+    const createdAtMoment = query?.certificationUrlByRowId?.createdAt
+      ? moment(query?.certificationUrlByRowId?.createdAt)
+      : null;
+    const currentMoment = moment();
+    // TODO(Dylan): handle expiry of url validation on the back end
+    // Sets an expiry of 7 days for the certification URL
```
matthieu-foucault: If you're creating the mutation chain to certify in this PR, you could handle the expiry there. Another option is to use a computed column. Doing both would be best

app/package.json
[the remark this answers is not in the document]
wenzowski: sinon isn't used by anything, only commented out code which can be deleted

schema/deploy/function_set_random_id.sql
```diff
@@ -0,0 +1,44 @@
...
+create function ggircs_portal.set_random_id()
+  returns trigger as $$
+
+declare
+  new_id varchar(1000);
+begin
+  while new_id in (select id from ggircs_portal.certification_url) or new_id is null loop
+    if new_id in (select id from ggircs_portal.certification_url) then
+      raise warning 'ggircs_portal.set_random_id() experienced hash collision on %s - rerunning hash function & generating new id', new_id;
+    end if;
```
wenzowski: needs to happen after the `replace()` to match correctly

schema/deploy/function_set_random_id.sql
```diff
@@ -0,0 +1,44 @@
...
+begin;
+create extension if not exists pgcrypto;
+create function ggircs_portal.set_random_id()
+  returns trigger as $$
+
+declare
+  new_id varchar(1000);
+begin
+  while new_id in (select id from ggircs_portal.certification_url) or new_id is null loop
+    if new_id in (select id from ggircs_portal.certification_url) then
```
wenzowski: This is currently running the select statement twice per iteration of the loop.

Have a look at [`loop` and `exit`](https://www.postgresql.org/docs/current/plpgsql-control-structures.html#PLPGSQL-CONTROL-STRUCTURES-LOOPS) for this?

schema/deploy/function_set_random_id.sql
dleard: nevermind it works.

app/containers/Applications/ApplicationWizardConfirmation.tsx
```diff
@@ -18,29 +16,39 @@ interface Props {
...
-  // Change application status to 'pending' on application submit
-  const submitApplication = async () => {
+  const [copySuccess, setCopySuccess] = useState('');
+  const [url, setUrl] = useState();
+  const copyArea = useRef(url);
+
+  const copyToClipboard = () => {
+    const el = copyArea;
+    el.current.select();
+    document.execCommand('copy');
```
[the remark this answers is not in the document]
dleard: I've added a throw error on a failed copy and tagged this with a TODO for now.

app/containers/Applications/ApplicationWizardConfirmation.tsx
```diff
@@ -18,29 +16,39 @@ interface Props {
...
       variables
     );
     console.log(response);
-    // TODO: check response
-    router.push('/complete-submit');
+    if (window) {
+      setUrl(
+        `${window.location.host}/certification-redirect?rowId=${response.createCertificationUrl.certificationUrl.rowId}&id=${props.query.application.id}`
+      );
+    } else console.log('No window location found');
```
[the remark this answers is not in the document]
dleard: I've added a try-catch block with `throw error`

app/containers/Forms/CertificationPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * and allows the user to submit their application.
+ */
+
+interface Props {
+  application: CertificationPage_application;
+  relay: RelayProp;
+  isAnalyst: boolean;
+}
+
+export const CertificationPageComponent: React.FunctionComponent<Props> = props => {
```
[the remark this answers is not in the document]
dleard: You're right. It started out different, but now it's totally the same.

app/pages/certification-redirect.tsx
dleard: I added a trigger to create an expires_at column

app/containers/Forms/CertificationSignature.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const signature = sigCanvas.current
+      .getTrimmedCanvas()
+      .toDataURL('image/png');
+    const {environment} = props.relay;
+    const variables = {
+      input: {
+        id: props.application.id,
+        applicationPatch: {
+          certificationSignature: signature,
+          certifiedByUser: props.user
```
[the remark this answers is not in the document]
dleard: I added a trigger that updates the user_id in the application table on update of certification_signature.

PR #240, opened 2019-12-06
WIP(broken): split storybook out to a separate app

wenzowski:
makes a bunch of storybook things better & tries to extract storybook to its own folder...not yet working


wenzowski:
This went sideways. :(

PR #242, opened 2019-12-06
Feature/all facilities list

blueQcloud:


app/components/Layout/Subheader.tsx
```diff
@@ -17,7 +17,7 @@ const Subheader = () => (
             </Link>
           </li>
           <li>
-            <a href="#">My Applications</a>
+            <a href="/facilities-list">My Applications</a>
```
matthieu-foucault: You should be using a Next.js Link instead of an `a` tag.

app/containers/Facilities/FacilitiesListContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+              <SortableTableHeader
+                key={header.columnName}
+                sort={handleEvent}
+                headerVariables={header}
+              />
+            ))}
+            <th />
+          </tr>
+        </thead>
+        {/* filter will not work if there's duplicate facilities in the table, bc the key is not unique */}
```
matthieu-foucault: Does this still apply? The key needs to be unique, otherwise React may be very confused

app/containers/Facilities/FacilitiesListContainer.tsx
```diff
@@ -0,0 +1,143 @@
+import React, {useEffect} from 'react';
+import {graphql, createRefetchContainer} from 'react-relay';
+import {Table, Container, Row, Col} from 'react-bootstrap';
+import SortableTableHeader from 'components/SortableTableHeader';
+import SearchBox from 'components/SearchBox';
+import FacilitiesRowItemContainer from './FacilitiesRowItemContainer';
+
+export const FacilitiesList = props => {
```
matthieu-foucault: SonarCloud is complaining, rightly so, about this component being a duplicate of `ApplicationsList`. The only difference as far as I can tell is how the rows are rendered. To remove that code duplication, we need to create a new reusable component with a `renderRow` prop.

app/containers/Facilities/FacilitiesRowItemContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+> = {
+  DRAFT: 'warning',
+  PENDING: 'info',
+  REJECTED: 'danger',
+  APPROVED: 'success',
+  CHANGES_SUBMITTED: 'primary',
+  REQUESTED_CHANGES: 'secondary'
+};
+export const FacilitiesRowItem = props => {
+  console.log(props);
```
matthieu-foucault:
```suggestion
```

app/containers/Facilities/FacilitiesRowItemContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+  'info' | 'danger' | 'success' | 'warning' | 'primary' | 'secondary'
+> = {
+  DRAFT: 'warning',
+  PENDING: 'info',
+  REJECTED: 'danger',
+  APPROVED: 'success',
+  CHANGES_SUBMITTED: 'primary',
+  REQUESTED_CHANGES: 'secondary'
+};
+export const FacilitiesRowItem = props => {
```
matthieu-foucault: props need to be typed

app/containers/Facilities/FacilitiesRowItemContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+      pathname: hasSwrsReport
+        ? '/ciip-application-swrs-import'
+        : '/ciip-application',
+      query: {
+        applicationId: response.createApplicationMutationChain.application.id
+      }
+    });
+  };
+
+  const applyButton = () => {
```
matthieu-foucault: `applyButton` should be a separate component.

app/containers/Organisations/UserOrganisation.tsx
```diff
@@ -14,6 +14,7 @@ export const UserOrganisationComponent = props => {
     PENDING: 'info',
     APPROVED: 'success'
   };
+  console.log(typeof userOrganisation.organisationByOrganisationId.rowId);
```
matthieu-foucault:
```suggestion
```

app/containers/Facilities/FacilitiesListContainer.tsx
blueQcloud: this has been fixed

app/containers/Facilities/FacilitiesListContainer.tsx
blueQcloud: So, a new component that will get all that data as props and outputs it in table rows?

schema/deploy/type_facility_application_status.sql
```diff
@@ -0,0 +1,25 @@
+-- Deploy ggircs-portal:type_facility_application_status to pg
+-- requires: schema_ggircs_portal
+-- requires: table_facility
+-- requires: table_application
+-- requires: type_ciip_application_status
+
+begin;
+
+create type ggircs_portal.facility_application_status as (
```
matthieu-foucault: This type needs to be renamed as it is confusing (sounds like it's the status of the relation between a facility and an application). `facility_search_result` would be better.

schema/deploy/function_search_all_facilities.sql
```diff
@@ -0,0 +1,64 @@
...
+                        from ggircs_portal.application_status
+                        group by application_id)a on a.application_id = t.application_id and a.max_date = t.created_at),
+
+                        applicationDetails as (
+                            select s.application_status as application_status, a.id as id, a.facility_id as facility_id
+                            from ggircs_portal.application as a join applicationStatus as s on a.id = s.id
+                        ),
+
+                        tempTable as (select f.*, ad.application_status as application_status, ad.id as application_id
+                        from ggircs_portal.facility as f
```
matthieu-foucault: This should only return the facilities that the user has access to. You can use the `session()` function to retrieve the `sub` field (which is a foreign key on `ciip_user.uuid`) and get the current user id. From there you need to do a join on the list of organisations the user has access to (via the `ciip_user_organisation` table), and only query the facilities from these organisations.

schema/data/portal/application.sql
```diff
@@ -3,7 +3,7 @@ begin;
 with rows as (
 insert into ggircs_portal.application(id, facility_id, reporting_year)
 overriding system value
-values (1,3, 2018), (2,2, 2019)
+values (1,3, 2018)
```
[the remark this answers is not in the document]
dleard: This was causing a duplicate because of a function call `select create_application_mutation_chain(2)` in schema/data/portal/form_result.sql. Since we were hardcoding (2,2,2019) and making this call it was making 2 applications for the facility with id=2. We handled the problem caused by having two applications for one facility by making the facilities search function more robust, but we probably shouldn't be making duplicate applications in the dummy data anyway.

schema/deploy/function_search_all_facilities.sql
dleard: to be done in #251

PR #243, opened 2019-12-06
Inject additional data in product form based on product type

matthieu-foucault:


app/containers/Forms/ProductionFields.tsx
```diff
@@ -1,127 +1,224 @@
...
-            as="select"
-            value={formData.product}
-            onChange={e => {
-              onChange({
-                ...formData,
-                product: (e.nativeEvent.target as HTMLInputElement).value,
-                productUnits: undefined
-              });
-            }}
+      <FieldTemplate
```
matthieu-foucault: @hamzajaved By using the registry's `FieldTemplate` and `StringField`, this allows for this component to be reused in both the `Form` and the `ApplicationDetails` pages. It will either render an input or the string value based on what is defined in the `JsonSchemaForm`'s props. Although passing all of these props is a bit verbose, this makes the component fully generic. We can do the same type of change to the Emissions component later, to avoid the current code duplication

matthieu-foucault:
This can use a review, while I am adding more products and additionalData forms in the data files

app/containers/Applications/ApplicationWizardStep.tsx
```diff
@@ -50,6 +50,7 @@ const ApplicationWizardStep: React.FunctionComponent<Props> = ({
 
   const onValueChanged = async change => {
     const {formData} = change;
+    console.log(formData);
```
wenzowski: how would you feel about turning on [no-console](https://eslint.org/docs/rules/no-console)

app/containers/Forms/FormFieldTemplate.tsx
```diff
@@ -53,12 +66,12 @@ const FormFieldTemplate: React.FunctionComponent<FieldTemplateProps> = ({
     >
       <Form.Label htmlFor={id}>
         {label}
-        {required ? '*' : null}
+        {required ? ' *' : null}
```
wenzowski: bumble or significant space?

app/containers/Forms/ProductionFields.tsx
```diff
@@ -1,127 +1,224 @@
...
+    FieldTemplate
+  }: {
+    FieldTemplate: React.FunctionComponent<any>;
+  } = registry as any;
+  // Not using the types definded in @types/react-jsonschema-form as they are out of date
+
+  const handleProductChange = (product: string) => {
+    const productUnits = query.allProducts.edges.find(
+      ({node}) => node.name === product
+    )?.node.units;
```
wenzowski: This would be more robust matching on the product `id` not the `name`.
We don't have a graphql field to the current product?

app/containers/Forms/ProductionFields.tsx
wenzowski: ^ this. 👍 
Can we get this committed as a `jsdoc` or `tsdoc` description block on the class?

app/containers/Forms/ProductionFields.tsx
```diff
@@ -1,127 +1,224 @@
...
+          formContext={formContext}
+          disabled={disabled}
+          readonly={readonly}
+          name="allocationFactor"
+          onChange={allocationFactor => {
+            onChange({
+              ...formData,
+              allocationFactor
+            });
+          }}
```
wenzowski: `handleX` callback style please

app/containers/Forms/ProductionFields.tsx
```diff
@@ -1,127 +1,224 @@
...
+              formContext={formContext}
+              disabled={disabled}
+              readonly={readonly}
+              name="quantity"
+              onChange={quantity => {
+                onChange({
+                  ...formData,
+                  quantity
+                });
+              }}
```
wenzowski: `handleX` callback style please

app/containers/Forms/ProductionFields.tsx
```diff
@@ -1,127 +1,224 @@
...
+              formData={formData.productUnits}
+              autofocus={autofocus}
+              idSchema={idSchema}
+              registry={registry}
+              errorSchema={errorSchema}
+              formContext={formContext}
+              name="units"
+              onChange={() => {
+                console.error("don't change me");
+              }}
```
wenzowski: Is there an advantage to `console.error` over `throw`?

app/containers/Forms/ProductionFields.tsx
```diff
@@ -1,127 +1,224 @@
...
         edges {
           node {
             name
             units
+            productFormByProductFormId {
+              productFormSchema
+            }
           }
         }
       }
```
wenzowski: How big is the `productFormSchema`? Should we be fetching `n` records for first render?

app/containers/Forms/SummaryFormArrayFieldTemplate.tsx
```diff
@@ -3,28 +3,28 @@ import {ArrayFieldTemplateProps} from 'react-jsonschema-form';
...
-        <TitleField
-          id={null}
-          title={title === 'gases' || title === 'Source Types' ? null : title}
-          required={false}
-        />
-      </Col>
-      {items.map(element => {
+      {items.map((element, i) => {
         return (
           <React.Fragment key={element.index}>
```
wenzowski: Can we not use `key=` on the `<>` fragment style?

app/relay.config.js
```diff
@@ -3,7 +3,8 @@ module.exports = {
   language: 'typescript',
   artifactDirectory: './__generated__',
   customScalars: {
-    Datetime: 'String'
+    Datetime: 'String',
+    JSON: 'any'
```
wenzowski: I seem to recall there were reasons why this couldn't be a useful type

app/tests/integration/Products/productCreator.test.tsx
```diff
@@ -27,7 +27,7 @@ describe('ProductCreator', () => {
           .find('label')
           .filter('[htmlFor="root_product"]')
           .text()
-      ).toBe('Product*');
+      ).toBe('Product *');
```
wenzowski: This answers the earlier question about significant space...disregard

schema/data/portal/form_json/electricity_and_heat.json
```diff
@@ -60,43 +136,50 @@
...
-        "sold",
-        "consumedOnSite",
         "generatedOnSite",
+        "consumedOnSite",
+        "sold",
         "*"
       ],
       "purchased": {
-        "ui:col-md": 4
+        "ui:col-md": 3
```
wenzowski: why do we need markup in our schema?

app/containers/Applications/ApplicationWizardStep.tsx
matthieu-foucault: @dleard may object to that (there's a lot of mutation responses being logged in the console right now)

app/containers/Forms/FormFieldTemplate.tsx
matthieu-foucault: Significant. I can put `<>&nbsp;*</>` instead

schema/data/portal/form_json/electricity_and_heat.json
matthieu-foucault: This is the `uiSchema`, it defines how things are rendered. The actual json-schema is the `schema` object above

app/containers/Forms/ProductionFields.tsx
matthieu-foucault: The value we use here will be stored in the form data (and in the database). Maybe `rowId` instead?

wenzowski: Ah, yes. Good call, `rowId` would be best

schema/data/portal/form_json/electricity_and_heat.json
wenzowski: It's particularly the tightly-coupled-to-bootstrap bit of the `col-md` class that I'm fussing about. It isn't hugely important in this PR, and it would be nice to decouple css from render config. Would it be hard to get the `uiSchema` to process `"ui:width": "one-quarter"` or something like that?

matthieu-foucault: It's actually not coupled to bootstrap, but to our code (which happens to use bootstraps naming convention): https://github.com/bcgov/cas-ciip-portal/blob/02ab1208dfe8241f677ab21814d89c09652a8ef3/app/containers/Forms/FormFieldTemplate.tsx#L38-L64

wenzowski: In chatting about this, I think there might be further design work surrounding `uiSchema` in future, but this is good as-is and shouldn't block this merge.

app/containers/Forms/SummaryFormArrayFieldTemplate.tsx
matthieu-foucault: Nope

app/containers/Forms/ProductionFields.tsx
matthieu-foucault: It's an object with a dozen properties (each with a type, title, min/max values, etc) at most. 
I'm sure this can be improved but I don't think it belongs in this PR

app/relay.config.js
matthieu-foucault: It's not really useful, but by default it's defined as an `unknown`, which we were casting as an `any`

PR #251, opened 2019-12-14
Industry can make revisions

dleard:
Adds versioning to an application
  - creates and intermediary revision table
  - moves status to be a child of the revision table
  - moves form results to be a child of the revision table
  - adds extra computed columns to application for version tracking
  - adds mutation chain for creating a new version
  - edits create application mutation chain to include versioning


schema/deploy/function_create_application_mutation_chain.sql
```diff
@@ -21,9 +21,12 @@ begin
   insert into ggircs_portal.application(facility_id)
   values (facility_id_input) returning id into new_id;
 
-  -- Insert new value with application_id fk into application_status
-  insert into ggircs_portal.application_status(application_id, application_status)
-  values (new_id, 'draft');
+  insert into ggircs_portal.application_revision(application_id, version_number)
```
matthieu-foucault: This function should call `create_application_revision_mutation_chain`

schema/deploy/view_ciip_application.sql
```diff
@@ -5,41 +5,7 @@
...
-       x.id,
-       (x.facility_data ->> 'facilityName')::varchar(1000) as facility_name,
-       (y.operator_data ->> 'name')::varchar(1000) as operator_name,
-       x.submission_date as submission_date,
-       (x.facility_data ->> 'bcghgid')::numeric as bcghgid,
-       z.application_status as application_status,
-       '2018'::varchar(1000) as reporting_year
-
-    from x,y,z where x.id = y.id and x.id = z.id
+    select 1 as id from ggircs_portal.application
```
matthieu-foucault: Is the view still needed somewhere?

dleard: I don't think so, but I'd left it in until I was sure. I'm going to remove it.

schema/deploy/function_application_revision_application_revision_status.sql
```diff
@@ -0,0 +1,24 @@
...
+    returns ggircs_portal.application_revision_status
+as
+$function$
+declare
+begin
+    return (
+        select row(_application_revision_status.*)
+        from ggircs_portal.application_revision_status as _application_revision_status
+        where _application_revision_status.application_id = application_revision.application_id
+        and _application_revision_status.version_number = (select max(version_number) from ggircs_portal.application_revision as _application_revision
```
matthieu-foucault: shouldn't this be `and _application_revision_status.version_number = application_revision.version_number`

schema/deploy/function_create_application_mutation_chain.sql
```diff
@@ -9,51 +9,16 @@ returns ggircs_portal.application
...
-  form_result jsonb;
-  init_function varchar(1000);
-  query text;
 begin
 
   --Insert new value into application
-  insert into ggircs_portal.application(facility_id)
-  values (facility_id_input) returning id into new_id;
+  insert into ggircs_portal.application(facility_id, reporting_year)
+  values (facility_id_input, (select to_char(now(), 'YYYY')::int)) returning id into new_id;
```
matthieu-foucault: Add `-- TODO: use a current_reporting_year function (which should raise an exception if we are not in the reporting window`

schema/deploy/table_form_result.sql
```diff
@@ -8,16 +8,19 @@ create table ggircs_portal.form_result (
   id integer primary key generated always as identity,
   form_id int not null references ggircs_portal.form_json(id),
   application_id int not null references ggircs_portal.application(id),
-  submission_date timestamp with time zone default now(),
+  version_number int not null default 1,
```
matthieu-foucault: not sure this should have a default

schema/deploy/function_application_application_revision_status.sql
```diff
@@ -0,0 +1,25 @@
+-- Deploy ggircs-portal:function_application_application_revision_status to pg
+-- requires: table_application_revision_status
+-- requires: table_application
+
+begin;
+
+create or replace function ggircs_portal.application_application_revision_status(application ggircs_portal.application)
```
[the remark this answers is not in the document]
dleard: At the moment, I'm using the one directly on Application as the latest status for the application, regardless of draft/submitted. I think this will probably get refactored out when we refactor the fragments from Application -> ApplicationRevision

schema/deploy/function_application_revision_application_revision_status.sql
dleard: yup

app/containers/Applications/ApplicationListContainer.tsx
```diff
@@ -29,14 +29,14 @@ export const ApplicationList = props => {
...
-    Status: 'application_status'
+    Status: 'application_revision_status'
   };
   const body = (
     <tbody>
       {edges.map(edge => (
         <ApplicationRowItemContainer
-          key={edge.node.id}
-          ciipApplication={edge.node}
+          key={edge.node.rowId}
```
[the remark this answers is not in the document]
dleard: The search result function does not return a relay id. I think it's because the id is not a primary key as it can change with each result. It's created with row_number() over () in the sql query. So relay only creates a rowId for us.

app/components/Forms/ApplicationFormNavbar.tsx
```diff
@@ -23,11 +25,16 @@ const ApplicationFormNavbarComponent: React.FunctionComponent<Props> = props =>
...
-                  applicationId: application.id
+                  applicationId: application.id,
+                  version: props.version
                 }
               }}
             >
-              <Nav.Link active={node.id === props.formResultId}>
+              <Nav.Link
+                active={
+                  node.id === props.formResultId && !props.confirmationPage
```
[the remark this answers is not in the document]
dleard: formResultId wasn't being set to undefined on goToConfirmation, it was carrying over the last Id. Fixed

PR #285, opened 2019-12-30
Reporter can add organisation

dleard:
- Generalizes 'Add Facility' component to accomodate organisations
- adds createOrganisation mutation
- fixes incorrect primary key in ciip_user_organisation table


app/components/AddOrganisationFacility.tsx
```diff
@@ -0,0 +1,125 @@
...
+import FormFieldTemplate from 'containers/Forms/FormFieldTemplate';
+import FormObjectFieldTemplate from 'containers/Forms/FormObjectFieldTemplate';
+import addOperatorSchema from 'components/organisation/addOrganisationSchema';
+import addFacilitySchema from 'components/facility/addFacilitySchema';
+
+interface Props {
+  onAddOrganisation?: (...args: any[]) => void;
+  onAddFacility?: (...args: any[]) => void;
+  organisationRowId?: number;
+  reportingYear?: number;
```
matthieu-foucault: Aren't all of those props required?

app/components/AddOrganisationFacility.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+const AddOrganisationFacility: React.FunctionComponent<Props> = props => {
+  const {
+    onAddOrganisation,
+    onAddFacility,
+    organisationRowId,
+    reportingYear
+  } = props;
+  const [modalShow, setModalShow] = useState(false);
```
matthieu-foucault: Rename to `isModalVisible`

app/components/AddOrganisationFacility.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const variables = {
+      input: {
+        facility: {
+          facilityName,
+          facilityType,
+          bcghgid,
+          naicsCode,
+          ...mailingAddress,
+          // TODO: Our facility table currently stores reportingYear as a string. Should it be an int? It's also a string in swrs & init functions
+          reportingYear: String(reportingYear),
```
matthieu-foucault: I don't think that reportingYear should be a field in the facility table.

app/components/facility/addFacilitySchema.ts
```diff
@@ -0,0 +1,70 @@
+import {JSONSchema6} from 'json-schema';
+
+const addFacilitySchema: JSONSchema6 = {
+  type: 'object',
+  title: 'Facility Details',
+  properties: {
```
matthieu-foucault: I would only keep the facility name and BC GHG ID. The other information doesn't really need to be in the facility table

app/components/organisation/addOrganisationSchema.ts
```diff
@@ -0,0 +1,67 @@
+import {JSONSchema6} from 'json-schema';
+
+const addOperatorSchema: JSONSchema6 = {
+  type: 'object',
+  title: 'Operator Details',
+  properties: {
+    operatorName: {
```
matthieu-foucault: Similarly to the facility, I would only keep the operator name in here (and maybe the CRA number)

app/components/AddOrganisationFacility.tsx
dleard: They are required conditionally, if AddOrganisationFacility is being rendered from the FacilityList container it needs the bottom 3 props, if being rendered from the Organisations container it only needs the first prop.

PR #290, opened 2020-01-03
Add import_swrs_organisation_facility function in the db

matthieu-foucault:
That function is now invoked when deploying the test data


matthieu-foucault:
This needs https://github.com/bcgov/cas-ggircs/pull/132 to be merged first

PR #304, opened 2020-01-10
Fix height for Org dropdown

dimak1:


app/containers/Organisations/Organisations.tsx
```diff
@@ -149,17 +149,25 @@ export const OrganisationsComponent: React.FunctionComponent<Props> = props => {
...
-                />
-              );
-            })}
+            <div
+              style={{
+                maxHeight: '250px',
+                overflow: 'hidden',
+                overflowY: 'scroll'
+              }}
+            >
```
matthieu-foucault: - This style should be on the dropdown component itself. 
- You should use a `style jsx` component instead of the inline style attribute

dimak1: - If on dropdown then Search input field will be scrollable as well, which we don't want probably

PR #320, opened 2020-01-14
UI Merge: Checksum form results

dleard:
- Adds a trigger to update form_result_status on update of application_revision_status in ggircs_portal.application_revision_status
 - Fixes types in several tables for updated_by and created_by
 - Updates update_timestamps trigger function to set updated_at & created_by with a user id
 - Adds tests for application_revision_status table & checksum trigger function


schema/test/unit/function_checksum_form_results_test.sql
```diff
@@ -0,0 +1,83 @@
...
+select plan(4);
+
+select has_function(
+  'ggircs_portal', 'checksum_form_results',
+  'Function checksum_form_results should exist'
+);
+
+-- Insert organisation & facility test data
+insert into ggircs_portal.organisation(operator_name) values ('test org');
+insert into ggircs_portal.facility(organisation_id, facility_name) values (1, 'test facility');
```
matthieu-foucault: there's already organisations and facilities in the database (inserted by the deploy-data.sh script), so you don't need those

PR #322, opened 2020-01-14
Data tests + fix schema unit tests to run against dev data

dimak1:


dimak1:
@matthieu-foucault I added data tests, but it takes slightly longer to run them as for each deploy we drop the database. Should I improve it or we dont care how long it takes?

matthieu-foucault:
@dimak1  as long as it stays a reasonable amount of time, then it's fine

dimak1:
> @dimak1 as long as it stays a reasonable amount of time, then it's fine

Less than 3 min for all tests

PR #329, opened 2020-01-20
define allowed groups for each page

junminahn:
define allowed groups for each page and add mock resolver for userGroups in relay testing

### DIFF
--- a/app/lib/user-groups.js
+++ b/app/lib/user-groups.js
@@ -1,21 +1,44 @@
 const GROUP_META = require('../data/groups.json');
 
-const getUserGroupLandingRoute = groups => {
-  let priorityGroup = GROUP_META[groups[0]];
+const getAllGroupNames = () => {
+  return Object.keys(GROUP_META);
+};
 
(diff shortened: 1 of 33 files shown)


app/lib/user-groups.js
```diff
@@ -0,0 +1,21 @@
+const GROUP_META = require('../data/groups.json');
+
+const getUserGroupLandingRoute = groups => {
+  let priorityGroup = GROUP_META[groups[0]];
+
+  for (let x = 1; x < groups.length; x++) {
+    const curr = GROUP_META[groups[x]];
```
matthieu-foucault: This may be undefined as "Realm Administrator" may not be the first group in the list

junminahn: I assumed that Realm Administrator does not have access to the portal, but I will add it into group list

matthieu-foucault: I think this function needs to be able to handle and ignore unknown groups.

junminahn: definitely, it shouldn't crash. I am thinking to set unkown groups to just guest

PR #382, opened 2020-02-10
Fix a couple of flaky cypress tests and fix package vulnerability

matthieu-foucault:


wenzowski:
Looks like @hamzajaved beat me to it

PR #395, opened 2020-02-14
Update incentive function

dleard:
- Fix ciip_incentive function to use real electricity / heat allocation values
- Add an id column to return type for unique key
- Coalesce incentive for product to equal 0 when no benchmark exists
- Add importedElectricityAllocationFactor & importedHeatAllocationFactor values to form_result dummy data.
- Change front-end to use incentive function instead of view
- Turn ciip_incentive function into a computed column on application_revision


schema/deploy/ciip_functions/ciip_incentive.sql
```diff
@@ -5,7 +5,7 @@
 
 begin;
 -- Function takes application id and version number:
-create or replace function ggircs_portal.ciip_incentive(app_id int , version_no int )
+create or replace function ggircs_portal.ciip_incentive(app_id text , version_no text)
```
matthieu-foucault: That will do for now. Ultimately, this should be a computed field of `applicationRevision`

schema/deploy/ciip_functions/ciip_incentive.sql
```diff
@@ -42,29 +41,37 @@ returns setof ggircs_portal.ciip_incentive_by_product as $function$
...
      -- Loop over products
      for product in select * from ggircs_portal.ciip_production
-                    where version_number = version_no and application_id = app_id
+                    where version_number = version_no::integer and application_id = app_id::integer
      loop
 
+        -- If includes_imported_energy is false, heat/elec allocation factor=0 else set values from production view
+        em_alloc_elec = 0;
+        em_alloc_heat = 0;
+        if (select includes_imported_energy from ggircs_portal.benchmark b where b.product_id = product.product_id) then
```
matthieu-foucault: You should use the `benchmark_data` below, which also filters by reporting year

schema/deploy/views/ciip_production.sql
```diff
@@ -20,7 +20,9 @@ begin;
        (x.production_data ->> 'productUnits')::varchar(1000) as product_units,
        (x.production_data ->> 'productionAllocationFactor')::numeric as production_allocation_factor,
        (x.production_data ->> 'paymentAllocationFactor')::numeric as payment_allocation_factor,
-       (x.production_data ->> 'additionalData')::jsonb as additional_data
+       (x.production_data ->> 'additionalData')::jsonb as additional_data,
+       (x.production_data ->> 'importedElectricityAllocationFactor')::jsonb as imported_electricity_allocation_factor,
+       (x.production_data ->> 'importedHeatAllocationFactor')::jsonb as imported_heat_allocation_factor
```
matthieu-foucault: should these be cast as `numeric`?

dleard: yep!

schema/deploy/ciip_functions/ciip_incentive.sql
dleard: @matthieu-foucault  I'll go ahead and make it into a computed column

PR #407, opened 2020-02-19
Example: graphile-worker (for email?)

dleard:
This is an example branch for looking at graphile-worker.

- I have added a trigger & function to the deploy file for ciip_user & a tasks folder with one 'hello' task for the worker to run.
- I've also modified the "yarn dev" command to run the graphile-worker on startup.
- On update of the ciip_user table the graphile-worker will print a hello message with the new user's first_name and last_name from the new row.
- There are some details on doing different tasks like having the worker send email here.
https://github.com/graphile/worker#more-detail-on-scheduling-jobs-through-sql


dleard:
example no longer needed. Same code + mail stuff in #410

PR #410, opened 2020-02-20
Graphile worker + Nodemailer

dleard:
- Adds a 'mail' task that sends an email with nodemailer & graphile-worker
- Properly creates trigger function with sqitch

*The current smtp host `smtp.mailtrap.io` is a placeholder. If you want to play with the worker's email function go to https://mailtrap.io/ and create an account, then put your user/password details in the .env file as per .env.example


app/email_templates/welcome.js
```diff
@@ -0,0 +1,30 @@
+const createWelcomeMail = customVariables => {
+  console.log(customVariables);
```
wenzowski: let's chat later this week about loggers...

app/email_templates/welcome.js
```diff
@@ -0,0 +1,30 @@
+const createWelcomeMail = customVariables => {
```
wenzowski: as a pattern, let's make sure we destructure the vars object on the way in so we have an interface definition (or we can give it a type with ts)

app/tasks/hello.js
```diff
@@ -0,0 +1,8 @@
+const createWelcomeMail = require('../email_templates/welcome.js');
+const dotenv = require('dotenv');
+dotenv.config();
+
+module.exports = async (payload, helpers) => {
+  console.log(process.env);
+  helpers.logger.info(createWelcomeMail(payload));
+};
```
wenzowski: sweet, nice & clean

app/tasks/mail.js
```diff
@@ -0,0 +1,34 @@
...
+
+module.exports = async (payload, helpers) => {
+  const transporter = nodemailer.createTransport({
+    host: 'smtp.mailtrap.io',
+    port: 2525,
+    auth: {
+      user: process.env.MAIL_IO_USER,
+      pass: process.env.MAIL_IO_PASS
+    }
+  });
```
wenzowski: is there a way we can consolidate this into a single connection string? (new issue, new PR)

app/tasks/mail.js
```diff
@@ -0,0 +1,34 @@
+const nodemailer = require('nodemailer');
+const createWelcomeMail = require('../email_templates/welcome.js');
+const dotenv = require('dotenv');
+dotenv.config();
+
+module.exports = async (payload, helpers) => {
```
wenzowski: same thing with `payload`, if we can destructure or type that would make this more robust

PR #421, opened 2020-02-24
Application status email

dleard:
- create trigger on ApplicationRevisionStatus table watching for changes to column application_revision_status.
- update trigger function to accept email types (welcome, status_change, ...etc) to select different worker tasks for different triggers.
- add confirmation / amendment email templates
- update task 'mail' to 'sendMail' and handle the status_change email type


dleard:
Registration Email
Welcome

dleard:
Submission Confirmation: 
confirmation

dleard:
Application amended (status changed): 
amended

dleard:
Question: When getting a user via application -> facility -> organisation -> ciip_user_organisation -> user... There could be more than one user returned via that connection. Should we only notify the user who last updated the application?

app/server/index.js
```diff
@@ -48,6 +50,24 @@ if (process.env.PGPORT) {
...
+    taskDirectory: path.resolve(__dirname, 'tasks')
+  });
+}
+
+// Start graphile-worker
+worker().catch(error => {
+  if (error) {
+    throw error;
+  }
+});
```
matthieu-foucault:
```suggestion
if (!process.env.CI) {
  worker().catch(error => {
    if (error) {
      throw error;
    }
  });
}
```

app/server/tasks/sendMail.js
```diff
@@ -0,0 +1,78 @@
...
+  type,
+  email,
+  firstName,
+  lastName,
+  facilityName,
+  operatorName,
+  status
+}) => {
+  const transporter = nodemailer.createTransport(
+    process.env.SMTP_CONNECTION_STRING
```
matthieu-foucault: throw an error if `SMTP_CONNECTION_STRING` is not defined in `.env` ?

app/.env.example
```diff
@@ -1,5 +1,4 @@
 RELAY_SERVER=<Server>
 PORT=<Port>
 GRAPHQL_API_URL=<URL>
-MAIL_IO_USER=<Username>
-MAIL_IO_PASS=<Password>
+SMTP_CONNECTION_STRING=<connection string>
```
matthieu-foucault: We probably need to setup a config map in OpenShift for the `.env` file. @wenzowski can you confirm?

app/server/tasks/sendMail.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  if (htmlContent === null) {
+    return console.error(
+      'Invalid message type, no message could be generated. Email not sent'
+    );
+  }
+
+  const message = {
+    from: 'BCCAS <example@cas.com>',
```
matthieu-foucault: this should be in the `.env` file

app/server/index.js
matthieu-foucault: I suspect that the e2e tests are failing because `.env` is missing and requires the SMTP string. An alternative to disabling the worker is to add our mailtrap credentials and create that `.env` file before running the e2e tests. That could be in a future PR though.

app/.env.example
wenzowski: Either a ConfigMap or a Secret. Don't we already have a Secret? Checking...

wenzowski: (there are password here, so a secret would be better)

matthieu-foucault: #423 is adding a secret. BTW the detect-secrets pre-commit hook was being overzealous so I had to disable it there. I guess we can merge this one as-is, preventing it from being deployable, and I can add the secret after #423 is merged

wenzowski: Ah, so we have a secret explicitly for postgres but not a secret for app config. Let's add a new secret with some example data just like we do here in `.env.example` and have that secret managed outside the lifecycle of our deployment process (create it with dummy data only if it doesn't exist). We'll still need to bind the secret into our deploymentconfig same as we do for postgres https://github.com/bcgov/cas-ciip-portal/blob/develop/openshift/deploy/deploymentconfig/app.yml#L54

app/server/tasks/sendMail.js
wenzowski: In the scope of another PR, can we explore https://github.com/mozilla/node-convict please? That will give us config errors and environment-specific defaults for free. Happy to explore other libs as well if anyone has a fave.

app/server/index.js
wenzowski: Relates to environment-specific cascading config defaults; see note about mozilla/node-convict

PR #448, opened 2020-03-06
(hold)Investigate worker signal

dleard:
- adds 2 24 jobs 24 hours apart to test signal handling while running on openshif


dleard:
jobs are left in the jobsQueue table after a shutdown via signal & continued / completed upon re-start.

PR #459, opened 2020-03-12
SMTP testing: MailHog

dleard:
- Add documentation for starting a mailhog server
- Add NO_MAIL flag to yarn dev
- Add cypress test that tests automatic email send points
- Add Circle job for starting MailHog server
- Begin idea on re-writing fixtures that start with a clean database (truncate application table for each)

### DIFF
--- a/app/cypress/integration/email.spec.js
+++ b/app/cypress/integration/email.spec.js
@@ -1,9 +1,41 @@
-describe('The environment variables are not stupid', () => {
-  it('should have envs ', () => {
-    if (Cypress.env('NO_MAIL')) {
-      cy.log('Flag: NO_MAIL is set. Email tests did not run');
-    } else {
-      cy.log('mail uh oh');
-    }
+describe('When a user is created', () => {
+  before(() => {
+    cy.sqlFixture('fixtures/add-user');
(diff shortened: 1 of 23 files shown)


app/cypress/integration/email.spec.js
```diff
@@ -0,0 +1,146 @@
...
+      cy.get('.btn')
+        .contains('Send to Certifier')
+        .click();
+      cy.wait(1000);
+      cy.request('localhost:8025/api/v1/messages').then(response => {
+        // eslint-disable-next-line jest/valid-expect
+        expect(response.status).to.eq(200);
+        // eslint-disable-next-line jest/valid-expect
+        expect(response.body[0].To[0].Mailbox).to.contain('certifier');
+        // eslint-disable-next-line jest/valid-expect
```
matthieu-foucault: Why do you have to disable this rule?

dleard: @matthieu-foucault It doesn't like some expressions like `to.contain`, it complains it isn't valid, but this is what I found to work. I think it's using mocha/chai expressions but eslint is linting it for jest?

PR #539, opened 2020-04-13
Feature: enforce commit message style and generate changelogs

matthieu-foucault:
https://youtrack.button.is/issue/GGIRCS-867

Supersedes #106


matthieu-foucault:
Implementation in #106 was actually good to merge

PR #561, opened 2020-04-20
Feature: Validate BC Corporate Registry Number format

kriscooke:
- Implements custom format validation of BC Corporate Registry Number in the "Administration Data" section of industry reporter form:
  * 1 to 3 letters followed by 7 digits
  * `^[A-Za-z]{1,3}\d{7}$`
- Updates test snaphots
- Updates E2E test fixtures with correct number format
- Add integration test

### DIFF
--- a/app/cypress/integration/application-form-validation.spec.js
+++ b/app/cypress/integration/application-form-validation.spec.js
@@ -116,6 +116,11 @@ describe('When reviewing a submitted application as an analyst', () => {
     cy.get('#root_operator_duns +div .error-detail').contains(
       'DUNS number should be nine digits'
     );
+    cy.get(
+      '#root_operator_bcCorporateRegistryNumber +div .error-detail'
+    ).contains(
+      'BC Corporate Registry number should be 1-3 letters followed by 7 digits'
+    );
 
     cy.get(
       '#root_operationalRepresentative_mailingAddress_postalCode +div .error-detail'
--- a/app/tests/integration/Forms/__snapshots__/Form.test.tsx.snap
+++ b/app/tests/integration/Forms/__snapshots__/Form.test.tsx.snap
@@ -72,7 +72,7 @@ exports[`Form should match the snapshot with the administration form 1`] = `
                 readOnly={false}
                 required={true}
                 type="text"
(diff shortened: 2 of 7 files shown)


matthieu-foucault:
Re-running the CircleCI e2e test as I suspect that the issue is due to a flaky failure

matthieu-foucault:
Nevermind, they are real failures, the button to send the application is not showing up in these tests because the BC registry number in the test data does not follow the format.
I would expect an e2e test in `application-form-validation.spec.js` to fail as well.
More details coming up

matthieu-foucault:
Updating this fixture should fix the email e2e test; 
https://github.com/bcgov/cas-ciip-portal/blob/e304dc86d8674bda2f8c7004acc6bbb1ed6ee268/schema/data/fixtures/email-setup.sql#L49

matthieu-foucault:
And this should fix the page access test for the reporter: https://github.com/bcgov/cas-ciip-portal/blob/e304dc86d8674bda2f8c7004acc6bbb1ed6ee268/schema/data/fixtures/reporter-all-access-setup.sql#L55

matthieu-foucault:
Now, we need to actually test that the format of the BC registry number gets validated: 
Here you can see where invalid values for the DUNS and BC registry numbers are inserted
https://github.com/bcgov/cas-ciip-portal/blob/e304dc86d8674bda2f8c7004acc6bbb1ed6ee268/app/cypress/integration/application-form-validation.spec.js#L32-L37

And here the assertion for the DUNS format message. You just need to add the same for the BC registry number:
https://github.com/bcgov/cas-ciip-portal/blob/e304dc86d8674bda2f8c7004acc6bbb1ed6ee268/app/cypress/integration/application-form-validation.spec.js#L115-L118

This test can be improved to actually ensure that a valid input passes the validation (we're only checking invalid input here), but that should be done in a future PR, as we will be changing some of the design of how the validation gets triggered

matthieu-foucault:
@kriscooke actually, your branch is out of date with the `develop` branch and needs a rebase either way. You can do an interactive rebase and squash the commits yourself. Happy to walk you through it if you need.

PR #565, opened 2020-04-22
Incentive admin can edit reporting period dates

kriscooke:
[GGIRCS-1354](https://youtrack.button.is/issue/GGIRCS-1354)
- [x] Incentive admin can view reporting period details
- [x] Incentive admin can edit reporting period details
  * Reporting periods that are complete\* cannot be edited (Edit button does not show)
  * \* assumption: 'complete' is defined by application response date having passed
  * Application dates that are in the past (such as the open date) for an open reporting period are disabled in the form.
  * Validation ensures past dates are not chosen and that open, close, and response dates occur in chronological order
- [x] Add new reporting period (see bugs)
  * Validation ensures chronological order of reporting period start/end dates, and that the reporting year entered is in range of the period start and end. However, there is nothing stopping the user from selecting past dates for the reporting period.
  * Includes the same validation of application dates described above
  * The only validated relationship between reporting period dates and application dates is that the application close should not occur before reporting period end (cannot accurately report emissions that haven't happened yet).
- [x] Snapshot tests

#### Some wider changes introduced in the course of this feature:
- Form styles have been extracted from `app/containers/Forms/Form` into a shared style module in `app/containers/Forms/FormSharedStyles.tsx`.
- Errors and help text that appear at the top of a form (in `FormFieldTemplate`) are now wrapped in a full-width div to prevent a layout bug when custom validation errors suddenly appear.
- Due to issues with `react-jsonschema-form`'s built-in `alt-date` and `alt-datetime` widgets (incompatibility with Bootstrap 4, and UTC format), [AltDateWidget and AltDateTimeWidget](https://github.com/rjsf-team/react-jsonschema-form/tree/v1.8.1/src/components/widgets) have been brought in as customized widgets. It would be nice to eventually come up with a better solution.


app/containers/Admin/ReportingYearTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+import {graphql, createFragmentContainer} from 'react-relay';
+import {ReportingYearTable_query} from '__generated__/ReportingYearTable_query.graphql';
+
+interface Props {
+  query: ReportingYearTable_query;
+}
+
+function formatDate(date) {
+  const d = new Date(date);
+  return d.toLocaleString();
```
matthieu-foucault, line 12: We are using the `moment-timezone` library for rendering of dates.
For example;
https://github.com/bcgov/cas-ciip-portal/blob/4ec58fc2e97ee12a81d4d4544ce38713bfb0e03c/app/pages/index.tsx#L39-L45

app/containers/Admin/ReportingYear/NewReportingYearFormDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </Modal.Header>
+      <Modal.Body>
+        <Container>
+          <JsonSchemaForm
+            omitExtraData
+            liveOmit
+            noHtml5Validate
+            schema={newReportingYearSchema.schema as JSONSchema6}
+            uiSchema={newReportingYearSchema.uiSchema}
+            formData={{}}
```
matthieu-foucault, line 43: No need to pass form data here

app/containers/Forms/Form.tsx
```diff
@@ -149,40 +150,7 @@ export const FormComponent: React.FunctionComponent<Props> = ({
...
-            padding: 0;
-            list-style: none;
-          }
-          ul.error-detail li.text-danger {
-            background: #ce5c5c;
-            color: white !important;
-            padding: 5px;
-          }
-        `}
+        {globalFormStyles}
```
kriscooke, line 153: These have been extracted into a [style module](https://github.com/bcgov/cas-ciip-portal/pull/565/files#diff-fb2e8a7bb50109eb53c0bcc780d27259) in order to share the form styles between the industry reporter form and other forms (in this case, the `/admin/reporting-years` form).

app/containers/Forms/FormFieldTemplate.tsx
```diff
@@ -44,8 +44,8 @@ const FormFieldTemplate: React.FunctionComponent<FieldTemplateProps> = ({
       <Col xs={12} className={classNames}>
         <Form.Row>
           {description}
-          {errors}
-          {helpText}
+          <div className="col-12">{errors}</div>
+          <div className="col-12">{helpText}</div>
```
kriscooke, line 48: Wrapping the errors and helptext (which are shown at the the top of the form) in a full-width container prevents a layout bug where the appearance of errors on validation suddenly squishes the first form field to the side. This way there's a full-width spot preserved for errors even before there are any errors to show.

app/containers/Forms/Form.tsx
matthieu-foucault: feel free to update the snapshots ( `yarn test -u`)

kriscooke: PS: Here's the [link to the docs](https://github.com/zeit/styled-jsx#external-css-and-styles-outside-of-the-component) that you were wondering about for creating a style module.

app/components/Forms/AltDateInput.tsx
```diff
@@ -0,0 +1,352 @@
+// eslint-disable
```
kriscooke, line 1: If we don't figure out some way to properly disable eslint here, it turns out there is a laundry list of fails that must be fixed manually, some of which seem to trigger a chain of required updates:
```
  components/Forms/AltDateInput.tsx:1:0
  ⚠  110:5   Unexpected todo comment.                                                                                                                                                                                                                                                                                      no-warning-comments
  ⚠  292:1   Function deepEquals has a complexity of 30. Maximum allowed is 20.                                                                                                                                                                                                                                            complexity
  ✖    1:0   Specify the rules you want to disable.                                                                                                                                                                                                                                                                        unicorn/no-abusive-eslint-disable
  ✖   52:14  Operands of + operation must either be both strings or both numbers. Consider using a template literal.                                                                                                                                                                                                       @typescript-eslint/restrict-plus-operands
  ✖   52:14  Operands of + operation must either be both strings or both numbers. Consider using a template literal.                                                                                                                                                                                                       @typescript-eslint/restrict-plus-operands
  ✖   98:5   State initialization should be in a class property                                                                                                                                                                                                                                                            react/state-in-constructor
  ✖  101:3   componentWillReceiveProps is deprecated since React 16.9.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops. Use https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles to automatically update your components.  react/no-deprecated
  ✖  184:24  Do not use Array index in keys                                                                                                                                                                                                                                                                                react/no-array-index-key
  ✖  198:11  Unexpected negated condition.                                                                                                                                                                                                                                                                                 no-negated-condition
  ✖  207:11  Unexpected negated condition.                                                                                                                                                                                                                                                                                 no-negated-condition
  ✖  250:21  Operands of + operation must either be both strings or both numbers. Consider using a template literal.                                                                                                                                                                                                       @typescript-eslint/restrict-plus-operands
  ✖  265:3   Parameter time is declared but its value is never read.                                                                                                                                                                                                                                                       @typescript-eslint/no-unused-vars-experimental
  ✖  362:3   Require compare argument.                                                                                                                                                                                                                                                                                     @typescript-eslint/require-array-sort-compare
  ✖  363:3   Require compare argument.                                                                                                                                                                                                                                                                                     @typescript-eslint/require-array-sort-compare
```

For example if I tackle `componentWillReceiveProps` by changing it to `UNSAFE_componentWillReceiveProps`, when I run xo again it then says: 
```
"UNSAFE_componentWillReceiveProps is unsafe for use in async rendering. Update the component to use getDerivedStateFromProps instead. See https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html"
```
...And then running xo again tells me:
```
UNSAFE_componentWillReceiveProps is unsafe for use in async rendering. Update the component to use getDerivedStateFromProps instead. See https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html.  react/no-unsafe
```
which isn't exactly a straightforward update.

matthieu-foucault: Let me look into disabling eslint

kriscooke: For reference, [this was the PR](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/308) that led me to believe we should be able to do something like: 
`// eslint-disable unicorn/no-abusive-eslint-disable`
or
`// eslint-disable unicorn/no-abusive-eslint-disable react/no-deprecated`

kriscooke: I can report marginal success using: `/* eslint-disable react/no-deprecated */` where you only list one single rule that you're unwilling to fix

matthieu-foucault: If that works then let's go with this, I'll add a card in the issue tracker about using a better date/time input

PR #566, opened 2020-04-22
Render archived products

dleard:


dleard:
@matthieu-foucault Our integration test for Forms was broken on develop for products/fuel. It's not getting the changes to the schema that is done in ProductRowIdField that are then passed to the SearchDropDownWidget. I've mocked the widget so that it just returns a string, which allows all the data to be rendered, but maybe there is a better way to mock it so that it actually does what it is supposed to?

app/components/SearchDropdown.tsx
```diff
@@ -13,16 +12,30 @@ interface Props {
...
+    });
+    return indexArray;
+  };
+
+  // Remove all undefined indexes from options (required so archived/deprecated products do not show as options in the dropdown)
+  const undefinedIndexes = getAllUndefinedIndexes(options, undefined);
+  for (let i = undefinedIndexes.length - 1; i >= 0; i--) {
+    options.splice(undefinedIndexes[i], 1);
+  }
+
```
matthieu-foucault, line 38: Props should be immutable, `options.splice` breaks that.
This could whole thing could also be written as `options.filter(o => o !== undefined)`.
Either way I don't think that handling some product specific things in here is the right solution.

app/tests/integration/Forms/Form.test.tsx
```diff
@@ -13,6 +13,25 @@ import {FormJson} from 'next-env';
...
+});
+
+// Todo: Mock so that the entire flow works properly
+
+// jest.mock('components/Forms/SearchDropdownWidget', () => ({onChange,
+//   schema,
+//   id,
+//   placeholder,
+//   value}) => {return <mock-SearchDropdownWidget schema={{...schema, enums: [1,2], enumNames: ['one', 'two'], state: ['active', 'active']}} />}
+// );
```
matthieu-foucault, line 34: Does this belong in this PR? Please make a separate PR to fix the tests if it doesn't

app/cypress/integration/application-production.spec.js
```diff
@@ -50,4 +50,19 @@ describe('The production tab', () => {
...
+      .clear()
+      .type('Aluminum');
+    cy.get('.dropdown-item').click();
+    cy.sqlFixture('fixtures/application-production-archive-product');
+    cy.visit(
+      `/reporter/ciip-application?formResultId=${formResultId}&applicationId=${applicationId}&version=1`
+    );
+    cy.get('#page-content');
+    cy.contains('Warning');
+  });
```
matthieu-foucault, line 67: Do we need this e2e test? You already have the unit test below

app/containers/Forms/ProductRowIdField.tsx
```diff
@@ -24,7 +24,8 @@ export const ProductRowIdFieldComponent: React.FunctionComponent<Props> = props
       schema: {
         ...props.schema,
         enum: props.query.allProducts.edges.map(({node}) => node.rowId),
-        enumNames: props.query.allProducts.edges.map(({node}) => node.name)
+        enumNames: props.query.allProducts.edges.map(({node}) => node.name),
+        state: props.query.allProducts.edges.map(({node}) => node.state)
```
matthieu-foucault, line 28: `state` is not part of the json schema definition, and should not be added here. 
Perhaps if the `formData` prop is not the id of an active product, this component could simply render a disabled text input instead of the `StringField`.
This way you woudn't have to tweak how the search dropdown works

app/cypress/integration/application-production.spec.js
dleard: yeah probably not. I'll pull it out

app/tests/integration/Forms/Form.test.tsx
dleard: It fixes the production & fuel tests returning nothing in the snapshot, so kind of? You can see the effect in the snapshot diff, but the issue was not indtroduced in this PR.

app/containers/Forms/ProductRowIdField.tsx
```diff
@@ -23,25 +24,69 @@ export const ProductRowIdFieldComponent: React.FunctionComponent<Props> = props
...
-  return <props.registry.fields.StringField {...fieldProps} />;
+  const activeProductIds = props.query.active.edges.map(({node}) => node.rowId);
+  const inactiveProductIds = [
+    ...props.query.archived.edges.map(({node}) => node.rowId),
+    ...props.query.deprecated.edges.map(({node}) => node.rowId)
+  ];
+  const inactiveProductNames = [
+    ...props.query.archived.edges.map(({node}) => node.name),
+    ...props.query.deprecated.edges.map(({node}) => node.name)
+  ];
```
matthieu-foucault, line 43: You should make use of `useMemo` to avoid iterating on all the product lists on every render

matthieu-foucault: Also do you mind updating that component's documentation above while you're at it. It's an out-of-data copy-paste, so it's not accurate at all

PR #571, opened 2020-04-24
Fix Storybook

dleard:
Storybook was missing a dev-dependency: babel-preset-react-app.


dleard:
@kriscooke fixed this. No need for this duplicate PR.

PR #588, opened 2020-04-28
Custom "alt-date" and "alt-datetime" type inputs

kriscooke:
For use in react-jsonschema-form in #565 . The components have been pasted directly from the corresponding JSX components in `rjsf-team/react-jsonschema-form` - but with the minor difference of an added Bootstrap 4 class to child elements of inline lists.

Needs help being converted to Typescript (interpreting and fixing TS errors).


app/components/Forms/AltDateTimeInput.tsx
```diff
@@ -3,8 +3,9 @@
 import React from "react";
 import PropTypes from "prop-types";
 import AltDateInput from "./AltDateInput";
+import { WidgetProps } from "react-jsonschema-form";
 
-function AltDateTimeInput(props) {
+function AltDateTimeInput(props: WidgetProps) {
```
matthieu-foucault, line 8: We've been using `const AltDateTimeInput: React.FunctionComponent<WidgetProps> = (props) =>` to ensure that the return type is also correct

app/components/Forms/AltDateInput.tsx
```diff
@@ -58,7 +59,7 @@ class AltDateInput extends Component {
     },
   };
 
-  constructor(props) {
+  constructor(props: WidgetProps) {
```
matthieu-foucault, line 62: Similar to the comment below, a component declared as `const AltTimeInput: React.FunctionComponent<Props> = (props) =>` with the following interface 
```ts
interface Props extends WidgetProps {
  time: boolean;
}
```
would provide better type checking.

I also see another component above, in the same file, which is discouraged.

Instead of using a state, you can use the `useMemo` hook

PR #589, opened 2020-04-28
Fix: incorrect review status

dleard:
The computed column `application_application_revision_status` was pulling the latest status for the application, where it should have been getting the status for the revision being viewed.
Resolved by adding a versionNumberInput variable to the computed column.


matthieu-foucault:
@dleard  looks like the SonarCloud Code Analysis check is having issues. Can you try squashing your commits and push again to see if that triggers it

matthieu-foucault:
(or just rebase on the latest `develop`)

PR #595, opened 2020-04-29
Fix: Do not allow status of non-current application revisions to change

dleard:
We were allowing users to change the status of old application versions which was causing some bugs & possibly allowing for 2 approved application revisions of the same application.
A fix for a related status-display bug was introduced in #592 but was not the right/full solution.
This PR:
- Adds a trigger that protects any non-current versions of an application from having the status changed
- Adds a conditional when reviewing an application that does not render the option to change the status if it is not the current version (this is where the #592 bug was found)
- Adds jest & pgtap tests to check the front end & trigger behave as expected

Bug card related to #592 is here: https://youtrack.button.is/issue/GGIRCS-1430


schema/deploy/trigger_functions/check_for_immutable_status.sql
```diff
@@ -0,0 +1,20 @@
+-- Deploy ggircs-portal:trigger_functions/check_for_immutable_status to pg
+-- requires: schema_ggircs_portal_private
+
+begin;
+
+create or replace function ggircs_portal_private.check_for_immutable_status()
```
matthieu-foucault, line 6: The name of the function could be more precise. It's not obvious that it's related to application revisions

PR #621, opened 2020-05-04
feat(product): Products refactor (back-end)

dleard:
Products table:
   - rename columns that abused reserved words (name description, state -> product_name, 
     product_description, product_state)
   - remove unnecessary columns
   - add unique partial index
   - add trigger to protect read-only products
   - create enum for product_state (DRAFT, PUBLISHED, ARCHIVED)

Other: 
- Add  / update pgTap tests
- Remove / update product-related mutations
- Refactor fragments / column name references that have changed
- Refactor functions that used changed/deleted mutations

** I have disabled the products-benchmarks cypress test until the end of the feature as there is no reason to continually update it until the feature is near completion**


app/containers/Forms/ProductField.tsx
```diff
@@ -27,11 +27,11 @@ export const ProductFieldComponent: React.FunctionComponent<Props> = (
...
   const {formData, query, onChange} = props;
 
-  const productIsActive = (formData, query) => {
+  const productIsPublished = (formData, query) => {
     const product = query.allProducts.edges.find(
       ({node}) => node.rowId === formData.productRowId
     )?.node;
-    if (product?.state === 'active' || !product) return true;
+    if (product?.productState === 'published' || !product) return true;
     return false;
```
matthieu-foucault, line 35: You could write this like:
```js
return query.allProducts.edges.some(
  ({node}) => node.rowId === formData.productRowId && node.state === 'active'
);
```

app/containers/Forms/ProductRowIdField.tsx
```diff
@@ -29,47 +29,51 @@ export const ProductRowIdFieldComponent: React.FunctionComponent<Props> = (
...
     () => ({
-      activeProductIds: props.query.active.edges.map(({node}) => node.rowId),
-      inactiveProductIds: [
+      publishedProductIds: props.query.published.edges.map(
+        ({node}) => node.rowId
+      ),
+      unpublishedProductIds: [
         ...props.query.archived.edges.map(({node}) => node.rowId),
-        ...props.query.deprecated.edges.map(({node}) => node.rowId)
+        ...props.query.draft.edges.map(({node}) => node.rowId)
```
matthieu-foucault, line 50: draft products are not needed here, they will never be in an application (can't go back to "draft" status)

app/containers/Forms/ProductRowIdField.tsx
```diff
@@ -29,47 +29,51 @@ export const ProductRowIdFieldComponent: React.FunctionComponent<Props> = (
...
       ],
-      inactiveProductNames: [
-        ...props.query.archived.edges.map(({node}) => node.name),
-        ...props.query.deprecated.edges.map(({node}) => node.name)
+      unpublishedProductNames: [
```
matthieu-foucault, line 52: rename to `archivedProductNames`

app/containers/Forms/ProductRowIdField.tsx
```diff
@@ -80,31 +84,31 @@ export const ProductRowIdFieldComponent: React.FunctionComponent<Props> = (
...
+      draft: allProducts(condition: {productState: DRAFT}) {
         edges {
           node {
             rowId
-            name
-            state
+            productName
+            productState
           }
         }
```
matthieu-foucault, line 113: Same comment as above, no need for draft products

app/containers/Products/ProductCreatorContainer.tsx
```diff
@@ -26,11 +27,10 @@ export const ProductCreator: React.FunctionComponent<Props> = ({
...
       input: {
         product: {
-          name: e.formData.name,
-          description: e.formData.description,
+          productName: e.formData.name,
+          productDescription: e.formData.description,
           units: e.formData.units,
-          state: 'active',
-          parent: [null],
+          productState: 'DRAFT' as CiipProductState,
```
matthieu-foucault, line 33: is this cast needed? maybe use `CiipProductState.DRAFT` instead (I forgot how enums work)

app/containers/Products/ProductCreatorContainer.tsx
```diff
@@ -26,11 +27,10 @@ export const ProductCreator: React.FunctionComponent<Props> = ({
     const variables = {
       input: {
         product: {
-          name: e.formData.name,
-          description: e.formData.description,
+          productName: e.formData.name,
+          productDescription: e.formData.description,
```
matthieu-foucault, line 31: no more product description

app/containers/Products/ProductListContainer.tsx
```diff
@@ -44,12 +44,12 @@ export const ProductList: React.FunctionComponent<Props> = ({
     const allProducts = query.searchProducts.edges;
 
     const displayNameToColumnNameMap = {
-      Product: 'name',
-      Description: 'description',
+      Product: 'product_name',
+      Description: 'product_description',
```
matthieu-foucault, line 48: remove product description

app/containers/Products/ProductRowItemContainer.tsx
```diff
@@ -158,66 +159,50 @@ export const ProductRowItemComponent: React.FunctionComponent<Props> = ({
...
-        newSubtractExportedHeatEmissions: product.subtractExportedHeatEmissions,
-        newSubtractGeneratedElectricityEmissions:
-          product.subtractGeneratedElectricityEmissions,
-        newSubtractGeneratedHeatEmissions:
-          product.subtractGeneratedHeatEmissions,
-        newRequiresProductAmount: product.requiresProductAmount,
-        newAddEmissionsFromEios: product.addEmissionsFromEios
+        id: product.id,
+        productPatch: {
+          productState: 'ARCHIVED' as CiipProductState
```
matthieu-foucault, line 168: see comment re. enum usage

schema/data/fixtures/products-benchmarks-setup.sql
```diff
@@ -8,10 +8,10 @@ delete from ggircs_portal.product;
 
 insert into ggircs_portal.product(
   id,
-  name,
-  description,
+  product_name,
+  product_description,
```
matthieu-foucault, line 12: remove product description

schema/deploy/tables/product.sql
```diff
@@ -5,11 +5,10 @@ begin;
 
 create table ggircs_portal.product (
   id integer primary key generated always as identity,
-  name varchar(1000) not null,
-  description varchar(10000),
+  product_name varchar(1000) not null,
+  product_description varchar(10000),
```
matthieu-foucault, line 9: remove product description

schema/deploy/tables/product.sql
```diff
@@ -5,11 +5,10 @@ begin;
...
   units varchar(1000),
-  state varchar(1000),
-  parent integer ARRAY,
+  product_state ggircs_portal.ciip_product_state default 'draft',
```
matthieu-foucault, line 11: add `not null` to prevent explicitly setting it to `null`

app/containers/Products/ProductCreatorContainer.tsx
dleard: `CiipProductState` is being exported from createProductMutation.graphql.ts like this: `export type CiipProductState = "ARCHIVED" | "DRAFT" | "PUBLISHED";` Is accessing the enum properties still possible in this way if it's being exported as a TYPE instead of an ENUM? 
It looks here like it can only be used as a type context: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

app/containers/Forms/ProductField.tsx
dleard: doesn't `.some` return true if any item in the array matches the requirements? So this would return true if any node matched an id and had an 'active' state? ie:
[
{rowId: 1, state: 'active'}, rowId = productRowId = 1 && state = 'active' => true 
{rowId: 2, state: 'archived'}, rowId = productRowId = 2 && state = 'archived' => false
{rowId: 2, state: 'archived'} => rowId = productRowId = 3 && state = 'archived' => false
]
.some on the above array would return: [true, false, false] === true?

app/containers/Products/ProductCreatorContainer.tsx
dleard: it looks like we've been casting other enums like CiipApplicationRevisionStatus. (app/components/SubmitApplication.tsx)

app/containers/Forms/ProductField.tsx
matthieu-foucault: `productRowId` is a parameter, so only one node would match the product id

dleard: changing it still appears to have a different effect than how it's currently written. can we leave it as is?

matthieu-foucault: sure, just put a TODO comment

PR #628, opened 2020-05-08
feat(certification): Optionally email certifier

dleard:
- Add checkbox to optionally notify certifier by email
- Updates certification_url table with boolean to indicate whether to send an email
- Updates trigger function to not send email on boolean = false
- Create pgTap tests to test trigger function
- Update jest/cypress tests

### DIFF
--- a/schema/deploy/tables/certification_url.sql
+++ b/schema/deploy/tables/certification_url.sql
@@ -18,7 +18,8 @@ create table ggircs_portal.certification_url (
   updated_by int references ggircs_portal.ciip_user,
   deleted_at timestamp with time zone,
   deleted_by int references ggircs_portal.ciip_user,
-  certification_request_sent_to varchar(1000),
+  send_certification_request boolean not null default false,
+  certifier_email varchar(1000),
   certification_request_sent_at timestamp with time zone,
   recertification_request_sent boolean default false,
   -- TODO(Dylan): revisit expiry / deprecation of generated URLs in the context of Authorization
@@ -59,7 +60,7 @@ create trigger _set_user_id
   execute procedure ggircs_portal.set_user_id('certification_url');
 
 create trigger _certification_request_email
-  before update of certification_request_sent_to on ggircs_portal.certification_url
+  before update of certifier_email on ggircs_portal.certification_url
   for each row
   execute procedure ggircs_portal_private.run_graphile_worker_job('certification_request');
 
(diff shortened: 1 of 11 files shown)


schema/deploy/tables/certification_url.sql
```diff
@@ -18,7 +18,8 @@ create table ggircs_portal.certification_url (
   updated_by int references ggircs_portal.ciip_user,
   deleted_at timestamp with time zone,
   deleted_by int references ggircs_portal.ciip_user,
-  certification_request_sent_to varchar(1000),
+  send_certification_request boolean not null default false,
```
wenzowski, line 21: UX, maybe already addressed: if this is false by default how do we ensure users know that this critical email won't be sent unless the toggle is clicked?

Out of scope on this PR but worth closing the loop with @rajpersram maybe?

schema/deploy/trigger_functions/run_graphile_worker_job.sql
```diff
@@ -70,35 +70,38 @@ begin
...
-              'firstName', application_details.first_name,
-              'lastName', application_details.last_name,
-              'email', new.certification_request_sent_to,
-              'facilityName', application_details.facility_name,
-              'operatorName', application_details.operator_name));
-
-      -- timestamp when the email was sent
-      new.certification_request_sent_at = now();
-      return new;
+      if (new.send_certification_request = true) then
```
wenzowski, line 73: I'd love to see `ggircs_portal_private.run_graphile_worker_job` split so instead of switching `when tg_argv[0] = 'certification_request'` each of those blocks is contained within a function. This function is really scheduling jobs rather than running them, right? Tech debt for later I think (should add to tracker)

schema/deploy/tables/certification_url.sql
dleard: Maybe the default should be true and the toggle should begin as clicked? I'll make a task in youtrack to visit the idea with @rajpersram

schema/deploy/trigger_functions/run_graphile_worker_job.sql
dleard: I'll make an unscheduled tech debt task for this in youtrack

schema/deploy/tables/certification_url.sql
dleard: https://youtrack.button.is/issue/GGIRCS-1492

schema/deploy/trigger_functions/run_graphile_worker_job.sql
dleard: https://youtrack.button.is/issue/GGIRCS-1493

PR #647, opened 2020-05-13
List all certification requests for current user by email

kriscooke:
- Adds a computed column that lists all non-expired certification requests directed to the current user (as determined by their email address).
- Adds a page at /certifer/requests` listing these requests where the certifier can click through to certify individual applications.
[GGIRCS-1393](https://youtrack.button.is/issue/GGIRCS-1393)

### DIFF
--- a/app/pages/certifier/requests.tsx
+++ b/app/pages/certifier/requests.tsx
@@ -1,0 +1,43 @@
+import React, {Component} from 'react';
+import {graphql} from 'react-relay';
+import {NextRouter} from 'next/router';
+import {CiipPageComponentProps} from 'next-env';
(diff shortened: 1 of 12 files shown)


app/pages/certifier/requests.tsx
```diff
@@ -0,0 +1,45 @@
...
+    }
+  `;
+
+  render() {
+    const {query} = this.props;
+    return (
+      <DefaultLayout
+        title="Certification Requests"
+        session={query.session}
+        needsUser={false}
```
dleard, line 37: Remove the needsUser & needsSession props. They default to true in DefaultLayout & should stay true as the user should be logged in to view this page.

PR #650, opened 2020-05-14
feat(product): Energy products are read only

dleard:
- Adds column is_read_only to product table
- Updates protect_read_only_products trigger function
- Updates ProductRowItemContainer to use isReadOnly column

** The title of the youtrack card is 'energy products should not be archivable'. I have assumed that this includes these products that affect the boolean options when creating or editing a product,:
- Exported Electricity / Heat
- Purchased Electricity / Heat
- Generated Electricity  / Heat
- Emissions from EIOs

But does not include other "energy" products that are not referenced by the create/edit options like:
- District energy (heat)


schema/data/prod/energy_product.sql
```diff
@@ -0,0 +1,33 @@
+begin;
+
+alter table ggircs_portal.product disable trigger _protect_read_only_products;
+
+with rows as (
+insert into ggircs_portal.product(id, product_name, units, product_state, requires_emission_allocation, requires_product_amount, is_read_only)
+overriding system value
+values
+(35, 'Exported electricity', 'MWh', 'published' ,true, true, true),
```
dleard, line 9: @matthieu-foucault should I have done this the other way around? With these energy products starting at id=1 and the test products being inserted after?

matthieu-foucault: @dleard yup :upside_down_face:

dleard: ok yeah that makes more sense, I've swapped them.

PR #671, opened 2020-05-20
Fix failing audit CI step

dleard:
- Fixes https://www.npmjs.com/advisories/1219
- I may have done a bad bad thing, or at least this is worth watching as no new version is currently available that is unaffected. So I have rolled this package back to the last unaffected version, which is a beta from a year ago.

This package is only used by storybook

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -243,7 +243,8 @@
     "**/acorn": "7.1.1",
     "jest/**/kind-of": "6.0.3",
     "**/minimist": "1.2.5",
-    "**/yargs-parser": "18.1.2"
+    "**/yargs-parser": "18.1.2",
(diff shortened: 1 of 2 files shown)


matthieu-foucault:
@dleard I think that the version you chose is technically affected, but wasn't analyzed by npm as it's a beta. I expect a patch would be released soon given the impact

matthieu-foucault:
see https://github.com/probablyup/markdown-to-jsx/issues/306

dleard:
@matthieu-foucault So is rolling back to this un-analyzed version in order to continue to merge things fine until there is a patch?

matthieu-foucault:
Yeah. I subscribed to the issue linked above to see how things go

PR #672, opened 2020-05-20
Feat(keycloak): dev test prod keycloak

dleard:
- Pulls the ${OC_PROJECT} variable from openshift/deploy//deploymentconfig/app.yml & maps it to a namespace
- Uses that mapping to dynamically create auth-server-url in the keycloak config
- keycloak config is now created in index.js (instead of keycloak.json) and passed as an argument to the instantiation
- Adds the current export files for dev/test/prod to the ciip_portal repo


openshift/deploy/deploymentconfig/app.yml
```diff
@@ -70,6 +70,8 @@ objects:
             value: https://ciip-portal-${OC_PROJECT}-feedback.pathfinder.gov.bc.ca
           - name: HOST
             value: https://ciip-portal-${OC_PROJECT}.pathfinder.gov.bc.ca
+          - name: PROJECT
```
matthieu-foucault, line 73: can you rename the env variable to `NAMESPACE`? "Project" is the openshift term, because they like to make things confusing, but if our app moves to k8s one day then I'd rather use the k8s nomenclature

PR #673, opened 2020-05-20
Link to batch certify page from dash and email

kriscooke:
- Adds link to certification requests list page (`/certifier/certify`):
  * on the reporter dash
  * in the request email to certifier
- Adds navigation of email and dash links to end-to-end tests

[GGIRCS-1394](https://youtrack.button.is/issue/GGIRCS-1394)


app/tests/unit/containers/Organisation/Organisations.test.tsx
```diff
@@ -31,31 +55,11 @@ describe('Organisations', () => {
...
-                }
-              }
-            ]
-          }
-        }
-      },
-      allOrganisations: {
-        edges: []
-      }
-    };
```
kriscooke, line 54: Just moved the mock queries to the top of the page so I could re-use them in my new test

PR #676, opened 2020-05-22
Fix yarn cache

kriscooke:
Ensure Yarn checks files on install. @matthieu-foucault @wenzowski  Did I understand the proposed fix correctly here?


kriscooke:
Update: closing as this doesn't work locally.

PR #667, opened 2020-05-19
Finalize landing page copy and key dates

kriscooke:
- Updates test site's current_timestamp mock to reflect 2019 reporting year
- Adds a new column, `swrs_deadline` to reporting year table for the purpose of showing key program dates on landing page
- Removes unused `applicationResponseTime` attribute from `reporting_year` table (rather, `swrs_deadline` replaces it)
- Landing page copy updates as in [GGIRCS-570](https://youtrack.button.is/issue/GGIRCS-570)

PR #678, opened 2020-05-22
Update additional date in landing page copy

kriscooke:
- Updates the application close date in another place on the landing page, just above Register and Apply (missed in guidance doc for #667 aka. [GGIRCS-570](https://youtrack.button.is/issue/GGIRCS-570)). This date is no longer hard-coded in the copy.
- Adds a snapshot for the login-redirect page, where the above "Register and apply" component is also included.

### DIFF
--- a/app/tests/unit/pages/login-redirect.test.tsx
+++ b/app/tests/unit/pages/login-redirect.test.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import {shallow} from 'enzyme';
+import LoginRedirect from 'pages/login-redirect';
+import {loginRedirectQueryResponse} from 'loginRedirectQuery.graphql';
+
+const query: loginRedirectQueryResponse['query'] = {
+  session: null,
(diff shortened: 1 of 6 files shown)


app/tests/unit/pages/login-redirect.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  it('It passes an application deadline to the RegistrationLoginButtons component', () => {
+    const wrapper = shallow(<LoginRedirect query={query} />);
+    expect(
+      wrapper
+        .find('RegistrationLoginButtons')
+        .first()
+        .prop('applicationDeadline')
+    ).toBe('August 31, 2020');
+  });
+});
```
kriscooke, line 30: In the course of this copy (date) update, I realized there was no test for this page. Added one here, as a test would have alerted me earlier to a failing build (I'd been unaware the RegistrationLoginButtons component I updated for the landing page is also present on this page).

PR #704, opened 2020-05-26
Finalize copy on certifier pages

kriscooke:
- Updates copy on certification redirect page per [GGIRCS-1516](https://youtrack.button.is/issue/GGIRCS-1516)
  * Adds unit test for this page
  * Fixes URL leading to the certification page (bug) - necessary for the test to pass
- Updates certification page per [GGIRCS-1517](https://youtrack.button.is/issue/GGIRCS-1517)
  * Removes disclaimer checklist in favor of showing the signature interface to start with
  * Updates copy using a reusable text component that's now shared with the `/resources/disclaimer` page
  * "expand" the lengthy text in scroll area by linking out in new tab to the `/resources/disclaimer` page (same text) which is no longer empty.


kriscooke:
[`reporter-certifier-access-spec`](https://github.com/bcgov/cas-ciip-portal/blob/feat/certifier-copy/app/cypress/integration/reporter-certifier-access.spec.js#L17-L19) is failing because the query I added, `ciipUserByCreatedBy` is coming back null - but only in the E2E test. 

The fixture `certification_url` did not originally include a `created_by` so I updated [the fixture](https://github.com/bcgov/cas-ciip-portal/blob/feat/certifier-copy/schema/data/fixtures/certifier-access-setup.sql#L368) with a `created_by` attribute (references the cypress test reporter user), but this doesn't fix it.  I also tried inserting a new user in the test setup just above the certification_url insert (and using that for created_by), to no avail. 🤔

Can someone see what I could be missing with the test environment?

PR #709, opened 2020-05-28
Fix: requires_emission_allocation sanity check

dleard:
- Adds a database exception in application_revision_ciip_incentive if:
  - 2 products are reported where requires_emission_allocation = false AND
  - Both products are non-energy products
- Adds front-end custom validation that will throw an error in the production form for the same conditions.

### DIFF
--- a/app/containers/Forms/Form.tsx
+++ b/app/containers/Forms/Form.tsx
@@ -98,6 +98,31 @@ export const FormComponent: React.FunctionComponent<Props> = ({
       return result?.requiresEmissionAllocation === true;
     });
 
+  const customValidation = (formData, errors) => {
+    let hasFalseRequiresEmissionAllocation = false;
+    let nonEnergyProductCount = 0;
+    const productsInConflict = [];
+    if (formData[0]?.productRowId) {
+      formData.forEach((product, index: number) => {
+        if (product.requiresEmissionAllocation === false)
+          hasFalseRequiresEmissionAllocation = true;
+        if (product.productRowId > 7) {
+          nonEnergyProductCount++;
+          productsInConflict.push(index + 1);
+        }
+      });
+    }
+
+    if (hasFalseRequiresEmissionAllocation && nonEnergyProductCount > 1)
+      errors['0'].addError(
+        `More than one non-energy product that does not require allocation of emissions cannot be reported together. -- Products in conflict: ${productsInConflict.join(
+          ','
+        )}`
+      );
+
+    return errors;
+  };
+
   const formClass = uiSchema?.['ui:className'] || '';
   return (
     <div className={formClass}>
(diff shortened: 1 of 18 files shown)


app/containers/Forms/Form.tsx
```diff
@@ -98,6 +98,31 @@ export const FormComponent: React.FunctionComponent<Props> = ({
...
 
+  const customValidation = (formData, errors) => {
+    let hasFalseRequiresEmissionAllocation = false;
+    let nonEnergyProductCount = 0;
+    const productsInConflict = [];
+    if (formData[0]?.productRowId) {
+      formData.forEach((product, index: number) => {
+        if (product.requiresEmissionAllocation === false)
+          hasFalseRequiresEmissionAllocation = true;
+        if (product.productRowId > 7) {
```
matthieu-foucault, line 109: This needs a TechDebt card to add an `is_energy_product` boolean and a comment in `schema/data/prod/energy_product.sql`, indicating that this line needs to be changed if any energy product is added/removed.

app/containers/Forms/Form.tsx
```diff
@@ -98,6 +98,31 @@ export const FormComponent: React.FunctionComponent<Props> = ({
...
+          productsInConflict.push(index + 1);
+        }
+      });
+    }
+
+    if (hasFalseRequiresEmissionAllocation && nonEnergyProductCount > 1)
+      errors['0'].addError(
+        `More than one non-energy product that does not require allocation of emissions cannot be reported together. -- Products in conflict: ${productsInConflict.join(
+          ','
+        )}`
```
matthieu-foucault, line 120:
```suggestion
        `${productsInConflict.join(
          ','
        )} cannot be reported together as at least one of these products does not require manual allocation of emissions`
```

I feel like this would make more sense

schema/deploy/views/ciip_production.sql
```diff
@@ -31,5 +32,6 @@ comment on column ggircs_portal.ciip_production.product_amount is 'The yearly am
 comment on column ggircs_portal.ciip_production.product_id is 'The id of the product';
 comment on column ggircs_portal.ciip_production.product_units is 'The units for the product';
 comment on column ggircs_portal.ciip_production.product_emissions is 'The amount of emissions, in tCO2e, allocated to the product';
+comment on column ggircs_portal.ciip_production.requires_emission_allocation is 'Whether or not the product requires reporting an emission allocaiton';
```
matthieu-foucault, line 35:
```suggestion
comment on column ggircs_portal.ciip_production.requires_emission_allocation is 'Whether or not the product requires reporting an emission allocation';
```

schema/deploy/computed_columns/application_revision_ciip_incentive.sql
```diff
@@ -48,6 +49,26 @@ returns setof ggircs_portal.ciip_incentive_by_product as $function$
...
     );
 
+    -- ** Test for invalid number of products when a reported product.requires_emission_allocation = false ** --
+    non_energy_product_count := 0;
+    for i in 1..array_length(reported_products, 1)
+      loop
+        if reported_products[i].is_energy_product = false then
+          non_energy_product_count = non_energy_product_count+1;
+        end if;
+      end loop;
```
matthieu-foucault, line 59: do you need a loop here?

PR #724, opened 2020-06-02
Fix: certifier rls fixes

dleard:


schema/deploy/policies/facility_policies.sql
```diff
@@ -25,9 +25,9 @@ $fn$
     join ggircs_portal.ciip_user cu
       on cu.email_address = cer.certifier_email
       and cu.uuid = (select sub from ggircs_portal.session());
-$fn$ language sql strict stable;
+$fn$ language sql strict stable security definer;
```
matthieu-foucault, line 28: perhaps add a comment here documenting why this function is a security definer

PR #725, opened 2020-06-02
Fix: incentive table details

dleard:


app/containers/Incentives/IncentiveSegmentContainer.tsx
```diff
@@ -15,37 +13,33 @@ const IncentiveSegmentContainer: React.FunctionComponent<Props> = ({
...
-      <td>
-        {process.env.NO_MATHJAX ||
-        getConfig()?.publicRuntimeConfig.NO_MATHJAX ? null : (
-          <MathJax.Context input="tex">
-            <MathJax.Node>{formula}</MathJax.Node>
-          </MathJax.Context>
-        )}
-      </td>
-      <td>CAD {incentiveProduct} </td>
+      <td>{Number.parseFloat(Number(emissionIntensity).toFixed(4))}</td>
```
matthieu-foucault, line 28: Are all of those `Number.parseFloat` needed?

dleard: I added those for any of the values that we are rounding to 4 decimal places. The reason is if some of those values evaluate evenly to less decimal places, without the Number.parseFloat hanging zeroes are added to the end. example:
**0.1** becomes **0.1000**

matthieu-foucault: makes sense

PR #732, opened 2020-06-05
feat: Front-end energy product validation

dleard:
- creates a validation error for each energy product if it is required+missing


dleard:
> Maybe we can add a techdebt card to rewrite this in a more maintainable way. E.g. generate a list or required products based on the energy options. I would not implement it in this sprint though

Alec and I talked about this in #730 and yes, that's a good idea. I'll add a card, if I didn't already, I forget if I did it or not.

PR #743, opened 2020-06-08
Batch certify

kriscooke:
- Select certification requests individually and Select All on `/certifier/requests` page
- Application and legal disclaimer text appears along with signature component
- Certifier can sign off on the selected applications at the same time (only the ones that are currently in view, based on the search/filter query).

Included is a refactor of the CertificationSignature component, which can now provide signing for multiple certification requests at a time.

[GGIRCS-1520](https://youtrack.button.is/issue/GGIRCS-1520)

### DIFF
--- a/app/components/SearchTableLayout.tsx
+++ b/app/components/SearchTableLayout.tsx
@@ -15,6 +15,7 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
   const {
     handleEvent,
     handleSelectAll,
+    allSelected = false,
     displayNameToColumnNameMap,
     body
   } = props;
@@ -43,6 +44,7 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
                   <br />
                   <input
                     type="checkbox"
+                    checked={allSelected}
                     onChange={(e) => handleSelectAll(e.target.checked)}
                   />
                 </label>
--- a/app/components/SearchTableLayout.tsx
+++ b/app/components/SearchTableLayout.tsx
@@ -4,13 +4,20 @@ import SortableTableHeader from 'components/SortableTableHeader';
 import SearchBox from 'components/SearchBox';
 interface Props {
   handleEvent: (...args: any[]) => void;
+  handleSelectAll?: (...args: any[]) => void;
+  allSelected?: boolean;
   displayNameToColumnNameMap: any;
   body: any;
 }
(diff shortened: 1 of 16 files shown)


app/components/SearchTableLayout.tsx
```diff
@@ -4,13 +4,21 @@ import SortableTableHeader from 'components/SortableTableHeader';
 import SearchBox from 'components/SearchBox';
 interface Props {
   handleEvent: (...args: any[]) => void;
+  handleSelectAll?: (...args: any[]) => void;
```
dleard, line 7: I know there's a lot of examples of this pattern `..args: any[]` all over, but we're trying to avoid being so general going forward, and I imagine at some point we'll end up doing some tech debt to fix these. I would suggest
```suggestion
  handleSelectAll?: (boolean) => void;
```

app/containers/Certifier/CertificationRequestsContainer.tsx
```diff
@@ -13,13 +13,26 @@ function formatListViewDate(date) {
...
 interface Props {
   direction: string;
   orderByField: string;
   searchField: string[];
   searchValue: string[];
   offsetValue: number;
+  forceRefetch?: number;
+  selections: string[];
   handleEvent: (...args: any[]) => void;
+  notifySelections: (selectedIds: string[]) => void;
```
dleard, line 35: :+1:

app/containers/Forms/CertificationSignature.tsx
```diff
@@ -1,21 +1,28 @@
...
+export const CertificationSignature: React.FunctionComponent<Props> = ({
+  relay,
+  certificationIdsToSign,
+  submitted = false,
+  reportSubmissions
+}) => {
   const sigCanvas: any = useRef({});
 
-  const uploadImage = (e) => {
+  const readImage = (e) => {
```
dleard, line 25: :+1:  Much better name

app/components/SearchTableLayout.tsx
dleard: Actually, this is the only change I see needed in this PR & it's pretty minor. Lets just merge it and I'll add this change in my current PR

kriscooke: I do usually try to make them more specific, but this was probably left over from an interim step when I was figuring things out and forgot to update it :)

PR #768, opened 2020-06-16
feat: use acme.sh to provision ssl certificates

wenzowski:
untested; best guess at how to do this given it seems we can't use any of the supported k8s Let's Encrypt provisioners

### DIFF
--- a/app/server/index.js
+++ b/app/server/index.js
@@ -1,4 +1,7 @@
 const express = require('express');
+const http = require('http');
+const https = require('https');
+const fs = require('fs');
 const {postgraphile} = require('postgraphile');
 const next = require('next');
 const PgManyToManyPlugin = require('@graphile-contrib/pg-many-to-many');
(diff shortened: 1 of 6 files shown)


wenzowski:
related: should probably pull in [mkcert](https://github.com/FiloSottile/mkcert) so dev env has a [certificate for localhost](https://letsencrypt.org/docs/certificates-for-localhost/)

app/server/index.js
```diff
@@ -294,11 +297,31 @@ app.prepare().then(() => {
     return handle(req, res);
   });
 
-  server.listen(port, (err) => {
-    if (err) {
-      throw err;
-    }
+  if (process.env.HOST.startsWith('https')) {
```
matthieu-foucault, line 300: `process.env.HOST` is `undefined` in our e2e test jobs. 
```suggestion
  if (process.env.HOST && process.env.HOST.startsWith('https')) {
```

wenzowski: goddamn prettier. I wrote `/^https/.test(process.env.HOST)`. fixing

wenzowski:
@matthieu-foucault maybe we should consider turning off those two offending eslint rules? test does an implicit cast `String(arg)` of the passed argument where `startsWith` is only defined on the prototype of `String`. https://tc39.es/ecma262/#sec-regexp.prototype.test

That's `unicorn/prefer-starts-ends-with` & `@typescript-eslint/prefer-string-starts-ends-with`

The example I wrote will not match on null or undefined when cast as strings `String(null) => 'null'` & `String(undefined) => 'undefined'`. If truthiness is easier to read, can do the chaining you propose.

PR #774, opened 2020-06-16
perf: lower limits/requests on pods and set pod activeDeadlineSeconds

matthieu-foucault:


helm/cas-ciip-portal/templates/airflow-dag-trigger.yaml
```diff
@@ -25,10 +25,18 @@ spec:
...
 {{ include "cas-ciip-portal.labels" . | indent 8 }}
     spec:
+      activeDeadlineSeconds: 1200
       restartPolicy: Never
       containers:
       - name: {{ template "cas-ciip-portal.fullname" . }}-dag-trigger
         image: {{ .Values.image.schema.repository }}:{{ .Values.image.schema.tag }}
+        resources:
+          limits:
+            cpu: 50m
```
wenzowski, line 35: We'll have to watch this doesn't hit a CFS interaction by being such a small fraction of a core. Only one way to find out!

wenzowski: quick [overview](https://neuvector.com/container-security/optimize-i-o-intensive-containers/) available

PR #781, opened 2020-06-17
Release/v1.0.0 rc.5 (no UI merge button)

matthieu-foucault:


matthieu-foucault:
This won't deploy because of https://github.com/zalando/spilo/issues/188#issuecomment-440279726
Fix in #782 (maybe rc.6 will actually make it to prod)

PR #791, opened 2020-06-19
feat: use secure cookie session backed by connect-pg-simple

wenzowski:
- set cookie to be secure when https is enabled
- remove the hardcoded symmetric key
- auto-generate a good symmetric key with helm
- auto-generate an ephemeral symmetric key in dev environment
- deploy a session table to the private postgres schema
- use connect-pg-simple as postgres backing store for session


wenzowski:
this patch is gross but we're up against the wall and we really needed to purge the hardcoded key
~~secret: 'change me pls for the love of Jibbers Crabst'~~

helm/cas-ciip-portal/templates/secret.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
     "helm.sh/hook-delete-policy": "before-hook-creation"
 type: Opaque
 data:
+  session-secret: {{ randAlphaNum 32 | b64enc | quote }}
```
wenzowski, line 12: Since this secret template is scheduled as a helm `pre-install` hook, the change to add a `session-secret` field will need to be manually patched in dev, test, and prod before releasing this patch or things will go boom.

wenzowski:
@dleard I'm not actually sure why the style test is failing?
`test/style/table_test.sql:23: ERROR:  relation "session" does not exist`

AFAIK `session` isn't a [reserved word](https://www.postgresql.org/docs/current/sql-keywords-appendix.html) and the new table has a comment describing its purpose?

junminahn:
> @dleard I'm not actually sure why the style test is failing?
> `test/style/table_test.sql:23: ERROR: relation "session" does not exist`
> 
> AFAIK `session` isn't a [reserved word](https://www.postgresql.org/docs/current/sql-keywords-appendix.html) and the new table has a comment describing its purpose?

maybe `search_path` in the test should include the schema name of `session` table `ggircs_portal_private`?
so something like `set search_path to ggircs_portal,public,ggircs_portal_private;` in
https://github.com/bcgov/cas-ciip-portal/blob/e326521266fc99e3fb799f89bd31f2e134e49e1f/schema/test/style/table_test.sql#L7
https://github.com/bcgov/cas-ciip-portal/blob/e326521266fc99e3fb799f89bd31f2e134e49e1f/schema/deploy/tables/session.sql#L6

wenzowski:
thanks @junminahn :) split update of reserved words list to #792 and altered table name since `session` was a reserved word in SQL-92

matthieu-foucault:
@wenzowski needs a rebase

dleard:
@wenzowski @junminahn Both good catches with `set search_path` and `reserved words`, These style tests were created quite a whilte ago and are a bit rough in some areas. In many places are likely set to look at specific schemas, and the list of reserved words were pulled from a postgres reserved words list without any real vetting. I'd really like to spend some time after we've made our release to go back into these style tests and clean them up, document them & make them all schema / project agnostic so that they can be readable and workable enough to become a resource that can be shared & used in the lab or government as a whole

matthieu-foucault:
- [x] session secrets generated and manually added to the namespaces

PR #792, opened 2020-06-19
fix: canonical list of reserved words

wenzowski:
Avoid use of any [SQL Key Words] from:
- SQL-92
- SQL:2011
- SQL:2016
- PostgreSQL

[SQL Key Words]: https://www.postgresql.org/docs/12/sql-keywords-appendix.html


wenzowski:
scraped out of pg docs & alphabetically sorted

PR #794, opened 2020-06-19
small tweaks to reduce docker app build time

matthieu-foucault:


matthieu-foucault:
Multi-stage docker is for next sprint maybe. I'll tweak the `.helmignore` in a separate PR

PR #795, opened 2020-06-19
chore: drop db during init-db cronjob for dev

dleard:


helm/cas-ciip-portal/values-wksv3k-dev.yaml
```diff
@@ -3,6 +3,9 @@ route:
   host: dev.ciip.gov.bc.ca
   insecure: false # cert has been issued
 
+dag:
+  preInitDb: "dropdb $(PORTAL_DATABASE)"
```
matthieu-foucault, line 7:
```suggestion
  preInitDb: "dropdb --if-exists $(PORTAL_DATABASE)"
```

helm/cas-ciip-portal/values.yaml
```diff
@@ -37,6 +37,9 @@ persistence:
   appUser: ciip_portal
   readonlyUser: ciip_readonly
 
+dag:
+  preInitDb: ""
```
matthieu-foucault, line 41:
```suggestion
  # a command to be executed at the beginning of the init-db job, running in the psql image above
  preInitDb: ""
```

PR #798, opened 2020-06-19
feat: nuke the siteminder session cookie on logout

wenzowski:
According to numerous Rocket Chat threads and https://github.com/bcgov/ocp-sso/issues/4 the logout function only clears the Keycloak session and never clears the SiteMinder SSO session.

This fix attempts to purge the SiteMinder cookie on logout when we're deployed on a subdomain of .gov.bc.ca

### DIFF
--- a/app/server/index.js
+++ b/app/server/index.js
@@ -3,14 +3,14 @@ const http = require('http');
 const https = require('https');
 const fs = require('fs');
 const {postgraphile} = require('postgraphile');
-const next = require('next');
+const nextjs = require('next');
 const PgManyToManyPlugin = require('@graphile-contrib/pg-many-to-many');
 
 const crypto = require('crypto');
 const pg = require('pg');
 const port = Number.parseInt(process.env.PORT, 10) || 3004;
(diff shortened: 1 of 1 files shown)


app/server/index.js
```diff
@@ -183,6 +183,16 @@ app.prepare().then(() => {
   };
   const keycloak = new Keycloak({store}, kcConfig);
 
+  // Nuke the siteminder session token on logout if we can
+  // this will be ignored by the user agent unless we're
+  // currently deployed to a subdomain of gov.bc.ca
+  //
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars-experimental
+  server.post('/logout', (req, res, next) => {
```
matthieu-foucault, line 191: renaming `req` to `_req` might turn off the eslint error?

wenzowski: thanks, fixed

matthieu-foucault:
@wenzowski just needs a rebase

matthieu-foucault:
@wenzowski I can confirm this works in the `wksv3k-dev` environment

PR #799, opened 2020-06-19
Feat: prod email config

dleard:
- Throttle emails to send max 100 per minute
- Does 'NOT' yet make any change to the 'senderEmail' value as there is still some discussion going on about what to do about it. https://youtrack.button.is/issue/GGIRCS-1646


docs/developer-guide.md
```diff
@@ -119,8 +119,47 @@ Here are some key differences between the two, found in [this beautiful post](ht
 - No need to test as many implementation details
 
 
+### Sqitch
+
+We use Sqitch to manage our database with Sqitch via the sqitch.plan in the schema directory.
```
matthieu-foucault, line 124:
```suggestion
We manage our database with [sqitch] via the sqitch.plan in the `/schema` directory.
```

docs/developer-guide.md
```diff
@@ -119,8 +119,47 @@ Here are some key differences between the two, found in [this beautiful post](ht
...
+Any database entities in the plan before a release are immutable and any changes to them must be done as follows:
+
+### Tables (or other non-idempotent changes)
+
+Changes to tables require creating a new entry in the sqitch plan for that table, for example:
+`sqitch add tables/application_001 --require tables/application`
+This will create a deploy,revert and verify files for tables/application_001
+- In the deploy file, any necessary changes can be made to the table with `alter table`.
+- In the revert file, any changes made in the deploy file must be undone, again with `alter table`
+- The verify file should just verify the existence of the table, and can probably be the same as the original verify file
```
matthieu-foucault, line 145: This actually follows the recommended approach from sqitch's author https://stackoverflow.com/a/35091927

schema/deploy/database_functions/graphile_worker_job_definer.sql
```diff
@@ -5,8 +5,17 @@ begin;
 
 create or replace function ggircs_portal_private.graphile_worker_job_definer(task text, payload json)
 returns void as $$
+  declare
+    calls_past_minute int;
+    defer_run_multiplier int;
   begin
-    perform graphile_worker.add_job(task, payload);
+
+    calls_past_minute := (select count(*) from ggircs_portal_private.graphile_worker_timestamp where called_at > (now() - interval '1 minute'));
```
matthieu-foucault, line 13: why not using the `run_at` column from `graphile_worker.jobs` ?

docs/developer-guide.md
dleard: Is my description different? Or did I miss something from the recommended approach? It looks pretty much the same to me.

schema/deploy/database_functions/graphile_worker_job_definer.sql
dleard: Cool, I wasn't aware of graphile_worker.jobs, it should remove the need for an extra table, I'll try doing it with graphile_worker.jobs

PR #815, opened 2020-06-24
Fix login redirects

kriscooke:
- Properly diverts certification-redirect page through login 
  * `needsUser` and `needsSession` props
- Front-end routing handles handles login redirects _before_ non-permissible resource redirects
  * without logging in, how were we to know if it is permissible or not for the user?
- Login button action properly URI-encodes the `redirectTo` query param
  * prevents losing some of the query params at destination
- End-to-end [test helper](https://github.com/bcgov/cas-ciip-portal/pull/815/files#diff-ac6ee8c82c5d20d9bb9c4d6ac70161d4R9) identifies all pages that require authentication, generates tests for each route and asserts that navigation is successfully routed to the final destination through login


matthieu-foucault:
@kriscooke what's the status of this PR? This seemed to be working last time we looked at it

kriscooke:
@matthieu-foucault Have updated the description with outstanding issues. Am about to cry 'uncle' on the test fixtures and would ❤️ it if some generous person wanted to give it a boost.

app/cypress/support/listAuthenticatedPages.js
```diff
@@ -0,0 +1,31 @@
...
+
+    for (const folder of folders) {
+      json[folder.name] = await getRoutes(`${currentPath}/${folder.name}`);
+    }
+
+    return folders.length === 0 ? files : json;
+  }
+
+  const routes = await getRoutes('pages', ['resources']);
+  if (writeFile) fs.writeFile(writeFile, JSON.stringify(routes));
```
matthieu-foucault, line 27: you might want to write to a cypress env variable instead of writing to a file: https://docs.cypress.io/api/cypress-api/env.html#Name-and-Value

app/cypress/support/index.js
```diff
@@ -1,3 +1,6 @@
 import './commands';
 import 'cypress-axe';
 require('cypress-plugin-retries');
+import authenticatedPages from './authenticatedPages.json';
+
+Cypress.env.AUTHENTICATED_PAGES = authenticatedPages;
```
matthieu-foucault, line 6: cypress will try importing this file before it's created. See comment below about writing directly to the env variable

app/cypress/support/listAuthenticatedPages.js
```diff
@@ -0,0 +1,31 @@
+const fs = require('fs').promises;
+const path = require('path');
+
+async function listAuthenticatedPages(
```
matthieu-foucault, line 4: should this be a cypress command instead, called at the beginning of the login-redirect spec file?

app/cypress/support/listAuthenticatedPages.js
```diff
@@ -0,0 +1,31 @@
...
+  writeFile = 'cypress/support/authenticatedPages.json'
+) {
+  async function getRoutes(currentPath, ignoreFolders) {
+    const json = {};
+    const entries = await fs.readdir(currentPath, {withFileTypes: true});
```
matthieu-foucault, line 9: You can use `readdirSync` here, this method doesn't need to be asynchronous

app/cypress/support/listAuthenticatedPages.js
```diff
@@ -0,0 +1,31 @@
...
+      .filter((file) => !file.isDirectory())
+      .map((file) => path.basename(file.name, '.tsx'));
+
+    const folders = entries.filter(
+      (folder) => folder.isDirectory() && !ignoreFolders.includes(folder.name)
+    );
+
+    for (const folder of folders) {
+      json[folder.name] = await getRoutes(`${currentPath}/${folder.name}`);
+    }
```
matthieu-foucault, line 21: If you want to keep async code and not ignore the linter warning, something like 
```js
const json = Promise
  .all(folders.map((folder) => getRoutes(`${currentPath}/${folder.name}`))
  .reduce((acc, curr, i) => acc[folders[i].name] = curr, {})
```
might work

schema/data/fixtures/login-redirects-setup.sql
```diff
@@ -0,0 +1,145 @@
...
+    end loop;
+  return 'true';
+end;
+$function$ language plpgsql strict volatile;
+grant execute on function e2e_test_initialize_form to ciip_administrator, ciip_analyst, ciip_industry_user;
+
+-- Initializes form results
+select e2e_test_initialize_form(200, 101);
+select e2e_test_initialize_form(300, 102);
+select e2e_test_initialize_form(400, 103);
```
matthieu-foucault, line 134: why is all of this needed? can't you set a reporting year and use `create_application_mutation_chain` instead?

kriscooke: What's necessary is to be able to explicitly set ids (overriding id generation) in order for the test to have known ids to compute the URL for certain routes with required query params, eg: `/reporter/application?applicationId=blah&version=1&formResultId=blah`. 

Unfortunately that seems to mean copying a bunch of the things that `create_application_mutation_chain` and `create_application_revision_mutation_chain` do. This is what @dleard and I were working on yesterday.

Similar affected routes are listed in REQUIRED_QUERY_PARAMS in the login-redirects test.

matthieu-foucault: if you do `truncate cascade restart identity` on the application table at the beginning of the fixture, you should have known ids without needing to override id generation

app/cypress/support/index.js
kriscooke: I had a `precypress` script that ensured the file is created before this import, but this just reminded me that on CI Cypress is run via `test:e2e` instead of the `cypress` script I use locally. So it needed a `pretest:e2e` as well for CI.

```json
"precypress": "node cypress/support/listAuthenticatedPages.js",
```

app/cypress/support/listAuthenticatedPages.js
kriscooke: I tried initially going that route, but unfortunately there are a few constraints that prevented it:
- `cy.exec` returns limited data, but I did try a solution writing to `process.stdout` and then parsing stdout in the command
- however, the scope of env variables dynamically set with `Cypress.env` is limited to within individual tests (if set in the command scope, then not present in the test)
  * even when the command is written in the spec file, the variable is used _to generate_ tests and is therefore needed to be scoped _outside_ a test.
  * cannot run a Cypress command outside of a test. Even running the command in a different, fake test in the same specfile does not prioritize the fake test to run before the loop that generates the real tests.

As result of that exploration, the written file seems to be necessary.

schema/data/fixtures/login-redirects-setup.sql
kriscooke: Thanks! Have updated accordingly

app/package.json
```diff
@@ -6,7 +6,9 @@
     "pretest": "npm run build:relay",
     "test": "jest",
     "test:e2e-snapshots": "percy exec -- cypress run",
+    "pretest:e2e": "node cypress/support/listAuthenticatedPages.js",
```
matthieu-foucault, line 9: You'll need to add a "pretest:e2e-snapshots" too

PR #816, opened 2020-06-24
Feat: Missing SWRS report warning

dleard:
- Removes the reporting_year param from facility_has_swrs_report.sql
- Adds a modal that appears on starting an application if no swrs report was found


matthieu-foucault:
Does the "Draft application started email" e2e test is for an application that doesn't have a SWRS report? If so the test needs to be updated to go through the new modal

app/containers/Applications/ApplyButtonContainer.tsx
```diff
@@ -58,10 +59,74 @@ export const ApplyButton: React.FunctionComponent<Props> = ({
...
+            its regulations.
+          </p>
+          <p>
+            For any questions, please{' '}
+            <a href="mailto: ghgregulator@gov.bc.ca">contact</a> the Climate
+            Action Secretariat at GHGRegulator@gov.bc.ca
+          </p>
+        </Modal.Body>
+        <Modal.Footer>
+          <Button variant="danger" onClick={async () => startApplication}>
```
matthieu-foucault, line 103:
```suggestion
          <Button variant="danger" onClick={startApplication}>
```

PR #817, opened 2020-06-25
fix: search functions no longer fail redirects

dleard:
- add some guard `if` clauses for when logged out


dleard:
replaced by #819

PR #820, opened 2020-06-25
Fix: require org access in application/facility rls

dleard:
- adds org access = 'approved' in rls using functions
- updates / adds pgTap tests


schema/test/unit/tables/form_result_status_test.sql
```diff
@@ -32,7 +32,7 @@ insert into ggircs_portal.organisation(id) overriding system value values(999),
 insert into ggircs_portal.facility(id, organisation_id) overriding system value values(999, 999), (1000, 1000);
 insert into ggircs_portal.application(id, facility_id) overriding system value values(999, 999), (1000, 1000);
 insert into ggircs_portal.application_revision(application_id, version_number) overriding system value values(999, 1), (999, 2), (1000, 1), (1000,2), (1000,3);
-insert into ggircs_portal.ciip_user_organisation(id, user_id, organisation_id) overriding system value values(999, 999, 999), (1000, 1000, 1000);
+insert into ggircs_portal.ciip_user_organisation(id, user_id, organisation_id, 'status') overriding system value values(999, 999, 999, 'approved'), (1000, 1000, 1000, 'approved');
```
matthieu-foucault, line 35:
```suggestion
insert into ggircs_portal.ciip_user_organisation(id, user_id, organisation_id, status) overriding system value values(999, 999, 999, 'approved'), (1000, 1000, 1000, 'approved');
```
syntax error here

PR #821, opened 2020-06-25
Fix: update on conflict() for import org facility

dleard:
- change on conflict for organisation to (swrs_organisation_id)
- change on conflict for facility to (swrs_facility_id)
- add uindex to each table for the above columns


schema/revert/tables/facility_001.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert ggircs-portal:tables/facility_001 from pg
+
+begin;
+
+-- Unique indexes cannot be dropped, migration will be rolled back by original revert
```
matthieu-foucault, line 5: what about https://www.postgresql.org/docs/12/sql-dropindex.html

PR #822, opened 2020-06-25
feat: yarn start uses --unhandled-rejections=strict option

matthieu-foucault:
This makes the app crash when the graphile worker fails to connect to the db,
instead of silently continuing to work

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -16,7 +16,7 @@
     "build:relay": "relay-compiler --verbose",
     "build:next": "next build",
     "build:storybook": "build-storybook -o static/storybook",
-    "start": "NODE_ENV=production node server",
+    "start": "NODE_ENV=production node --unhandled-rejections=strict server",
     "format": "xo --fix",
     "lint": "xo",
     "prestorybook": "npm run build:relay",


wenzowski:
I'm curious that the worker was still trying to work and the app didn't crash? Did you observe the behaviour in the worker post migration to the [shared `pgPool` instance](https://github.com/bcgov/cas-ciip-portal/pull/791/files#diff-52366f0190c9ffdf8f05378119aeee16R93) introduced in #791?

wenzowski:
Our k8s heartbeat service should probably be doing its http healthcheck on an endpoint that returns an http 503 on [`pgPool.idleCount === 0`](https://node-postgres.com/api/pool). Our `readinessProbe` should deregister the pod from the service to let our db pool drain, followed by our `livenessProbe` killing the pod if the pool does drain (longer timeout for liveness than readiness). What do you think? Would love to [dump pool health metrics to sysdig](https://docs.sysdig.com/en/integrate-node-js-application-metrics.html) as well, if possible, so we can see if we end up connection thrashing.

matthieu-foucault:
@wenzowski Yes. The issue was observed after #791 The error that was triggered was `cannot execute LISTEN during recovery`, which I suppose is specific to the worker

wenzowski:
I've thought about this a bit. The node livenessProbe and the readinessProbe are a bad idea here: since we have no fail whale to fall back to, we need to avoid causing a cascading failure of an unavailable postgres taking out our app pods. I think we actually need the pool heath as an express middleware that responds to users with its own 503 fail whale directly and watch our golden signals in sysdig.

wenzowski:
the act of requesting to LISTEN requires postgres to execute an update in the database to record the listener on the connection. In old versions, it put tuples into pg_catalog.pg_listener and I'm not sure where that mechanism is implemented now in pg 12. So that cannot execute LISTEN during recovery issue is being caused by somehow ending up with a connection pool with TCP sockets stuck on a read replica after failover instead of being closed and reopened on the new master

wenzowski:
so our node process itself needs a heartbeat on patroni and needs to initialize and swap in a new pool instance and [reap the existing pool](https://github.com/brianc/node-postgres/tree/master/packages/pg-pool#shut-it-down) whenever the master changes

PR #824, opened 2020-06-26
chore(release): 1.0.0-rc.8

matthieu-foucault:


matthieu-foucault:
@dleard ignore the failing `sqitch-files-before-tag-are-immutable` check. I should be comparing with `origin/develop`, not `origin/master`. Will fix in a subsequent PR

PR #826, opened 2020-06-26
Feat: signature feedback

dleard:
- adds a toast & redirect on signature

### DIFF
--- a/app/containers/Forms/CertificationSignature.tsx
+++ b/app/containers/Forms/CertificationSignature.tsx
@@ -1,7 +1,8 @@
-import React, {useRef} from 'react';
+import React, {useRef, useState} from 'react';
+import {useRouter} from 'next/router';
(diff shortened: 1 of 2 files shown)


app/containers/Forms/CertificationSignature.tsx
```diff
@@ -89,7 +94,7 @@ export const CertificationSignature: React.FunctionComponent<Props> = ({
             />
           )}
         </Col>
-        <Col md={{span: 3, offset: 2}}>
+        <Col md={{offset: 3, span: 3}}>
```
matthieu-foucault, line 97: The offset changed here. Is this on purpose?

dleard: Nope, I think I set my span for the rendered Toast message at 3, so I set the offset as the same. May as well change it back to what it was.

PR #835, opened 2020-06-29
fix: handle null facility names in search dropdown

dleard:


app/containers/Facilities/Facility.tsx
```diff
@@ -8,18 +8,17 @@ export const FacilityComponent = (props) => {
...
   }
 
-  if (
-    !facility.facilityName
-      .toLowerCase()
-      .includes(props.facilityInput.toLowerCase())
-  )
+  let {facilityName} = facility;
+
+  if (!facility.facilityName) facilityName = 'UNDEFINED FACILITY NAME';
```
matthieu-foucault, line 13: maybe replace this with `'empty name (swrs facility id = ....)'` so that the user can identify the facility if needed

PR #836, opened 2020-06-29
fix: diff table - NO DATA ENTERED

dleard:
values of 0 were being treated as falsey in the logic determining whether to display the value or NO DATA ENTERED. I fixed the logic to display the 0 instead of NO DATA ENTERED.
https://youtrack.button.is/issue/GGIRCS-1570

### DIFF
--- a/app/containers/Forms/SummaryEmissionGasFields.tsx
+++ b/app/containers/Forms/SummaryEmissionGasFields.tsx
@@ -59,19 +59,11 @@ const EmissionGasFields: React.FunctionComponent<FieldProps> = (props) => {
           (props.formData.annualEmission || previousEmission) ? (
             <>
               <span className="diffFrom">
-                {previousEmission || previousEmission === 0 ? (
-                  previousEmission
-                ) : (
-                  <i>[No Data Entered]</i>
-                )}
+                {previousEmission ?? <i>[No Data Entered]</i>}
               </span>
               &nbsp;---&gt;&nbsp;
               <span className="diffTo">
-                {props.formData.annualEmission ? (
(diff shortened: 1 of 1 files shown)


app/containers/Forms/SummaryEmissionGasFields.tsx
```diff
@@ -59,7 +59,11 @@ const EmissionGasFields: React.FunctionComponent<FieldProps> = (props) => {
           (props.formData.annualEmission || previousEmission) ? (
             <>
               <span className="diffFrom">
-                {previousEmission ? previousEmission : <i>[No Data Entered]</i>}
+                {previousEmission || previousEmission === 0 ? (
+                  previousEmission
+                ) : (
+                  <i>[No Data Entered]</i>
+                )}
```
matthieu-foucault, line 66:
```suggestion
              previousEmission ?? <i>[No Data Entered]</i>
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

app/containers/Forms/SummaryEmissionGasFields.tsx
```diff
@@ -90,7 +94,7 @@ const EmissionGasFields: React.FunctionComponent<FieldProps> = (props) => {
           (props.formData.annualCO2e || previousAnnualCO2e) ? (
             <>
               <span className="diffFrom">
-                {previousAnnualCO2e ? (
+                {previousAnnualCO2e || previousAnnualCO2e === 0 ? (
```
matthieu-foucault, line 97: also use nullish coalescing

PR #838, opened 2020-06-30
'Dashboard' button in header

kriscooke:
- Updates 'Industry'/'Admin' label for dashboard button in header to 'Dashboard' per [GGIRCS-1665](https://youtrack.button.is/issue/GGIRCS-1665) (encompasses all roles)
- Improves accessibility of 'Dashboard' and 'Profile' links in header by using anchor tag
  * use of form + submit buttons for the sole purpose of navigation not in line with best practise. 
    - ie: those that use get method with no form data
  * anchor tags gives the user access to the right-click context menu to open in new tab/window, copy link address/text, shows a preview of the target URL on hover, etc.

### DIFF
--- a/app/components/Layout/Header.tsx
+++ b/app/components/Layout/Header.tsx
@@ -33,12 +33,9 @@ const HeaderLayout = ({isLoggedIn = false, isRegistered = false}) => (
                   <Form.Row>
                     {isRegistered && (
                       <Col>
-                        <a
-                          href="/user/profile"
-                          className="btn btn-outline-light"
-                        >
-                          Profile
-                        </a>
+                        <Link href="/user/profile">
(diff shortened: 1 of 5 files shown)


app/components/Layout/Header.tsx
```diff
@@ -45,11 +33,12 @@ const HeaderLayout = ({
                   <Form.Row>
                     {isRegistered && (
                       <Col>
-                        <Form action="/user/profile" method="get">
-                          <Button type="submit" variant="outline-light">
-                            Profile
-                          </Button>
-                        </Form>
+                        <a
```
matthieu-foucault, line 36: This should be using a Nextjs `Link`

kriscooke: Good catch!

PR #839, opened 2020-06-30
Pages for footer links from BCGov

kriscooke:
- Adds content to pages linked in footer per [GGIRCS-1589](https://youtrack.button.is/issue/GGIRCS-1589)
  * Accessibility links to outside BCGov accessibility page (referenced in DevHub) because it hosts a bunch of accessibility resources that don't make sense to copy and host on this site.


app/tests/unit/pages/resources/privacy.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const query: privacyQueryResponse['query'] = {
+  session: {
+    ' $fragmentRefs': {
+      defaultLayout_session: true
+    }
+  }
+};
+
+const router = {
```
dleard, line 14: @kriscooke The router prop shouldn't be necessary for any of these pages, so it can be removed from all the unit tests

kriscooke: It is required, TS complains without. It comes as part of `CiipPageComponentProps`

kriscooke: That said, TS isn't totally happy with what I've got in the query either:
```
No overload matches this call.
  Overload 1 of 2, '(props: Readonly<Props>): Privacy', gave the following error.
    Type '{ query: { applicationId: string; hasSwrsReport: boolean; version: number; }; }' is not assignable to type 'NextRouter'.
      Type '{ query: { applicationId: string; hasSwrsReport: boolean; version: number; }; }' is missing the following properties from type 'BaseRouter': route, pathname, asPath, basePath
  Overload 2 of 2, '(props: Props, context?: any): Privacy', gave the following error.
    Type '{ query: { applicationId: string; hasSwrsReport: boolean; version: number; }; }' is not assignable to type 'NextRouter'.
      Type '{ query: { applicationId: string; hasSwrsReport: boolean; version: number; }; }' is not assignable to type 'BaseRouter'.ts(2769)
next-env.d.ts(16, 3): The expected type comes from property 'router' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<Privacy> & Readonly<Props> & Readonly<{ children?: ReactNode; }>'
```

dleard: hm. If the router isn't required for all pages that extend CiipPageComponentProps, maybe we should update the interface definition of `router` in `next-env.d.ts` to be optional? @matthieu-foucault thoughts?

dleard: I'm inclined to change the definition to make the router optional in the interface, but this `<ApplicationDisclaimer query={query} router={null} />` will also silence that angry TS message & it's a little more clear in the test file that the contents of router has nothing to do with the test

kriscooke: Me too - went that route since I didn't hear any complaints :)

PR #848, opened 2020-07-03
Set robots header to noindex for non-production environments

kriscooke:
- Configures a middleware setting the `X-Robots-Tag` header for all non-production domains, per [GGIRCS-1631](https://youtrack.button.is/issue/GGIRCS-1631)
  * covers all assets, including static assets like images
  * [read about the directives here](https://developers.google.com/search/reference/robots_meta_tag#directives)

### DIFF
--- a/app/server/index.js
+++ b/app/server/index.js
@@ -154,6 +154,15 @@ app.prepare().then(() => {
   server.use(bodyParser.json());
   server.use(cors());
 
+  // Tell search + crawlers not to index non-production environments:
+  server.use(({res, next}) => {
+    if (!process.env.NAMESPACE || !process.env.NAMESPACE.endsWith('-prod')) {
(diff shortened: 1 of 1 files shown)


app/server/index.js
```diff
@@ -154,6 +154,15 @@ app.prepare().then(() => {
   server.use(bodyParser.json());
   server.use(cors());
 
+  // Tell search + crawlers not to index non-production environments:
+  server.use(({res, next}) => {
+    if (process.env.NODE_ENV !== 'production') {
```
matthieu-foucault, line 159: `NODE_ENV` will be `production` for all namespaces, as defined in the `package.json`'s start script. `process.env.NAMESPACE` is what you want to look at here
```suggestion
    if (!process.env.NAMESPACE || !process.env.NAMESPACE.endsWith('-prod') {
```

kriscooke: Thanks, I'd seen that and was going to ask about it!

PR #849, opened 2020-07-03
Install test analytics module

kriscooke:
Installs test tracker snippet for GDX snowplow analytics, currently intended for test environment only.

### DIFF
--- a/app/pages/_document.tsx
+++ b/app/pages/_document.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import Document, {Html, Head, Main, NextScript} from 'next/document';
+import Analytics from 'components/Layout/Analytics';
 
 class MyDocument extends Document {
   render() {
@@ -14,6 +15,7 @@ class MyDocument extends Document {
         <body>
           <Main />
           <NextScript />
+          <Analytics />
         </body>
       </Html>
     );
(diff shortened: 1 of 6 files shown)


matthieu-foucault:
@kriscooke I would suggest adding an `ENABLE_ANALYTICS` env var in the relay runtime config. That variable should be defined from our helm chart (in the `helm/cas-ciip-portal/templates/deploy.yaml` template). Maybe with something like:

```yaml
analytics:
  enable: false
```
in our default values.yaml, and set to true in the dev and test values

helm/cas-ciip-portal/templates/deploy.yaml
```diff
@@ -61,6 +61,8 @@ spec:
             value: {{ .Values.env.adminEmail }}
           - name: FEEDBACK_SITE_URL
             value: {{ .Values.fider.host }}
+          - name: ENABLE_ANALYTICS
+            value: {{ .Values.analytics.enable }}
```
matthieu-foucault, line 65: k8s environment variables have to be strings
```suggestion
            value: {{ .Values.analytics.enable | quote }}
```

app/pages/_document.tsx
```diff
@@ -14,6 +18,7 @@ class MyDocument extends Document {
         <body>
           <Main />
           <NextScript />
+          {CONFIG.ENABLE_ANALYTICS && <Analytics />}
```
matthieu-foucault, line 21: The environment variable is going to be a string (see below)
```suggestion
          {CONFIG.ENABLE_ANALYTICS === 'true' && <Analytics />}
```

PR #862, opened 2020-07-06
Register button leads directly to SSO registration page

kriscooke:
Direct link to SSO registration page per [GGIRCS-1703](https://youtrack.button.is/issue/GGIRCS-1703).

Adds 2 public runtime env variables to the Next config: 
- `HOST`
  * to compute the post-registration `redirect_uri`
- `NAMESPACE`
  * to compute the keycloak namespace to link to (ie: `sso[-dev|-test].pathfinder` vs. `sso.pathfinder`)

### Decision: 
We've decided to ship this for now without support for post-registration redirects, due to some difficulties that would add precious time to debug before impending use by the public. We'll prioritize and revisit support at a later date. Those difficulties are:
- Return trip to `/login` after registering returns a 403 (however the user is still created, and subsequently clicking "Login" automatically logs them in).
  - the login service improperly decodes the `redirectTo` query param. Not sure why or how relevant.
- Failing Storyshots test suite related to using the router in the Header component.

~TODO:~
- ~Ensure `redirectTo` query param is respected~

### DIFF
--- a/app/components/Layout/Header.tsx
+++ b/app/components/Layout/Header.tsx
@@ -49,9 +49,9 @@ const HeaderLayout = ({isLoggedIn = false, isRegistered = false}) => (
                 ) : (
                   <Form.Row>
                     <Col>
-                      <LoginButton type="submit" variant="outline-light">
-                        Register
-                      </LoginButton>
+                      <Link href="/register">
+                        <a className="btn btn-outline-light">Register</a>
+                      </Link>
                     </Col>
                     <Col>
                       <LoginButton type="submit" variant="outline-light">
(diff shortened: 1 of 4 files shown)


app/components/Layout/Header.tsx
```diff
@@ -2,6 +2,10 @@ import React from 'react';
 import {Row, Col, Form, Button} from 'react-bootstrap';
 import Link from 'next/link';
 import LoginButton from 'components/LoginButton';
+import getConfig from 'next/config';
+
+const {NAMESPACE, HOST} = getConfig()?.publicRuntimeConfig;
+const registrationUrl = `https://sso${NAMESPACE}.pathfinder.gov.bc.ca/auth/realms/pisrwwhx/protocol/openid-connect/registrations?client_id=cas-ciip-portal&response_type=code&scope=openid&redirect_uri=${HOST}/login?auth_callback=1`;
```
matthieu-foucault, line 8: Can we do a redirect on the server-side on a GET request to `/register` instead of having this registrationUrl defined here? 
The components of the keycloak registration URL are available on the server in the `kcConfig` variable (`auth-server-url`, `realm`, `resource`)

PR #869, opened 2020-07-07
feat: add comments field to every form, rendered by ProblemReportField

matthieu-foucault:


matthieu-foucault:
We don't support reporting multiple problems. They would have to edit the textarea and add more text in there. I am not sure making the link unclickable is a solution, as they might want to edit their comment.
Thinking of tweaking the button/paragraph text a bit to clarify that the note will only be visible by CAS when the application is submitted

kriscooke:
I see, so the reason it's not cleared is to edit the original comment - makes sense. If we then fix the Cancel action to not clear the field then, it should be good 👍

PR #872, opened 2020-07-08
chore: pin lodash version to 4.17.18 to resolve npm security advisory

matthieu-foucault:
see https://www.npmjs.com/advisories/1523/versions


matthieu-foucault:
@dleard `4.17.17` was bad, trying `4.17.18`.

PR #879, opened 2020-07-10
fix: only load graphiql when in local environment

dleard:


app/server/index.js
```diff
@@ -220,7 +220,7 @@ app.prepare().then(() => {
   server.use(
     postgraphile(pgPool, process.env.DATABASE_SCHEMA || 'ggircs_portal', {
       appendPlugins: [PgManyToManyPlugin],
-      graphiql: true,
+      graphiql: process.env.IS_LOCAL_ENVIRONMENT === 'true',
```
matthieu-foucault, line 223:
```suggestion
      graphiql: process.env.NODE_ENV !== 'production',
```

dleard: won't this allow graphiql to run in test.ciip? Which we don't want?

dleard: nevermind, it won't. I'll make the change

PR #904, opened 2020-07-15
Reporting docs

rajpersram:
PR to incorporate the introduction of the reporting documentation along with its supporting images. Thanks!


matthieu-foucault:
@kriscooke technically, the commit type should be `docs`, not `feat`. This is not important now, but after v1, if we rely on those to automatically bump the appropriate version number, this would increase the minor version number instead of the patch (see https://www.conventionalcommits.org/en/v1.0.0/)

PR #929, opened 2020-07-20
fix: Immediately remove requested org from access dropdown

kriscooke:
- Refetch Organisations query after `createUserOrganisationMutation`.
[GGIRCS-1463](https://youtrack.button.is/issue/GGIRCS-1463)

### DIFF
--- a/app/containers/Organisations/UserOrganisation.tsx
+++ b/app/containers/Organisations/UserOrganisation.tsx
@@ -59,6 +59,7 @@ export default createFragmentContainer(UserOrganisationComponent, {
     fragment UserOrganisation_userOrganisation on CiipUserOrganisation {
       id
       status
+      organisationId
       organisationByOrganisationId {
         id
         rowId


dleard:
@kriscooke The query in the mutation & the connection are adding duplicate values in the local relay store.

tldr:
- remove the query from the createUserOrganisation mutation return object
- add `organisationId` to the fragment in UserOrganisation.tsx
- (optional): add `cursor` (above `node`) in the return object in the mutation, not needed to work but maybe good for clarity?

explanation:
The connection adds to the local relay store via a range add & then in the same mutation the query in the return object adds to the store again with the value from the database, there are not 2 rows going into the database, just the local relay store. So doing both the return query and the connection in the mutation is the problem.

The filter on line 201 of Organisations.tsx is definitely right, but organisationId will always be undefined without a refresh because of what is (or isn't) being returned in the mutation. The mutation is returning the fragment `...UserOrganisation_userOrganisation`. You can see that in that fragment, (in UserOrganisation.tsx) there is no `organisationId` value, so that value will always be undefined when the mutation returns. The reason it works on refresh is because there is an `organisationId` value in the parent fragment in Organisations.tsx (which is not returned by the mutation).

The reason one `cursor` value was undefined is the same, we are getting back only what is returned by the mutation, and no cursor is defined in the return value.

PR #938, opened 2020-07-22
validate on blur

matthieu-foucault:


kriscooke:
@matthieu-foucault Is it feasible to validate on mount along with this work as well as it covers the same UX concerns? That way, the user would see errors as soon as they load a given form tab.

app/next-env.d.ts
```diff
@@ -55,3 +58,7 @@ declare module 'react-jsonschema-form/lib/utils' {
     idPrefix: string
   ): IdSchema;
 }
+// Missing type declarations from /lib/validate
+declare module 'react-jsonschema-form/lib/validate' {
+  export function toErrorList(errprSchema: ErrorSchema): any[];
```
kriscooke, line 63: Typo:
```suggestion
  export function toErrorList(errorSchema: ErrorSchema): any[];
```

matthieu-foucault:
@kriscooke the problem with validating on mount is that the required empty fields would all have errors. We could try to filter for fields that have a value in them, but that may be tricky. I'd create another feature PR for that

kriscooke:
I see, I see. 🤔 I suppose it could use some more thought- we wouldn't want all the required fields to have errors the first time a given form is loaded (before they've even had a chance to work on it).

matthieu-foucault:
@kriscooke to you want me to look into validating on mount in this PR or should we schedule that for later?

kriscooke:
It sounds like some more thought should be given to how we would want it to work, or if it would work at all, so probably not in this PR.

Offhand, I can't think of how we would implement it without just yelling at the user for fields they haven't even tried to input yet. Conversely, following your suggestion to only validate required fields on mount that already have a value, it could still be confusing or send the wrong signal to those who are almost ready to submit, but are missing a required field - only, they can't see that when they load the form tab (as it would be specially omitted from the mount validation).

matthieu-foucault:
Note that there is a minor bug/missing feature, which may be confusing here: readonly fields should be validated on change otherwise the user may end up with the following form:

image

kriscooke:
Fields that shouldn't be validated because the user can't interact with them probably have no business being a real form input.

matthieu-foucault:
I decided to change the forms jsonschema instead, and to remove the "required" setting on values that are read-only, as removing them from the form data is potentially a breaking change

PR #960, opened 2020-07-27
Notice re: Supported browser versions

kriscooke:
Implements a short, dismissible blurb stating the supported browsers to use for the best user experience. This is implemented on the logged-in views for industry users only.

(Related, but incidental changes):
- Removes a no-longer-necessary 110px of whitespace padding the top of page content on viewports 992px and wider.
- Turns on on the subheader navigation and page title (like all other related pages) on the user profile page

[GGIRCS-1747](https://youtrack.button.is/issue/GGIRCS-1747)


kriscooke:
Superceded by #1163

PR #965, opened 2020-07-28
feat: redirect user to a more friendly page if keycloak auth fails

matthieu-foucault:


app/pages/403.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <Row className="justify-content-center" style={{paddingTop: '3em'}}>
+          <Col md={{span: 6}} style={{textAlign: 'center'}}>
+            <h1>Something went wrong</h1>
+            <p>
+              If you got here from the registration page, this is an issue that
+              we are actively working to fix. We apologize for the
+              inconvenience.
+              <br />
+              You should be able to log in with the credentials you entered in
+              the registration form.
```
kriscooke, line 22: Changing this:

> You should be able to log in with the credentials you entered in the registration form.

to:

> If you click "Login" above, you can now log in with the credentials you entered in the registration form.


...would make the suggested solution sound more definitive and tells them exactly what to do next (click Login).

kriscooke: Otherwise, I think this is good to go!

matthieu-foucault: @kriscooke There might be other ways to end up on this page if keycloak fails. I updated the text following your recommendation

PR #973, opened 2020-07-29
feat: authenticate user when landing on index page if logged in KC

matthieu-foucault:
This can be tested locally by:
1) logging out of CIIP
2) logging in the keycloak realm at https://sso-dev.pathfinder.gov.bc.ca/auth/admin/pisrwwhx/console/#/realms/pisrwwhx
3) open/refresh the CIIP index page, and you should be automatically logged in.

Prior to this feature, the user has to click on the login button.

### DIFF
--- a/app/server/index.js
+++ b/app/server/index.js
@@ -393,6 +393,9 @@ app.prepare().then(() => {
 
   server.get('/register', ({res}) => res.redirect(302, kcRegistrationUrl));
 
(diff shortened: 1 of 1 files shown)


matthieu-foucault:
Bug was unscheduled following unsuccessful attempt at fixing. Closing this PR, as I am not sure we'll ever get around to fixing this.

PR #990, opened 2020-08-04
Feat: guest accessibility

dleard:
Index:
- wrap content in default-layout in a div with role=main
https://dequeuniversity.com/rules/axe/3.4/landmark-one-main?application=axeAPI
- set top level h tag to h1
https://dequeuniversity.com/rules/axe/3.4/page-has-heading-one?application=axeAPI
- increase lower level h tags to h2/h3
https://dequeuniversity.com/rules/axe/3.4/heading-order?application=axeAPI
- add alt text to all img tags
https://dequeuniversity.com/rules/axe/3.4/image-alt?application=axeAPI
- increase contrast for a tag
https://dequeuniversity.com/rules/axe/3.4/color-contrast?application=axeAPI

Disclaimer:
- add a role & aria-label to default-layout title div
https://dequeuniversity.com/rules/axe/3.4/region?application=axeAPI

Privacy/Contact/Copyright:
- increase contrast for a tags
https://dequeuniversity.com/rules/axe/3.4/color-contrast?application=axeAPI


app/pages/_document.tsx
```diff
@@ -9,41 +9,46 @@ class MyDocument extends Document {
...
-            color="#036"
-          />
-          <link rel="icon" href="/static/icons/bcid-favicon-32x32.png" />
-        </Head>
-        <body>
-          <Main />
-          <NextScript />
-          {CONFIG.ENABLE_ANALYTICS === 'true' && <Analytics />}
-        </body>
+        <html lang="en">
```
kriscooke, line 12: Is there a reason for adding a second additional `<html>` tag here? There should be just [one per document](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html).

dleard: I thought it was giving me a typescript error, but I think it was just my editor being really slow on validation

app/pages/index.tsx
```diff
@@ -239,6 +240,16 @@ export default class Index extends Component<Props> {
           #photo {
             margin: 53px 0;
           }
+          h1 {
+            font-size: 1.75rem;
+            margin-bottom: 15px;
```
kriscooke, line 245:
```suggestion
            margin-bottom: 20px;
```

app/pages/index.tsx
```diff
@@ -239,6 +240,16 @@ export default class Index extends Component<Props> {
...
+          }
+          h2 {
+            font-size: 1.5rem;
+          }
+          h3 {
+            font-size: 1.24rem;
+          }
```
kriscooke, line 252:
```suggestion
          }
          h5 {
            margin-bottom: 0.5rem;
          }
```

app/pages/index.tsx
```diff
@@ -239,6 +240,16 @@ export default class Index extends Component<Props> {
...
             margin: 53px 0;
           }
+          h1 {
+            font-size: 1.75rem;
+            margin-bottom: 15px;
+          }
+          h2 {
+            font-size: 1.5rem;
+          }
+          h3 {
```
kriscooke, line 250:
```suggestion
          .value-prop h3 {
```

dleard:
Ok cool, I was just looking at fixing that, but it's a pretty minor amount of whitespace added. Thanks for the detailed review

kriscooke:
It is pretty painful to layer on accessibility after the fact, so hopefully the tests can help prevent us from going through it again. That's the purpose of the screenshot utility though - ability to keep the visual interface of a launched application stable while making underlying changes.

I think we get a bit de-sensitized to the expectation there's often a diff we will ignore due to the "Cypress typing bug" - but we should always be watching out and trying to avoid any diff unrelated to that (and which isn't intentionally part of the feature/fix added).

PR #1013, opened 2020-08-10
[WIP] Accessibility, Reporter scope

kriscooke:


app/components/PaginationBar.tsx
```diff
@@ -77,7 +77,7 @@ export const PaginationBarComponent: React.FunctionComponent<Props> = (
         <Pagination.First onClick={() => handlePaginationByPageNumber(1)} />
         <Pagination.Prev onClick={previousTenPagination} />
         {startPage !== 1 && <Pagination.Ellipsis />}
-        <Pagination>{items}</Pagination>
+        {items}
```
kriscooke, line 80: The examples for the Bootstrap Pagination component show only `<Pagination.Subcomponent>`s as children of a `<Pagination>` component - not nesting `<Pagination>` instances within each other.

Reason for change: This was causing there to be a nested `<ul>` as a direct child of another `<ul>`; the accessibility rules permit only `<li>` elements to be direct children of a `<ul>` or `<ol>`.

app/components/SearchBox.tsx
```diff
@@ -34,24 +34,23 @@ const SearchBox: React.FunctionComponent<Props> = ({
     type: 'object',
     properties: {
       searchField: {
+        title: 'Search By:',
         type: 'string',
         enum: dropdownSortItems
       },
       searchValue: {
+        title: 'Search Term:',
```
kriscooke, line 42: Sets `<label>`s for SearchBox's "search by" dropdown and search term `<input/>`s 
(rule: all inputs must have a corresponding label)

app/components/Layout/Header.tsx
```diff
@@ -32,7 +32,7 @@ const HeaderLayout = ({isLoggedIn = false, isRegistered = false}) => {
 
   return (
     <header>
-      <nav>
+      <nav aria-label="Main">
```
kriscooke, line 35: Rule: If multiple `<nav>` (aka. a page / accessibility landmark) elements exist, they must have an `aria` attribute set which labels them as distinct. 

When navigating via a screenreader, it's possible to go direct to a `<nav>` and this will let it announce the two `<nav>`s (one here, one in Subheader) by their distinct name.

app/components/SearchBox.tsx
```diff
@@ -34,24 +34,23 @@ const SearchBox: React.FunctionComponent<Props> = ({
...
         type: 'string'
       }
     }
   };
 
   const uiSchema = {
     searchField: {
-      'ui:placeholder': 'Search By: ',
```
kriscooke, line 48: When things are labelled by using the input's `placeholder` attribute, this isn't recognizable as a label by a screenreader. The best practice is to use a `<label>` element as shown above.

app/components/SearchBox.tsx
```diff
@@ -34,24 +34,23 @@ const SearchBox: React.FunctionComponent<Props> = ({
...
-      'ui:col-md': 5,
-      classNames: 'hide-title'
+      'ui:col-md': 5
     },
     searchValue: {
-      'ui:col-md': 7,
-      classNames: 'hide-title'
```
kriscooke, line 54: The `.hide-title` class would hide the [`<label>`s required to be set above](https://github.com/bcgov/cas-ciip-portal/pull/1013/files#diff-015f9994a81656a7479a8042902172d0R37) using `display: none;`. 

(Visual hiding of DOM objects using CSS also hides those elements from screenreaders.)

app/components/SearchBox.tsx
```diff
@@ -76,16 +75,6 @@ const SearchBox: React.FunctionComponent<Props> = ({
           </ButtonGroup>
         </div>
       </JsonSchemaForm>
-      {/* Hide searchField / searchValue JsonSchemaForm titles  */}
-      <style jsx global>
-        {`
-          .hide-title label.form-label {
-            display: none;
-          }
-        `}
```
kriscooke, line 85: In addition to hiding the `<label>`s mentioned above (in SearchBox), this was actually polluting the global scope and hiding other `<label>`s associated with other `<input>`s.

It's no longer necessary, since we need to show labels.

app/containers/Applications/ApplicationWizardConfirmation.tsx
```diff
@@ -306,9 +306,9 @@ export const ApplicationWizardConfirmationComponent: React.FunctionComponent<Pro
   return (
     <>
       <h1>Summary of your application:</h1>
-      <h5>
+      <p style={{fontSize: '1.25rem', fontWeight: 500}}>
         Please review the information you have provided before continuing.
-      </h5>
+      </p>
```
kriscooke, line 311: Rule: Semantic/hierarchical Headings.

Changed to a paragraph because the text content doesn't make semantic sense as a heading. 

(ie: you wouldn't use this text to denote a page section that can be navigated to directly in a screenreader. The heading "Summary of your application" just above it would be used instead).

app/containers/Applications/ApplicationWizardConfirmation.tsx
```diff
@@ -188,7 +188,7 @@ export const ApplicationWizardConfirmationComponent: React.FunctionComponent<Pro
               />
             </Form.Group>
             <Button
-              variant="info"
+              variant="primary"
```
kriscooke, line 191: Changes the button color to one with AA contrast ratio.

app/containers/Forms/EmissionGasFields.tsx
```diff
@@ -48,9 +48,16 @@ const EmissionGasFields: React.FunctionComponent<FieldProps> = ({
...
+              margin: 0,
+              padding: 0,
+              color: '#797676',
+              lineHeight: '17px'
+            }}
           >
-            <small>{formData.gasDescription}</small>
+            <label htmlFor={`${idSchema?.$id}_annualEmission`}>
+              <small>{formData.gasDescription}</small>
+            </label>
```
kriscooke, line 60: Rule: all inputs must have a label.

The content of `<small>` labels a corresponding `<input>`, so it's wrapped with a `<label>` element accordingly. 

(Side note: `htmlFor` is one of those React-specific attribute names - like `className` - to avoid clashing with reserved JS keywords. In plain HTML, this would be a `for` attribute like so: `<label for="id-of-input-this-labels">`.

app/containers/Forms/EmissionSourceFields.tsx
```diff
@@ -8,16 +8,16 @@ const EmissionSourceFields: React.FunctionComponent<FieldProps> = ({
...
-        <h6>Tonnes</h6>
+        <p className="unit">Tonnes</p>
       </Col>
       <Col xs={12} md={2}>
-        <h6>GWP</h6>
+        <p className="unit">GWP</p>
       </Col>
       <Col xs={12} md={3}>
-        <h6>Tonnes (CO2e)</h6>
+        <p className="unit">Tonnes (CO2e)</p>
```
kriscooke, line 20: Rule: semantic headings

These `h6`s have been changed to paragraphs because, as unit names, they did not delineate content sections of the page. (ie: sections that you'd navigate directly to through a screenreader menu)

app/containers/Forms/EmissionSourceFields.tsx
```diff
@@ -31,12 +31,17 @@ const EmissionSourceFields: React.FunctionComponent<FieldProps> = ({
         }
       `}</style>
       <style jsx>{`
-        h6 {
+        .unit {
           text-align: center;
```
kriscooke, line 35: Corresponds to changing the above tags from `h6` --> `p` while accounting for their specific styling.

app/containers/Forms/SummaryFormFieldTemplate.tsx
```diff
@@ -118,9 +118,6 @@ const FormFieldTemplate: React.FunctionComponent<FieldTemplateProps> = ({
           padding-top: 0;
           padding-bottom: 50px;
         }
-        .summary-card .card-header h2 {
-          font-size: 30px;
-        }
```
kriscooke, line 123: This was polluting styles located just outside the scope of this form field component. It's effectively been moved to the correct location.

app/containers/Organisations/Organisations.tsx
```diff
@@ -148,10 +148,10 @@ export const OrganisationsComponent: React.FunctionComponent<Props> = (
               </>
             ) : (
               <Dropdown className="search-dropdown">
-                <Card.Title>
+                <h2>
                   Request access to apply for the CIIP on behalf of a Reporting
                   Operation:
-                </Card.Title>
+                </h2>
```
kriscooke, line 154: Unfortunately headings accessibility won't always let us blindly use helpers like `<Card.Title>`, which arbitrarily sets it to a particular heading number (which may not be the number you need).

schema/data/prod/form_json/administration.json
```diff
@@ -2,8 +2,9 @@
   "schema": {
     "type": "object",
     "properties": {
-      "comments": {
-        "type":"string"
+      "adminComments": {
+        "type":"string",
+        "title": "comments"
```
kriscooke, line 7: Rule: Each id must be unique on the page.

The property names are used by react-jsonschema-form to set ids on some of the corresponding form elements (prefixed by `root_`). Because the comments property was called "comments" for each form tab (administration, emission, fuel, production) this was causing it to create many duplicate `root_comments` ids.

kriscooke:
@matthieu-foucault We could use your help debugging this (potentially async) issue causing the production form test in `application-form-validation.spec.js` test to fail. We found that it was fixed locally by adding a wait period in `beforeEach`, however that supposed fix does not translate to CI. 

The exact problem causing the test to fail is that Cypress is unable to click an item in the production form dropdown list. At the time of the Cypress interaction, it seems there are possibly no items to populate the list - hence it has the `disabled` attribute and contains text that says something like "No matching results"(?)

What's especially confusing is that it's not clear how the changes in this PR might relate to this. 🤔 

```
  3 passing (4m)
  1 failing

  1) When reviewing a submitted application as an analyst
       The application production form shows validation errors:
     CypressError: Timed out retrying: `cy.click()` failed because this element:

`<a class="dropdown-item disabled" href="#">No matc...</a>`

has CSS `pointer-events: none`

`pointer-events: none` prevents user mouse interaction.

Fix this problem, or use {force: true} to disable error checking.

https://on.cypress.io/element-cannot-be-interacted-with
```

app/components/Layout/Header.tsx
```diff
@@ -33,7 +33,7 @@ const HeaderLayout = ({isLoggedIn = false, isRegistered = false, children}) => {
   return (
     <header>
       <div className="nav-header">
-        <nav>
+        <nav aria-label="Main">
```
kriscooke, line 36: Rule: If multiple `<nav>` (aka. a page / accessibility landmark) elements exist, they must have an `aria` attribute set which labels them as distinct. 

When navigating via a screenreader, it's possible to go direct to a `<nav>` and this will let it announce the two `<nav>`s (one here, one in Subheader) by their distinct name. Announced as: "Main navigation", "Secondary navigation" as labelled.

schema/data/prod/form_json/administration.json
matthieu-foucault: This is a breaking change, as it would cause comments from already submitted applications to not be rendered anymore. To fix that a11y issue, you should use the idPrefix prop instead: https://react-jsonschema-form.readthedocs.io/en/v1.8.1/advanced-customization/#id-prefix

matthieu-foucault:
Looking into the e2e test issue. In the meantime, my comment above is something that can be worked on.

matthieu-foucault:
@kriscooke @dleard The e2e test issue is not an async issue, it is a side effect of the `reporter-access-setup` fixture, which is now executed in the a11y test, before the `application-form-validation` test, and truncating the `products` table. 
To fix this, the `application-form-validation` test should have a `cypress.sqlFixture('dev/product')` in a `before` hook.

schema/data/fixtures/reporter-all-access-teardown.sql
```diff
@@ -1,5 +1,8 @@
 begin;
 
+-- Re-populate products table, which was truncated in test setup:
+\i ../schema/data/dev/product.sql
+
```
kriscooke, line 5: This path is relative to the folder from which the `cypress run` command is run, which is the same in all environments thanks to npm/yarn scripts. I recognize it's kind of ugly, because it doesn't make a lot of sense when viewed alone as part of this fixture.

We can either leave the path hard-coded like this (which seems a bit gross, but has the advantage of being very straightforward), or we can use a variable like so:
```
\set product_seeds :DATA_DIR '/dev/product.sql'
\i :product_seeds
```
...by adding a variables flag to the psql script in `cypress/support/commands.js`:
```
psql -v DATA_DIR=$(pwd)/../schema/data -d ciip_portal_dev
```


What do you think @matthieu-foucault?

matthieu-foucault: I don't think that the teardown should be responsible for re-populating the products table, which other tests may or may not need. We can leave it as-is for now, but as @dleard mentioned offline, there is a need for rewriting our e2e fixtures. We're currently assuming that the `dev` data is deployed when starting a test, but I think that each e2e test file should assume an empty database, and load the fixtures it needs in a `before`/`beforeEach` hook

matthieu-foucault: I will be adding a card in our issue tracker for that fixture refactor

kriscooke: I suppose I was operating on the principle that tests should be independent of each other - teardown should undo what's done in setup. There are other test suites that use this setup/teardown, so the idea was to fix the side effect each place it's used.

But certainly agreed that they all need to be re-conceived and re-organized.

matthieu-foucault: Yes, tests should not have side effect, but I would add that they should not assume that the database contains any data. We currently have a fair amount of data added before we start cypress via https://github.com/bcgov/cas-ciip-portal/blob/3e42f82e4fafbf83c130290f274ade7051727fa8/.circleci/config.yml#L380
which is not a great practice and will be changed when we tackle https://youtrack.button.is/issue/GGIRCS-1908

PR #1015, opened 2020-08-11
fix: drop happo version number

dleard:


dleard:
This was a test to see if happo being down was a result of the patch, which it was not

PR #1020, opened 2020-08-13
Feat: pending certification state

dleard:
adds a (Pending Certification) note to the draft state of an application if a certifierUrl exists for that draft.


app/containers/Facilities/FacilitiesRowItemContainer.tsx
```diff
@@ -26,6 +26,15 @@ export const FacilitiesRowItemComponent: React.FunctionComponent<Props> = ({
   query
 }) => {
   const {applicationRevisionStatus} = facilitySearchResult;
+  const isPendingCertification = Boolean(
```
matthieu-foucault, line 29: would this be better renamed to `hasCertificationUrl`, since an application is only pending certification when it has a url *and* is a draft

PR #1038, opened 2020-08-20
Feat: empty search table result message

dleard:
- adds a message if the search table result is empty
- adds some unit tests for SearchTableLayout

### DIFF
--- a/app/components/SearchTableLayout.tsx
+++ b/app/components/SearchTableLayout.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {Container, Row, Col, Table} from 'react-bootstrap';
+import {Container, Row, Col, Table, Alert} from 'react-bootstrap';
 import SortableTableHeader from 'components/SortableTableHeader';
 import SearchBox from 'components/SearchBox';
 interface Props {
@@ -21,6 +21,17 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
     body,
     isLoading
   } = props;
+
+  const noSearchResults =
+    body.props.children.length === 0 ? (
+      <span id="no-search-results">
+        <Alert variant="secondary">
+          <p>&quot;No matching results to show.&quot;</p>
+          <p>Clicking &apos;Reset&apos; will clear your search parameters.</p>
+        </Alert>
+        <br />
+      </span>
+    ) : null;
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


app/components/SearchTableLayout.tsx
```diff
@@ -22,6 +22,17 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
     isLoading
   } = props;
 
+  const noSearchResults =
+    body.props.children.length === 0 ? (
+      <span id="no-search-results">
+        <Alert variant="secondary">
+          <p>&quot;No matching results to show.&quot;</p>
+          <p>Clicking &apos;Reset&apos; will clear your search parameters.</p>
```
matthieu-foucault, line 30: Did we get a design done for this feature. I believe one of the recommendations from @rajpersram, to be implemented in the future, was to remove the `Reset` button: https://youtrack.button.is/issue/GGIRCS-1626

dleard: No, there was no specific design done for this. There was just a suggestion of what to show in the card, I added the 'Reset button' instruction and it could definitely come out.

dleard:
@matthieu-foucault  There is some probably technical debt here that @kriscooke and I have been discussing in teams:

I have been using the length of `children` in SearchTableLayout to determine if there is anything to show, or if it should show the 'no results..' message. The CertifierRequestsContainer contains all the code for the rows, whereas the other search containers factor the row out into a fragment container.

`the other search tables follow a "list container" -> "row item container" hierarchy, and the CertifierRequests is all in one "list container". I think it could be factored out so that each row has its own fragment container, a la "FacilitiesListContainer" -> "FacilitiesRowItemContainer", but it was a more complicated list container than any of the other ones with the select all functionality. An extra conditional fixes it for now without having to do a full refactor on the CertifierRequests container. Adding a children[0] specifically to handle that component is not ideal, but it is the quick fix.`

I guess the question is, should we go with the extra conditional in SearchTableLayour specific to CertifierRequests and add a refactor tech debt card for CertifierRequests? Or should the refactoring happen now?

app/components/SearchTableLayout.tsx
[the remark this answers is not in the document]
dleard: :+1: I'll remove it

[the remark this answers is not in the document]
dleard: It looks like those buttons have been moved up so that they are no longer aligned with the search boxes. (Done somewhere in develop, not in this PR.) I've made a bug card & added it to the backlog https://youtrack.button.is/issue/GGIRCS-1919

matthieu-foucault:
@dleard you can add a techdebt card. I wouldn't tackle refactoring these tables now

dleard:
Tech debt card for refactoring the CertifierRequestsContainer: https://youtrack.button.is/issue/GGIRCS-1920

app/components/SearchTableLayout.tsx
```diff
@@ -22,6 +22,17 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
...
 
+  const noSearchResults =
+    body.props.children.length === 0 ||
+    body?.props?.children[0]?.length === 0 ? (
+      <span id="no-search-results">
+        <Alert variant="secondary">
+          <p>&quot;No matching results to show.&quot;</p>
+        </Alert>
+        <br />
+      </span>
```
kriscooke, line 33: - The quotes are unnecessary. (sorry - I can see this was probably a result of a rather literal interpretation of what I wrote in the card 😊)

- To remove some of the awkward extra whitespace shown below, this whole span block can be simplified to:
```
    <Alert variant="secondary">
      No matching results to show.
    </Alert>
```
Only the [Alert and simple text content is needed](https://react-bootstrap.netlify.app/components/alerts/) - the span (which is an [inline element](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#Inline_vs._block-level_elements_a_demonstration)) is unnecessary and can be removed:

before:
Screen Shot 2020-08-25 at 9 06 28 AM

after:
Screen Shot 2020-08-25 at 9 07 32 AM

kriscooke (dismissed):

PR #1071, opened 2020-08-28
Feat: reporter should be able to start/submit revisions to an application when the application window is closed

matthieu-foucault:


app/containers/Applications/ApplyButtonContainer.tsx
```diff
@@ -53,11 +53,7 @@ export const ApplyButton: React.FunctionComponent<Props> = ({
   ]);
 
   if (!canOpenApplication) {
-    return (
-      <Button disabled variant="info">
-        The application window is closed
-      </Button>
-    );
+    return null;
```
kriscooke, line 56: Returning a (not visible, empty div) in this case would prevent the last table column from collapsing when there are no buttons present in that column. (The column's width is determined by its contents):

```suggestion
    return (
      <div style={{min-width: '9.7em'}} />
    );
```

kriscooke:
@matthieu-foucault The timed out Happo job is probably something we experienced before (also just before deployment), and which was fixed relatively quickly with correspondence to Happo - they were having some internal error in the Internet Explorer workers, which wasn't being alerted in their logs.

I wrote to support yesterday eve, and they've fixed it overnight (European support FTW 😄), so the next job should complete.

There's no interface to re-trigger existing jobs, but you can make some fake amended commit or rebase and push to trigger a new job.

matthieu-foucault:
@kriscooke I restarted the CI workflow, let's see if that fixes it

PR #1078, opened 2020-09-01
fix: keycloak access token is refreshed when needed

matthieu-foucault:
In order to test this PR, you should reduce the `SSO Session Idle` setting in keycloak to 2/3 minutes, and the `Access token lifespan` to something smaller.
You should not encounter any issue as long as you keep interacting with the application (contrary to the current behavior where you get logged out either way).
If you wait for the time you set in the SSO Session Idle, you should get redirected to the `login-redirect` page, asking you to log in again


dleard (approved):
Tested & works!

dleard (approved):

PR #1080, opened 2020-09-02
fix: Always show reviewer comments regardless of revision status

kriscooke:
Fixes [GGIRCS-1948](https://youtrack.button.is/issue/GGIRCS-1948)

### DIFF
--- a/app/containers/Applications/ApplicationCommentsContainer.tsx
+++ b/app/containers/Applications/ApplicationCommentsContainer.tsx
@@ -57,8 +57,8 @@ export const ApplicationCommentsComponent: React.FunctionComponent<Props> = (
                   if (
                     !review &&
                     formResult.applicationByApplicationId
-                      .applicationRevisionStatus.applicationRevisionStatus !==
-                      'REQUESTED_CHANGES'
+                      .applicationRevisionStatus.applicationRevisionStatus ===
+                      'SUBMITTED'
                   )
                     return null;
                   return (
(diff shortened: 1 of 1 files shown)


app/containers/Applications/ApplicationCommentsContainer.tsx
```diff
@@ -54,13 +54,6 @@ export const ApplicationCommentsComponent: React.FunctionComponent<Props> = (
               <tbody>
                 {formResult.internalGeneralComments.edges.map(({node}) => {
                   if (node.resolved && !showResolved) return null;
-                  if (
-                    !review &&
-                    formResult.applicationByApplicationId
-                      .applicationRevisionStatus.applicationRevisionStatus !==
-                      'REQUESTED_CHANGES'
```
dleard, line 61: Now that I've looked into this a little further, I remember the reason we had this code block. It is so that the reporter can't see the comments until the analyst is finished with the review and has changed the status. So the problem isn't the entire code block, it's the logic.

Fix: change 'REQUESTED_CHANGES' to 'SUBMITTED'. This will block the reporter from seeing applications whose status has not been changed by the analyst to one of the "actionable" states rather than blocking comments from all states except 'REQUESTED_CHANGES'

kriscooke: Good catch - tricky part is, SUBMITTED will not help us in showing the comments on a new revision, as it will be DRAFT. Need a different condition here

kriscooke: I think what we need is to check whether the current revision's _immediately prior_ revision has REQUESTED_CHANGES... 🤔

dleard: It's a `does not equal` comparison. So applications in draft status will be able to see the comments

dleard: whoops, I mean change the !== 'Requested_Changes' to === 'Submitted'. My bad, that's important

dleard: So if the application is in 'SUBMITTED' status (ie reviewer still needs to change the status), then the comments don't show up. But for any other status (including draft) they will be visible

dleard (approved):

dleard (approved):

PR #1103, opened 2020-09-09
Feat: readable dates

dleard:
- Change UTC format dates to readable format

### DIFF
--- a/app/functions/formatDates.ts
+++ b/app/functions/formatDates.ts
@@ -1,0 +1,23 @@
+import moment from 'moment-timezone';
+
+const TIMEZONE = 'America/Vancouver';
+const FORMAT_TYPE = {
+  seconds: 'MMM D, YYYY hh:mm:ss A (z)',
+  minutes: 'MMM D, YYYY hh:mm A (z)',
+  hours: 'MMM D, YYYY hh A (z)',
+  days_numbered: 'DD-MM-YYYY',
+  days_string: 'MMM Do, YYYY',
+  database_timestamp: 'YYYY-MM-DDTHH:mm:ss'
+};
+
+export const dateTimeFormat = (dateTime: string, format: string) => {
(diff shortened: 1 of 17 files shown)


kriscooke (commented):
I think we should put off a decision on this until Raj returns next week (see comment in [GGIRCS-1620](https://youtrack.button.is/issue/GGIRCS-1620)

matthieu-foucault (changes requested):
Can we have a single file with `dateFormat` and `dateTimeFormat` functions in it? 
This is the best way to ensure consistency across the application, instead of copy/pasting the time zone and format strings.
`moment.tz` probably shouldn't be used directly by other files unless there's some date comparison needed (e.g in `ReportingYearTable` or `KeyDates`)

kriscooke:
Agreed, and in that file we may need a couple of possible date formats (eg: date only, date with timestamp etc.)

dleard:
@kriscooke @matthieu-foucault I removed the moment imports and TIME_ZONE/DATE_FORMAT variables from all files but AltDateInput.tsx & replaced them with calls to the functions in formatDates.ts

app/functions/formatDates.ts
```diff
@@ -0,0 +1,23 @@
+import moment from 'moment-timezone';
+
+const TIMEZONE = 'America/Vancouver';
+const FORMAT_TYPE = {
+  seconds: 'MMM D, YYYY hh:mm:ss A (z)',
+  minutes: 'MMM D, YYYY hh:mm A (z)',
+  hours: 'MMM D, YYYY hh A (z)',
+  days_numbered: 'DD-MM-YYYY',
+  days_string: 'MMM Do, YYYY',
+  database_timestamp: 'YYYY-MM-DDTHH:mm:ss'
```
kriscooke, line 10: It doesn't look like `database_timestamp` and `hours` formats are currently used anywhere - maybe we should just start with the minimum required number of formats and add any others as they become necessary?

kriscooke (approved):

PR #1114, opened 2020-09-14
fix: Hide gases button shouldn't overlap 'saving' indicator

kriscooke:
Fixes "Hide gases with no emissions" button from overlapping the form 'Saving' indicator.
[GGIRCS-1925](https://youtrack.button.is/issue/GGIRCS-1925)

### After:
Screen Shot 2020-09-14 at 11 36 49 AM

### Before: 
93124549-a1422400-f67e-11ea-9e79-35c78a17eae4

### DIFF
--- a/app/containers/Forms/EmissionField.tsx
+++ b/app/containers/Forms/EmissionField.tsx
@@ -35,8 +35,6 @@ export const EmissionFieldComponent: React.FunctionComponent<FieldProps> = (
         {`
           .emission-toggle {
             text-align: right;
-            position: relative;
-            top: -52px;
           }
           .emission .zero-emission,
           .emission-form .zero-emission {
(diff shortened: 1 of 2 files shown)


dleard:
@kriscooke looks good to me. Just needs a rebase.

app/containers/Forms/EmissionField.tsx
```diff
@@ -36,7 +36,6 @@ export const EmissionFieldComponent: React.FunctionComponent<FieldProps> = (
           .emission-toggle {
             text-align: right;
             position: relative;
```
matthieu-foucault, line 38: Is the `position:relative` still needed?

kriscooke: Excellent point - no, I'm sure it's not. Can remove

matthieu-foucault: Yes please. I think it would be confusing to see this css rule which has no effect.

matthieu-foucault (approved):

PR #1116, opened 2020-09-14
Feature: save & exit/continue buttons

dleard:
- Update Form buttons to match UX suggestions in https://youtrack.button.is/issue/GGIRCS-1625

### DIFF
--- a/app/containers/Forms/FormSharedStyles.tsx
+++ b/app/containers/Forms/FormSharedStyles.tsx
@@ -14,7 +14,7 @@ export default css.global`
   .form-submit {
     border: 1px solid #bbb;
     padding: 30px;
-    background: #eee;
+    background: #e0e0e0;
     border-radius: 6px;
   }
   .errors {
@@ -32,5 +32,10 @@ export default css.global`
     background: #ce5c5c;
     color: white !important;
     padding: 5px;
+  }
+  .exit-button {
+    background: white !important;
+    color: #003366;
+    margin-left: 5px;
   }
 `;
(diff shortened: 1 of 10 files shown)


kriscooke (changes requested):
"Save and Exit" button needs sufficient text color contrast on hover (in the screenshot, I'm hovering over the button):

Screen Shot 2020-09-15 at 10 17 37 AM

app/containers/Forms/FormSharedStyles.tsx
```diff
@@ -33,4 +33,9 @@ export default css.global`
     color: white !important;
     padding: 5px;
   }
+  .exit-button {
+    background: white !important;
```
matthieu-foucault, line 37: not a huge fan of the `!important`. Maybe an inline style on the `Button` element is a better alternative? This would be a temporary solution until we explore using the bcgov boostrap theme.

kriscooke:
@dleard I think part of the button CSS issue is that it would be easier to use an [outline button](https://react-bootstrap.github.io/components/buttons/#outline-buttons) like`outline-secondary`:
Screen Shot 2020-09-15 at 2 08 32 PM

dleard:
@kriscooke Thanks for finding that :hover issue, I'll make some changes to the :hover attribute. For the outline-*, I can definitely use that, but Raj's design called for a white background, so I'd still have to manually edit that. The transparent background of outline-* buttons puts the text directly on the grey background, which seems like an accessibility issue?

@matthieu-foucault I can move the background color to inline to avoid !important on the .exit-button, but I'll still need to use !important to edit the :hover attribute to fix the issue Kristen found with the text disappearing, since I don't think :hover can be set inline.

matthieu-foucault:
@dleard You can leave some `!important` for now I suppose. We can revisit that solution when looking at using the bcgov bootstrap theme

kriscooke (approved):

PR #1132, opened 2020-09-22
docs: _status documentation

dleard:
- documents the reason behind creating _status tables, types and computed columns

### DIFF
--- a/docs/stack.md
+++ b/docs/stack.md
@@ -181,7 +181,10 @@ Snake case. Don&#39;t use reserved words (c.f. style test)
 
 ### \*status properties
 
-Use an enum, keep status rows immutables, add a status computed field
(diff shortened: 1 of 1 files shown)


matthieu-foucault (approved):
LGTM, just needs a rebase.
Do we have security policies that ensure that statuses are immutable? We should probably annotate the table with `@omit update,delete` (or something like that) so that postgraphile does not generate those mutations

dleard:
> LGTM, just needs a rebase.
> Do we have security policies that ensure that statuses are immutable? We should probably annotate the table with `@omit update,delete` (or something like that) so that postgraphile does not generate those mutations

Good idea, in fact all tables should have `@omit delete`. I've added a tech debt card to add all the appropriate omits to each table. https://youtrack.button.is/issue/GGIRCS-2011

matthieu-foucault (approved):

PR #1133, opened 2020-09-22
Steps for dev environment setup

pbastia:
see https://youtrack.button.is/issue/GGIRCS-1971

View [rendered markdown here](https://github.com/bcgov/cas-ciip-portal/blob/76643c9cffc330edad1232a9edceaebfbfc395c9/docs/dev-environment-setup.md) for easy reading.


docs/dev-environment-setup.md
```diff
@@ -0,0 +1,233 @@
...
+  * you can verify pre-commit works by running: `pre-commit run --all-files` from the project root.
+- You may need to troubleshoot individual steps in `make install_asdf_tools` on your system; in that case, find this target in the root `Makefile` and run the steps individually.
+- `asdf reshim` should be run after asdf installations to update symlinks for the installed packages.
+- Set the version of Postgres installed by asdf [as the global version](https://asdf-vm.com/#/core-manage-versions?id=set-current-version), necessary to prevent later problems installing Sqitch, our database migration tool.
+- `psql --version` should verify the installed version of Postgres.
+- [Yarn](https://yarnpkg.com/) is used as the Javascript package manager, and was installed in this step by asdf from the `.tool-versions`. All `yarn` commands must be run from the same directory as the `package.json` (within `app/`).
+
+**If you're on Linux**, you can now skip to [Step 4](#4-set-up-a-working-perl-5-environment).
+
+<!-- If using a BCGov Macbook, see: https://apps.nrs.gov.bc.ca/int/confluence/display/DEVGUILD/One-time+Setup? -->
```
kriscooke, line 31: I've left this in a comment for now until we can verify: After reading through the link from @pbastia and others, it doesn't seem like there is anything novel here that couldn't be deduced from following the [asdf documentation](https://asdf-vm.com/#/core-manage-asdf?id=install)?

If someone thinks it's helpful and can describe the link in a sentence, we can uncomment it.

docs/dev-environment-setup.md
```diff
@@ -0,0 +1,233 @@
...
+```
+pre-commit run --all-files
+```
+
+If you are impatient and your work is isolated to Javascript, it may be faster to run only the related formatter (`xo`), but it may not catch everything (such as the end-of-file fixer and trailing whitespace):
+
+```
+yarn format
+```
+
```
kriscooke, line 175: I've left @pbastia 's notes in a comment at the bottom of the doc for reference until we verify they're resolved by the steps in the full-text descriptions above.

docs/dev-environment-setup.md
```diff
@@ -0,0 +1,233 @@
...
+<!-- ### Start & configure psql server
+- https://www.postgresql.org/download/macosx/ for an installer
```
kriscooke, line 177: I've stuck with guiding the user to try to let asdf install Postgres (with XML plugin option) - given nothing else works, I presume they would be able to find the manual download on their own?

docs/dev-environment-setup.md
```diff
@@ -0,0 +1,233 @@
...
+- And, make sure the PERL5LIB environment variable is properly set (wonky in my case)
+
+```
+~/.bashrc:
+    export PERL5LIB=~/lib/perl5/lib/perl5/
+
+    PERL_LOCAL_LIB_ROOT="/Users/pbastian/lib/perl5${PERL_LOCAL_LIB_ROOT:+:${PERL_LOCAL_LIB_ROOT}}"; export PERL_LOCAL_LIB_ROOT;
+    PERL_MB_OPT="--install_base \"/Users/pbastian/lib/perl5\""; export PERL_MB_OPT;
+    PERL_MM_OPT="INSTALL_BASE=/Users/pbastian/lib/perl5"; export PERL_MM_OPT;
+```
```
kriscooke, line 197: I was at a bit of a loss to explain the other Perl env variables (other than `PERL5LIB`) - @pbastia can you confirm these were necessary and/or do you have a link to StackOverflow or somewhere this solution was described?

kriscooke (commented):
@pbastia Let me know if these steps describe what's in your notes, and/or if anything is left out.

I've rebased it to update it with `develop` (in case you notice something funky if you try to pull).

docs/dev-environment-setup.md
kriscooke: I think what I'm trying to get at is, how does this advice relate specifically to a Macbook provided by BCGov and would not relate to any other device?

pbastia: I think you're right, this is just standard asdf install. I don't think there is any need for this anymore

docs/dev-environment-setup.md
pbastia: Probably, there's a million ways of installing psql on OSX through the standalone app, or homebrew, or many others.
Since asdf is installing it for that particular version and application, does it make sense to use it vs having a single local postgres server for all apps?

docs/dev-environment-setup.md
pbastia: These are some local::lib configuration variables, https://metacpan.org/pod/release/GETTY/local-lib-1.006007/lib/local/lib.pm

docs/dev-environment-setup.md
kriscooke: > Since asdf is installing it for that particular version and application, does it make sense to use it vs having a single local postgres server for all apps?

One reason for using asdf is to ensure they install the specific version of Postgres - you'll notice in the `.tool-versions` we specify 11.4 (whereas Postgres is now up to 13).

I guess the other idea behind using it is that a developer may use their machine for other applications, for which they may use other package managers and other database versions.

That said, there's a step which [sets the asdf installed version of postgres as global](https://github.com/bcgov/cas-ciip-portal/pull/1133/files#diff-fb7bdbe3eb6696759bbe84ebb71578a5R25), for Sqitch to recognize the existing Postgres when installed instead of installing its own.

docs/dev-environment-setup.md
kriscooke: OK. I recall that part of the issue you were debugging here is that your newly installed Perl was installed in a folder nested within perl itself? `~/lib/perl5/lib/perl5/`

Do we know now how it got into that state? 

Just trying to figure out if/how we should generalize this advice and whether it's likely they will need it.

matthieu-foucault:
https://github.com/bcgov/cas-ciip-portal/blob/develop/docs/developer-guide.md may need to be reviewed/deleted in this PR (or in another PR soon) I think that there are things in that file that pertain to the dev setup, and other things that should be in `stack.md` (and obsolete things such as watchman)

docs/dev-environment-setup.md
```diff
@@ -0,0 +1,239 @@
...
+
+[pre-commit](https://pre-commit.com/) runs a variety of formatting and lint checks configured in [`.pre-commit-config.yaml`](https://github.com/bcgov/cas-ciip-portal/blob/develop/.pre-commit-config.yaml) which are required for a pull request to pass CircleCI.
+
+`pre-commit install` will [configure a pre-commit hook to run before every commit](https://pre-commit.com/#usage); alternatively, you can run it manually with:
+
+```
+pre-commit run --all-files
+```
+
+If you are impatient and your work is isolated to Javascript, it may be faster to run only the related formatter (`xo`), but it may not catch everything (such as the end-of-file fixer and trailing whitespace):
```
matthieu-foucault, line 176: TODO (I may do this today of next time I do tech debt work): XO is too slow, switch back to plain eslint

docs/dev-environment-setup.md
```diff
@@ -0,0 +1,239 @@
...
+    - If you're on **MacOS** and having trouble with Perl libraries ignoring your newly installed (and thus, properly functioning) Perl in favour of an older system Perl, you can use the [`$PERL5LIB` environment variable](http://mvp.kablamo.org/dependencies/perl5lib/) to point it to the desired location. For example, point it to the new Perl you installed with non-root / regular user permissions instead of the root-installed system Perl.
+
+## 5. Install pgTAP
+
+Ensure the postgres service is running with `pg_ctl start`.
+
+```
+cd schema && make install_pgtap
+```
+This clones and builds [pgTAP](https://pgtap.org/), the database unit testing library from source. The cloned repository will be left behind (visible with `git status`), which you can safely remove after the installation is complete.
```
matthieu-foucault, line 93: We should add the pgtap folder to our .gitignore

docs/dev-environment-setup.md
matthieu-foucault: We should update our tool versions to use postgres 12.2, as it is the version that is used in OpenShift

matthieu-foucault: Re - sqitch and having postgres set as global, is that only required for installation, or for running sqitch as well?

docs/dev-environment-setup.md
kriscooke: Although that would prevent the user from noticing and removing it, as it's no longer necessary? Not sure that matters much either way

docs/dev-environment-setup.md
```diff
@@ -0,0 +1,249 @@
...
+
+```
+cd app && yarn test
+```
+Front-end unit tests are heavily snapshot-based. Work that changes the DOM will result in a diff from the last accepted snapshot and cause related tests to fail. You can update the snapshots and review / accept the diff with `yarn test -u`.
+
+### End-to-end Tests with Cypress
+
+```
+cd app && yarn cypress
```
dleard, line 167: cypress requires that the server be running. `yarn cypress` by itself will start the cypress gui, but will show an error that no server is running.

docs/dev-environment-setup.md
pbastia: I think homebrew installed perl in `~/lib/perl5/` and then local::lib created its own `./lib/perl5/` folder in the local perl installation directory. CPAN manages this, so as long as it can find the installed modules things fit together I believe

docs/dev-environment-setup.md
pbastia: Could the `make install_pgtap` command remove that installation folder at the end of the script automatically?

docs/dev-environment-setup.md
kriscooke: I'm not entirely sure. It comes up because without doing so, Sqitch did not find the right Postgres and instead installed another copy of Postgres during its own installation. I think this also tells it which `psql` to use during runtime.

I installed via asdf, but there's a bit of description about this on the [Sqitch Homebrew formula](https://github.com/sqitchers/homebrew-sqitch#--with-postgres-support).

kriscooke: (also, new ticket for updating Postgres to 12: https://youtrack.button.is/issue/GGIRCS-2003)

docs/dev-environment-setup.md
kriscooke: Sounds like a good solution to me

docs/dev-environment-setup.md
```diff
@@ -0,0 +1,256 @@
+# Setting up the Development Environment
+
+## 1. Repository and Version Control
+
+- Ensure you have push access (most likely by being added to the **@bcgov/cas-developers** GitHub team) to [bcgov/cas-ciip-portal](https://github.com/bcgov/cas-ciip-portal).
```
matthieu-foucault, line 5: Also needs to be added to the bcgov/cas team, which is required for the CircleCI security context (needed to run some jobs which need access to our openshift cluster)

docs/dev-environment-setup.md
kriscooke: You're right, I guess that's not obvious. Adding, as well as a note to run with `yarn dev` and not bypassing auth.

docs/dev-environment-setup.md
kriscooke: Would you want to take that one on, @pbastia ?

pbastia: Sounds good! I created https://youtrack.button.is/issue/GGIRCS-2005 that I can take on

kriscooke:
> https://github.com/bcgov/cas-ciip-portal/blob/develop/docs/developer-guide.md may need to be reviewed/deleted in this PR (or in another PR soon) I think that there are things in that file that pertain to the dev setup, and other things that should be in `stack.md` (and obsolete things such as watchman)

@matthieu-foucault Agreed - it's not exactly equivalent, since it doesn't cover the initial dev environment setup - which is why we started a new doc, as we wanted to describe a setup exercise with explicit steps without getting too deep in detail.

I think that page pertains more to practices (things that come after Step 10 - after they're set up), and could be addressed on its own in another PR?

docs/dev-environment-setup.md
kriscooke: @pbastia I think this is resolved now?

pbastia: I think we've got everything in there now! we can delete the comments for sure

docs/dev-environment-setup.md
```diff
@@ -0,0 +1,219 @@
...
+## 8. Run Tests
+
+### Unit Tests with Jest
+
+```
+cd app && yarn test
+```
+Front-end unit tests are heavily snapshot-based. Work that changes the DOM will result in a diff from the last accepted snapshot and cause related tests to fail. You can update the snapshots and review / accept the diff with `yarn test -u`.
+
+### End-to-end Tests with Cypress
```
matthieu-foucault, line 171: Note for a future PR: some cypress tests require a local smtp server (e.g. mailhog) and the `.env` file to have some variables in there.

matthieu-foucault (approved):
Looks good to me. See my comment above re. mailhog

kriscooke (approved):

docs/dev-environment-setup.md
kriscooke: Good call - should be addressed by 9914307

PR #1157, opened 2020-09-29
Add AS_PENDING and AS_CERTIFIER flags to docs

kriscooke:
I was surprised to learn we had a new `AS_PENDING` dev flag, so noticed it needed to be added to the docs. `AS_CERTIFIER` was also missing, so added that as well.


matthieu-foucault (dismissed):
just needs a rebase

matthieu-foucault (approved):

PR #1159, opened 2020-09-30
Replace XO with Prettier + ESLint

matthieu-foucault:
This should allow us to us the corresponding VSCode plugins, as the performance of the XO one made it unusable.

I switched to the airbnb style, and added a couple of settings in the `.prettierrc` file which I would like to remove eventually, to use the defaults instead, but I didn't want this PR to affect every single file


app/.prettierrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  singleQuote: true,
+  trailingComma: 'none',
```
kriscooke, line 3: I tend to be pretty easygoing with formatting rules, but I would like to know what is so terrible about trailing commas? I personally appreciate how they minimize git diffs by not necessitating that the previous line be edited when adding a new properties to objects.

app/components/Application/ApplicationDetailsCardItemCustomFields.tsx
```diff
@@ -49,11 +49,11 @@ const customFields = (
           return (
             <>
               <span id={id && `${id}-diffFrom`} className="diffFrom">
-                {prevValue ? prevValue : <i>[No Data Entered]</i>}
+                {prevValue ?? <i>[No Data Entered]</i>}
               </span>
               &nbsp;---&gt;&nbsp;
               <span id={id && `${id}-diffTo`} className="diffTo">
-                {currentValue ? currentValue : <i>[No Data Entered]</i>}
+                {currentValue ?? <i>[No Data Entered]</i>}
```
kriscooke, line 56: This stuck out as a weird/unfamiliar syntax to me, but [it does seem to be quite useful](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator#Assigning_a_default_value_to_a_variable).

app/.prettierrc.js
matthieu-foucault: I have no issue with one or the other, just wanted to avoid changing every single file in this PR. I intend to remove this file in the next PR, and let prettier do its thing

app/components/Application/ApplicationDetailsCardItemCustomFields.tsx
matthieu-foucault: This and [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) were introduced in Typescript at some point last year. I'm happy it made its way to javascript

docs/dev-environment-setup.md
```diff
@@ -243,7 +250,7 @@ cd app && yarn test:e2e --spec cypress/integration/accessibility/*
 pre-commit run --all-files
 ```
 
-If you are impatient and your work is isolated to Javascript, it may be faster to run only the related formatter (`xo`), but it may not catch everything (such as the end-of-file fixer and trailing whitespace):
+If you are impatient and your work is isolated to Javascript, it may be faster to run only the linter and formatter (`eslint` and `prettier`), but it may not catch everything (such as the end-of-file fixer and trailing whitespace):
 
 ```
 yarn format
```
kriscooke, line 256: To be equivalent to the previous `yarn format`, I think this snippet should be changed to:
```suggestion
yarn lint && yarn format
```

kriscooke (approved):

pbastia (approved):

PR #1163, opened 2020-10-02
Display static info page for unsupported browsers

kriscooke:
- Server-side middleware uses [Bowser](https://github.com/lancedikson/bowser) to detect unsupported browsers based on Browserstack testing [documented in the README](https://github.com/bcgov/cas-ciip-portal/blob/develop/README.md#supported-browsers)
- Supported browsers continue through the normal request flow, while unsupported browsers are redirected to a static page: `/static/update-browser.html`
- HTML/CSS-only page has no interactivity except the hyperlinks to each of the browser download pages.
  - Register/Login links omitted from header, as it would only frustrate the user since they cannot use the site.
- Updates unsupported browsers based on double-check of README versions in Browserstack
  - Safari 10 unsupported due to login bug in [GGIRCS-1950](https://youtrack.button.is/issue/GGIRCS-1950)
  - Samsung Internet version in README was inaccurate: 11 works, but 9 broken due to GGIRCS-1950. 10 is not available for testing in Browserstack.
    - err on the side of "maybe it works?" for handling of Samsung 10
    - removed from README due to inability to test for certain
  - Remove mobile Chrome version from README (inaccurate, and Browserstack doesn't let us choose to test in detail).
- Supercedes #960 

IE11

[GGIRCS-1820](https://youtrack.button.is/issue/GGIRCS-1820)


app/server/index.js
```diff
@@ -180,6 +182,15 @@ app.prepare().then(() => {
     next();
   });
 
+  // Renders a static info page on unsupported browsers.
+  // Files in /static/ are excluded.
+  server.use(/^(?!\/static\/.+$).*/gi, (req, res, next) => {
```
matthieu-foucault, line 187: Sonarcloud is angry about using a regex here.
instead I would suggest using

```js
if (req.path.startsWith('/static/')) return next();`
```

app/static/update-browser.html
```diff
@@ -0,0 +1,112 @@
+<!DOCTYPE html>
+<html>
```
matthieu-foucault, line 2:
```suggestion
<html lang="en">
```
suggested by sonarcloud

app/static/update-browser.html
```diff
@@ -0,0 +1,112 @@
...
+      <h2>CleanBC Industrial Incentive Program</h2>
+    </header>
+    <main>
+      <div id="heading">
+        <h1>Update your browser</h1>
+        <p>Secure use of site functionality requires an updated browser. Choose one of the links below to learn how to update your browser.</p>
+        <p>For mobile, please visit your device's app store.</p>
+      </div>
+      <div id="browser-info">
+        <table class="table">
```
matthieu-foucault, line 77: The table should have a description: https://www.w3.org/TR/WCAG20-TECHS/H39

app/static/update-browser.html
```diff
@@ -0,0 +1,112 @@
...
+                  <img class="browser-logo" src="/static/icons/chrome.png" alt="Google Chrome"/>
+                </a>
+              </td>
+            </tr>
+          </tbody>
+        </table>
+      </div>
+    </main>
+  </body>
+</html>
```
matthieu-foucault, line 112: missing end-of-file new line

matthieu-foucault (approved):

matthieu-foucault (changes requested):
used the wrong option above

pbastia (dismissed):

app/static/update-browser.html
kriscooke: It isn't so much an actual data table, as it is an oldschool layout device (like in email templates) for bad browsers.

In response to browsers improvement over time, we can normally avoid table layout hacks that are bad for accessibility, but it remains an issue of course for a page that needs to be very backwards-compatible.

In this case, I think the "table" content (the browser links) should be accessibly described both by their alt text.

kriscooke: It's interesting that Axe has never highlighted this as missing in all of our other tables 🤔

matthieu-foucault: This is one of the things that SonarCloud is complaining about.

matthieu-foucault: see https://sonarcloud.io/project/issues?id=bcgov_cas-ciip-portal&pullRequest=1163&resolved=false&types=BUG

kriscooke: Ohhh, I hadn't noticed. Very strange that's taking this opportunity to complain nonetheless...

kriscooke: Actually I guess SonarCloud doesn't render JSX and therefore only complains about this stuff for straight-up `.html` pages (of which there are zero others in this project).

matthieu-foucault (approved):

PR #1166, opened 2020-10-03
Manage Products styling updates

kriscooke:
In `/admin/products-benchmarks`:
- Left-aligns product names
- Capitalizes CIIP in "CIIP Benchmarked"
- Product status presented as a pill badge instead of plain text
- Vertically centers row field text
[GGIRCS-1628](https://youtrack.button.is/issue/GGIRCS-1628)

### DIFF
--- a/app/containers/Products/ProductRowItemContainer.tsx
+++ b/app/containers/Products/ProductRowItemContainer.tsx
@@ -389,6 +389,9 @@ export const ProductRowItemComponent: React.FunctionComponent<Props> = ({
           opacity: 0.5;
           cursor: pointer;
         }
+        .table td {
+          vertical-align: middle;
+        }
         `}
       </style>
     </>
--- a/app/containers/Products/ProductRowItemContainer.tsx
+++ b/app/containers/Products/ProductRowItemContainer.tsx
@@ -1,6 +1,6 @@
 import React, {useMemo} from 'react';
 import {graphql, createFragmentContainer, RelayProp} from 'react-relay';
-import {Modal, OverlayTrigger, Tooltip} from 'react-bootstrap';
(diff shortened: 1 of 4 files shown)


app/containers/Products/ProductRowItemContainer.tsx
```diff
@@ -376,6 +389,9 @@ export const ProductRowItemComponent: React.FunctionComponent<Props> = ({
           opacity: 0.5;
           cursor: pointer;
         }
+        table td {
+          vertical-align: middle !important;
+        }
```
matthieu-foucault, line 394:
```suggestion
        .table td {
          vertical-align: middle;
        }
```
You don't need the `!important`, just a higher specificity

app/containers/Products/ProductRowItemContainer.tsx
```diff
@@ -277,15 +277,28 @@ export const ProductRowItemComponent: React.FunctionComponent<Props> = ({
...
+    DRAFT: 'warning',
+    PUBLISHED: 'success',
+    ARCHIVED: 'secondary'
+  };
+
   return (
     <>
       <tr>
-        <td>{product.productName}</td>
+        <td style={{textAlign: 'left'}}>{product.productName}</td>
```
matthieu-foucault, line 292: use the `text-left` class instead?

app/containers/Products/ProductRowItemContainer.tsx
matthieu-foucault: Alternatively, use the `.align-middle` class (which does use an `!important`, so I'm not sure how I feel about it)

matthieu-foucault (changes requested):

matthieu-foucault (approved):

PR #1175, opened 2020-10-06
Position SearchTable header sort icons

kriscooke:
- Positions sorting icons in SearchTable header to right of column labels
- For consistency, "Select All" checkbox is also positioned to right of column label (eg: in `/certifier/requests`)
[GGIRCS-2017](https://youtrack.button.is/issue/GGIRCS-2017)
Screen Shot 2020-10-06 at 2 41 02 PM
Screen Shot 2020-10-06 at 2 26 41 PM

### DIFF
--- a/app/components/SearchTableLayout.tsx
+++ b/app/components/SearchTableLayout.tsx
@@ -64,7 +64,6 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
               <th className="master-select">
                 <label>
                   Select All
-                  <br />
                   <input
                     type="checkbox"
                     checked={allSelected}
@@ -100,6 +99,7 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
         }
         .master-select label {
           margin-bottom: 0;
+          padding-right: 24px;
           width: 100%;
           display: flex;
           flex-direction: column;
@@ -110,6 +110,11 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
           right: 0;
           bottom: 0;
           left: 0;
+        }
+        .master-select input[type='checkbox'] {
+          position: absolute;
+          right: 10px;
+          top: calc(50% - 6px);
         }
         .search-table.loading td {
           opacity: 0;
--- a/app/components/SortableTableHeader.tsx
+++ b/app/components/SortableTableHeader.tsx
@@ -16,13 +16,20 @@ const SortableTableHeader: React.FunctionComponent<Props> = ({
 }) => {
   const {columnName, displayName} = headerVariables;
   return (
-    <th
-      style={{background: '#003366', color: 'white'}}
-      onClick={() => sort('sortColumn', columnName)}
-    >
-      {displayName}
-      <br />
-      {displayName && <FontAwesomeIcon color="white" icon={faSort} />}
+    <th onClick={() => sort('sortColumn', columnName)}>
+      <span>{displayName}</span>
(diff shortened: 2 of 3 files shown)


matthieu-foucault:
@kriscooke In safari screenshot the checkbox for "select all" isn't aligned properly. Can you confirm that this is the case? https://happo.io/a/336/p/406/compare/8690490de0a5fbde96649d34c57cec28c451fb39/e886697483588863f00034a80929d91df874bfe3?t=diffs&component=Batch%20certification%20page&target=safari

app/components/SortableTableHeader.tsx
```diff
@@ -16,13 +16,20 @@ const SortableTableHeader: React.FunctionComponent<Props> = ({
...
-    <th
-      style={{background: '#003366', color: 'white'}}
-      onClick={() => sort('sortColumn', columnName)}
-    >
-      {displayName}
-      <br />
-      {displayName && <FontAwesomeIcon color="white" icon={faSort} />}
+    <th onClick={() => sort('sortColumn', columnName)}>
+      <span>{displayName}</span>
+      <span style={{height: '100%', position: 'absolute', right: '0.75em'}}>
```
matthieu-foucault, line 21: Doesn't a flexbox work, instead of using absolute/relative positioning?

app/components/SearchTableLayout.tsx
```diff
@@ -64,7 +64,6 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
               <th className="master-select">
                 <label>
                   Select All
-                  <br />
                   <input
```
matthieu-foucault, line 67: `input` should be a sibling of `label` (you're going to need an `id` and `htmlFor` for accessibility, which would allow you to use a flexbox for `master-select` (I see it's already a flexbox for some reason, it just needs to have `flex-direction: row`.

matthieu-foucault (changes requested):

app/components/SearchTableLayout.tsx
kriscooke: There's no expectation that `<input />` be a sibling of `<label>` instead of a child ([reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)). The purposes of making it a child are:
- clicking a `<label>` element natively causes the associated input to be checked/unchecked without the use of any custom/duplicate event listeners.
- avoids the need for specifying excess ids to associate the label with the input, as the parent-child relationship automatically associates it.

This PR only needs to change CSS positioning, but I had to remove the `<br>` which is no longer considered good practise anyway (versus using CSS for spacing and positioning).

app/components/SortableTableHeader.tsx
kriscooke: Positioning the icon independent of the box model flow was needed in order to maintain its flush-right positioning regardless of label text or viewport width. Flex layout was apt to place it on the following line next to the last text character, among other problems. 

Basically: fluid behavior of the icon is undesirable in this case. Only too happy to accept tweaks if you'd like to take a crack at it yourself though :)

kriscooke:
@matthieu-foucault I've tweaked the checkbox positioning for Safari (which doesn't retain its default vertical positioning on absolute horizontally-positioned elements).

As mentioned in my other comment, input accessibility remains intact (child input is associated with parent label, and Axe passes), and I don't think flex layout can be of particular help in this instance. Always happy to accept tweaks though.

matthieu-foucault (approved):
Happy with the current styles.

PR #1176, opened 2020-10-07
Create Product style update

kriscooke:
In the Create Product form at `/admin/products-benchmarks`, and now in the CIIP application form at `/reporter/application` as well:
- Grey background color for form
- Darker border color
- Form card has more padding
- Close button is grey instead of red
- Card header is blue with white text instead of grey
- "help" text color has been darkened to maintain AA contrast after background color change
[GGIRCS-1629](https://youtrack.button.is/issue/GGIRCS-1629)

Screen Shot 2020-10-06 at 5 31 55 PM

Screen Shot 2020-10-06 at 5 13 33 PM

### DIFF
--- a/app/containers/Products/ProductCreatorContainer.tsx
+++ b/app/containers/Products/ProductCreatorContainer.tsx
@@ -84,7 +84,11 @@ export const ProductCreator: React.FunctionComponent<Props> = ({
               onSubmit={saveProduct}
             >
               <Button type="submit">Add Product</Button>
-              <Button variant="danger" onClick={toggleShowCreateForm}>
+              <Button
+                variant="secondary"
+                style={{marginLeft: '0.5em'}}
+                onClick={toggleShowCreateForm}
+              >
                 Close
               </Button>
             </JsonSchemaForm>
--- a/app/containers/Products/ProductCreatorContainer.tsx
+++ b/app/containers/Products/ProductCreatorContainer.tsx
@@ -59,10 +59,18 @@ export const ProductCreator: React.FunctionComponent<Props> = ({
 
   return (
     <>
-      <div>
-        <Card style={{marginTop: '10px'}}>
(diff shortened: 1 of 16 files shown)


app/containers/Products/ProductCreatorContainer.tsx
```diff
@@ -76,19 +85,20 @@ export const ProductCreator: React.FunctionComponent<Props> = ({
...
           </Card.Body>
         </Card>
       </div>
-      <style jsx>{`
-        .hidden-title label {
-          display: none;
-        }
-        div {
-          margin-bottom: 50px;
+      <style jsx global>{`
```
kriscooke, line 99:
```
      <style jsx global>
```

This was necessary to punch through into the form's help text style to give it a darker color. 

That was required because the darker (grey) form background made the contrast with the help text insufficient for AA contrast.

matthieu-foucault: Given that the CIIP application forms should have similar changes applied to them for consistency, I would suggest changing the class of the help button instead:
https://github.com/bcgov/cas-ciip-portal/blob/8690490de0a5fbde96649d34c57cec28c451fb39/app/containers/Forms/FormFieldTemplate.tsx#L30

kriscooke:
@matthieu-foucault After our brief look at this yesterday:
- I've just updated the selector in the "Save & Continue" button unit test
- I decided not to change the Happo screenshot to encompass the whole page in this PR, because on some further investigation I realized I would need to tweak CSS in `layouts/default-layout` (changing the `html, body` tag `height: 100%;` to `min-height: 100%;`) that will affect every page. This PR is scoped to a couple of forms, so this didn't seem like the right place to do that.

matthieu-foucault (approved):
Tested the application to look at the product and application forms, and it all looks good

PR #1178, opened 2020-10-09
perf: facility_has_swrs_report doesn't need to get all the data

matthieu-foucault:


dleard (approved):
Note: This computed column is actually not needed, it's only used in the ApplyButtonContainer & the necessary info is returned by the search function (column: last_swrs_reporting_year). But since we may end up refactoring / removing the search functions in the future it's probably best to keep it & leave the implementation as is.

matthieu-foucault:
@dleard yes, this was the simpler change, I didn't want to refactor the front-end code

PR #1191, opened 2020-10-13
Feature/hide comments field if empty 1877

pbastia:
- Automatically hides the comments field on the PDF if they are not present
- Empty comments just dispaly as an empty string on the web form summary page


app/containers/Pdf/PdfFieldTemplate.tsx
```diff
@@ -10,27 +10,31 @@ const styles = StyleSheet.create({
 const PdfFieldTemplate: React.FunctionComponent<FieldTemplateProps> = (
   props
 ) => {
+  const getLocalLabel = (props: FieldTemplateProps) => {
+    if (props.label === 'comments') {
+      return null;
+    } // taken care of by the custom field.
+    if (props.label === 'gases' || props.label === 'sourceTypeName')
+      return null;
```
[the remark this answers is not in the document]
matthieu-foucault: Maybe just stick a comment in there to indicate that there's a better way to do this

matthieu-foucault (changes requested):

app/containers/Pdf/PdfFieldTemplate.tsx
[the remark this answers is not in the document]
pbastia: Agreed 100%, it's bad practice to switch on labels that might change in the future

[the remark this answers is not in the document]
pbastia: I added a `disableRenderingIfEmpty: true` field in the uiSchema.
Anything with this flag set will have the template not render the field's template if set to true

app/containers/Forms/SummaryFormFieldTemplate.tsx
```diff
@@ -9,10 +9,17 @@ const FormFieldTemplate: React.FunctionComponent<FieldTemplateProps> = ({
...
+  uiSchema
 }) => {
   if (hidden) return null;
 
+  if (
+    uiSchema['ui:options']?.disableRenderingIfEmpty &&
+    !children[0]?.props?.formData
+  )
+    return null;
+
```
matthieu-foucault, line 22: move this below the two `if` statements, to ensure that the schema is neither an object nor an array. 
To ensure that you render the field if it has a value of `0`, you should use
```suggestion
  if (
    uiSchema['ui:options']?.disableRenderingIfEmpty &&
    (children[0]?.props?.formData === null || children[0]?.props?.formData === undefined)
  )
    return null;

```

app/components/Application/ApplicationDetailsCardItemCustomFields.tsx
```diff
@@ -29,6 +29,9 @@ const customFields = (
   };
 
   const CUSTOM_FIELDS: Record<string, React.FunctionComponent<FieldProps>> = {
+    ProblemReportField: (props) => (
+      <span id={props.idSchema?.$id}>{props.formData ?? null}</span>
+    ),
```
matthieu-foucault, line 34: No sure that you need this now that you're handling the conditional rendering in the `FormFieldTemplate`

app/containers/Pdf/PdfFieldTemplate.tsx
```diff
@@ -7,30 +7,37 @@ const styles = StyleSheet.create({
...
+
+  return <Text>{`\n${props.label}: `}</Text>;
+};
+
 const PdfFieldTemplate: React.FunctionComponent<FieldTemplateProps> = (
   props
 ) => {
+  if (
+    props.uiSchema['ui:options']?.disableRenderingIfEmpty &&
+    !props.children[0]?.props?.formData
```
matthieu-foucault, line 28:
```suggestion
    (props.children[0]?.props?.formData == null || props.children[0]?.props?.formData == undefined)
```

app/containers/Applications/ApplicationDetailsPdf.tsx
```diff
@@ -57,6 +57,7 @@ const styles = StyleSheet.create({
 });
 
 const CUSTOM_FIELDS = {
+  ProblemReportField: (props) => <Text>{props.formData}</Text>,
```
matthieu-foucault, line 60: Same comment as above: is this still needed?

app/containers/Pdf/PdfFuelFieldsTemplate.tsx
```diff
@@ -16,6 +16,13 @@ export const PdfFuelFieldsTemplate: React.FunctionComponent<Props> = ({
       <Text style={{fontSize: 15, letterSpacing: 2}}>
         {'\n'}Fuel:{'\n'}
       </Text>
+      {formData.comments ? (
+        <Text style={{marginBottom: 2, marginTop: 2}}>
+          {`comments: ${formData.comments}`}
+          {'\n'}
+        </Text>
+      ) : null}
+
```
matthieu-foucault, line 25: Is this needed? `PdfFieldTemplate` seems to be handling the conditional rendering

app/containers/Pdf/PdfProductionFieldsTemplate.tsx
```diff
@@ -16,6 +16,13 @@ export const PdfProductionFieldsTemplate: React.FunctionComponent<Props> = ({
       <Text style={{fontSize: 15, letterSpacing: 2}}>
         {'\n'}Production:{'\n'}
       </Text>
+      {formData.comments ? (
+        <Text style={{marginBottom: 2, marginTop: 2}}>
+          {`comments: ${formData.comments}`}
+          {'\n'}
+        </Text>
+      ) : null}
+
```
matthieu-foucault, line 25: Is this needed? `PdfFieldTemplate` seems to be handling the conditional rendering

matthieu-foucault (changes requested):
Works well! Just needs a bit of cleanup as I think some changes are not needed now that you added `disableRenderingIfEmpty`

app/components/Application/ApplicationDetailsCardItemCustomFields.tsx
pbastia: the `?? null` part is redundant, but the `FormFieldTemplate` passes the children to be rendered as-is from what I understand, so we still need a case for that type of field (or it'll get caught by the `StringField` case which is what we wanted to depart from)

app/containers/Forms/SummaryFormFieldTemplate.tsx
pbastia: Good catch! although if the comment has a value of numeric 0, something has gone wrong since it should be a string with value `"0"`. That might be confusing for the user to see "comments: 0" when they haven't entered anything?

matthieu-foucault: My reasoning here would be for other fields that would eventually use that uiSchema option in the future, not necessarily for the comments form

app/components/Application/ApplicationDetailsCardItemCustomFields.tsx
matthieu-foucault: The departure from `StringField` was because that component renders `[no data entered]` when the value is null-ish. Now that there is the uiSchema option, I think that we can use `StringField` again.

app/containers/Forms/SummaryFormFieldTemplate.tsx
pbastia: Excellent point 👍

app/components/Application/ApplicationDetailsCardItemCustomFields.tsx
pbastia: True - and then we get the diff logic already showing up :)

app/containers/Pdf/PdfFuelFieldsTemplate.tsx
pbastia: The issue here is that by using a Custom Field we shortcut all the templating/widget customization that the jsonschema-form library gives us, and the contract is that we are responsible for 100% of the rendering.
So this object won't pass through the template we defined in `PdfFieldTemplate.tsx` from what I'm seeing.
Since this is reused between `PdfFuelFieldsTemplate.tsx` and `PdfProductionFieldsTemplate.tsx` I'd suggest making a custom component to display the comments in this case

app/containers/Applications/ApplicationDetailsPdf.tsx
```diff
@@ -153,30 +153,35 @@ export const ApplicationDetailsPdf: React.FunctionComponent<Props> = (
...
-                tagName={'VIEW' as any}
-                widgets={customWidget}
-                formContext={{query}}
-              >
-                <View />
-              </JsonSchemaForm>
-            </View>
-          ))}
+          {formResults.map(({node}) => {
+            console.log(node);
```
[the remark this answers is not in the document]
pbastia: PDF removed now, resolving

app/containers/Pdf/PdfFieldTemplate.tsx
pbastia: PDF removed, resolving

matthieu-foucault (approved):

PR #1193, opened 2020-10-13
Fix: generated heat/electricity not included in incentive calculation

dleard:
Subtracts generated heat/electricity from product_emissions in the incentive calculation if required.
Adds missing pgTap tests.


matthieu-foucault (dismissed):

matthieu-foucault:
@dleard looks good. Just needs a rebase

matthieu-foucault (approved):

PR #1197, opened 2020-10-16
feat: update keycloak url for dev-test-prod

dleard:
dev: dev.oidc.gov.bc.ca
test: test.oidc.gov.bc.ca
prod: oidc.gov.bc.ca


matthieu-foucault (approved):
Manually tested with all three `NAMESPACE` env variables. Looks good

PR #1215, opened 2020-10-23
PDF Download style tweaks

kriscooke:
@matthieu-foucault I noticed a couple more tweaks I wanted to add to #1214 after it was merged:

- Shows the program title and logo from the site header in print (just not the navigation)
- Opens the PDF download in a new tab
- Hides the certification signature interface in print
  - not interactive on paper
  - if they have signed before printing, it won't be visible in the PDF anyway, which is confusing
- Replaces the non-interactive "expand" link to the full text disclaimer with the full URL of the link destination.

(web view):
Screen Shot 2020-10-23 at 4 18 56 PM

(print view):
Screen Shot 2020-10-23 at 4 18 13 PM

[GGIRCS-2037](https://youtrack.button.is/issue/GGIRCS-2037)

### DIFF
--- a/app/containers/Applications/ApplicationDetailsContainer.tsx
+++ b/app/containers/Applications/ApplicationDetailsContainer.tsx
@@ -163,7 +163,8 @@ export const ApplicationDetailsComponent: React.FunctionComponent<Props> = (
       </a>
       <style jsx global>{`
         @media print {
-          header,
+          header .header-right,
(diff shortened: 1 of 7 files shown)


matthieu-foucault:
@kriscooke  When pending certification, there are some inputs that we may want to hide
image

matthieu-foucault (approved):

PR #1234, opened 2020-10-28
ci(commitlint): add commitlint github action into ci

junminahn:
- lint commit messages (https://github.com/marketplace/actions/commit-linter)


matthieu-foucault:
@junminahn  thanks! can you rebase and force-push your branch again, I'm not sure that the action will run since they were previously disabled

matthieu-foucault (approved):

PR #1258, opened 2020-11-06
feat: work in progress to enable pdf print with weasyprint

matthieu-foucault:


app/containers/Applications/ApplicationDetailsContainer.tsx
```diff
@@ -50,17 +49,17 @@ export const ApplicationDetailsComponent: React.FunctionComponent<Props> = (
...
-  //   const blob = await response.blob();
-  //   FileDownload(blob, 'CIIP_Application.pdf');
-  // };
+  const handleDownloadPdf = async () => {
+    const response = await fetch('/print-pdf', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({html: document.documentElement.outerHTML})
```
matthieu-foucault, line 58: @wenzowski I would assume that this is how you allow untrusted HTML, and how you get hacked?

app/server/routes/print-pdf.js
```diff
@@ -2,18 +2,18 @@ require('isomorphic-fetch');
 const express = require('express');
 const router = express.Router();
 const consola = require('consola');
-const fs = require('fs');
+// const fs = require('fs');
 const WEASYPRINT_HOST = process.env.WEASYPRINT_HOST || 'http://localhost';
 const WEASYPRINT_PORT = process.env.WEASYPRINT_PORT || '5001';
 module.exports = async () => {
   router.post('/', async (req, res) => {
```
matthieu-foucault, line 9: This route should be protected by keycloak

app/containers/Applications/ApplicationDetailsContainer.tsx
matthieu-foucault: Using `renderToStaticMarkup`, which would fix the security issue, might be a bit tricky as the Relay containers have to be rendered within a QueryRenderer

matthieu-foucault (commented):
The current code is intended to be run alongside a weasyprint image.
`docker run -p 5001:5001 bcgovimages/weasyprint` to start weasyprint, and add the following to your `.env` file. 
```
WEASYPRINT_HOST=http://localhost
WEASYPRINT_PORT=5001
```

Weasyprint seems to have issues with the styles. Given that we need to make the html contents trusted anyway (see https://weasyprint.readthedocs.io/en/stable/tutorial.html#security), Puppeteer might be a better solution

wenzowski:
What I was mumbling about this morning was the idea of making a new express endpoint, running [a static nextjs page renderer server-side](https://nextjs.org/docs/basic-features/pages#server-side-rendering), piping fully-formed html output over to a puppeteer microservice (passing api key to do so), piping back the pdf through express untouched to download by client side. This way we are confident that our puppeteer microservice is behind a server-side api and can't be directly pwnd

app/containers/Applications/ApplicationDetailsContainer.tsx
wenzowski: yes this would allow injection of whatever html a user injects into `document` via dev tools (incl malicious scripts)

wenzowski: should be possible to `renderToStaticMarkup` via nextjs server-side dynamic page rendering (though if relay is needed to fetch page contents we'd need to fix the issue with relay not running on server-side-render...should be possible to fix...used to work...)

matthieu-foucault:
Following our latest UX testing feedback, it appears that our users would be fine without any pdf support, as long as they can reliably access their past submitted applications

PR #1274, opened 2020-11-12
fix: replace pushd in bash command with cd

dleard:


dleard:
does not fix issue. Issue was in dockerfile (https://github.com/bcgov/cas-ciip-portal/pull/1275)

PR #1275, opened 2020-11-12
fix: add copy command in dockerfile for mocks schema

dleard:


helm/cas-ciip-portal/templates/cron-deploy-data.yaml
```diff
@@ -71,7 +71,7 @@ spec:
                   set -euo pipefail;
                   data/deploy-data.sh --oc-project=$(NAMESPACE);
                   if {{ .Values.env.enableDbMocks }}; then
-                    pushd "../mocks_schema";
+                    pushd ../mocks_schema;
```
pbastia: This will still fail I think since we copy that folder in the `${HOME}` folder

```suggestion
                    pushd mocks_schema;
```

pbastia (changes requested):

pbastia (approved):

PR #1289, opened 2020-11-18
Allow editing of the last application close date, even if past

kriscooke:
- If there's currently an open application window, then no previous application window dates can be changed
  - prevents overlapping application windows, which the system cannot handle
- If there's **no** open application window, then the most recently closed application close date can be changed
- Adds validation to ensure when editing date ranges a) for the application window, and b) for the reporting years start/end that the proposed dates do not overlap with that of another reporting period.

[GGIRCS-2022](https://youtrack.button.is/issue/GGIRCS-2022)

### DIFF
--- a/app/containers/Admin/ReportingYear/ReportingYearTable.tsx
+++ b/app/containers/Admin/ReportingYear/ReportingYearTable.tsx
@@ -7,7 +7,7 @@ import createReportingYearMutation from 'mutations/reporting_year/createReportin
 import ReportingYearFormDialog from './ReportingYearFormDialog';
 import NewReportingYearFormDialog from './NewReportingYearFormDialog';
 import {nowMoment, defaultMoment, dateTimeFormat} from 'functions/formatDates';
-import {validateExclusiveApplicationWindow} from 'containers/Admin/ReportingYear/reportingYearValidation';
+import {validateExclusiveDateRanges} from 'containers/Admin/ReportingYear/reportingYearValidation';
 
 interface Props {
   relay: RelayProp;
@@ -39,9 +39,9 @@ export const ReportingYearTableComponent: React.FunctionComponent<Props> = (
     return edge.node.reportingYear;
   });
 
-  const applicationWindowValidator = useMemo(() => {
+  const exclusiveDateRangesValidator = useMemo(() => {
     return (year, formData, errors) => {
-      return validateExclusiveApplicationWindow(
+      return validateExclusiveDateRanges(
         year,
         props.query.allReportingYears.edges,
         formData,
@@ -130,7 +130,7 @@ export const ReportingYearTableComponent: React.FunctionComponent<Props> = (
         createReportingYear={createReportingYear}
         clearForm={clearForm}
         existingYearKeys={existingYearKeys}
-        validateExclusiveApplicationWindow={applicationWindowValidator}
+        validateExclusiveDateRanges={exclusiveDateRangesValidator}
       />
       <ReportingYearFormDialog
         show={dialogMode === 'edit'}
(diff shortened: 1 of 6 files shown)


app/containers/Admin/ReportingYear/ReportingYearTable.tsx
```diff
@@ -75,7 +117,7 @@ export const ReportingYearTableComponent: React.FunctionComponent<Props> = (
     <>
       <div style={{textAlign: 'right'}}>
         <Button
-          style={{marginTop: '-220px'}}
+          style={{marginTop: '-330px'}}
```
kriscooke, line 120: @pbastia Unfortunately I realized too late that we'd have to do some gymnastics to try to use your new [`titleControls` button](https://github.com/bcgov/cas-ciip-portal/pull/1290/files#diff-f6385c6dabb76b0f2eb0463c90bf2433c347a773f6167522cf6ebf99ae1e8e74R108) in DefaultLayout with the reporting years admin page - the action needed for the button is known only within the scope of `ReportingYearTable` and isn't available at the top page level... 🤔

app/containers/Admin/ReportingYear/ReportingYearTable.tsx
```diff
@@ -17,10 +18,28 @@ function formatListViewDate(date: string) {
...
+    .slice()
+    .sort((yearA, yearB) => {
+      const yearACloseTime = defaultMoment(yearA.node.applicationCloseTime);
+      const yearBCloseTime = defaultMoment(yearB.node.applicationCloseTime);
+
+      return yearACloseTime.isBefore(yearBCloseTime)
+        ? 1
+        : yearBCloseTime.isBefore(yearACloseTime)
+        ? -1
+        : 0;
```
matthieu-foucault, line 32: double ternary operators (a quinary operator? :upside_down_face: ) are a bit hard to read. Consider if/else or some other notation for it.

app/containers/Admin/ReportingYear/reportingYearValidation.ts
```diff
@@ -109,4 +187,9 @@ function validateUniqueKey(existingKeys, formData, errors) {
   }
 }
 
-export {validateApplicationDates, validateAllDates, validateUniqueKey};
+export {
+  validateApplicationDates,
+  validateAllDates,
+  validateUniqueKey,
+  validateExclusiveDateRanges
```
matthieu-foucault, line 194: I would love to see unit tests for all of these functions :)

matthieu-foucault (changes requested):

app/containers/Admin/ReportingYear/ReportingYearTable.tsx
kriscooke: What I've done in the meantime was to tweak the currently implemented absolute positioning of the New Reporting Year button slightly to center it vertically between the title header and the table - there was a problem with using that method to put it in its original place (making it impossible to click on the button).

We need to find a better solution in the component hierarchy for these buttons that do operations on an inner page component (adding items to the SearchTable).

app/tests/unit/containers/Admin/ReportingYear/reportingYearValidation.test.ts
```diff
@@ -0,0 +1,582 @@
...
+      `${ERRORS.PAST_DATE} Application open time`
+    );
+  });
+  it('Present or future application open date should not add an error', () => {
+    const resultingErrors1 = validateApplicationDates(
+      existingYears,
+      {
+        reportingPeriodStart: '2030-01-01T00:00:00-08:00',
+        reportingPeriodEnd: '2030-12-31T23:59:59-08:00',
+        applicationOpenTime: TODAY,
```
matthieu-foucault, line 244: I suppose this is fine, as long as we look into mocking the current date before the summer of 2031 :) I really wouldn't bother mocking things now anyway if we're planning on migrating away from Moment.js for our dates

matthieu-foucault (approved):
Looks good to me

matthieu-foucault (approved):

PR #1290, opened 2020-11-19
Bug/search and reset buttons alignment

pbastia:
Addressing https://youtrack.button.is/issue/GGIRCS-1919

### DIFF
--- a/app/components/SearchBox.tsx
+++ b/app/components/SearchBox.tsx
@@ -47,10 +47,18 @@ const SearchBox: React.FunctionComponent<Props> = ({
 
   const uiSchema = {
     searchField: {
-      'ui:col-md': 5
+      'ui:col-md': 5,
+      'ui:placeholder': 'Search By:',
+      'ui:options': {
+        label: false
+      }
     },
     searchValue: {
-      'ui:col-md': 7
+      'ui:col-md': 7,
+      'ui:placeholder': 'Search Term',
+      'ui:options': {
+        label: false
+      }
     }
   };
 
@@ -64,8 +72,8 @@ const SearchBox: React.FunctionComponent<Props> = ({
         ObjectFieldTemplate={FormObjectFieldTemplate}
         onSubmit={handleSubmit}
       >
-        <div className="search-actions">
-          <ButtonGroup style={{marginLeft: '20px'}}>
+        <div className="search-actions" style={{textAlign: 'right'}}>
+          <ButtonGroup>
             <Button variant="success" type="submit">
               Search
             </Button>
--- a/app/components/SearchBox.tsx
+++ b/app/components/SearchBox.tsx
@@ -47,18 +47,10 @@ const SearchBox: React.FunctionComponent<Props> = ({
 
   const uiSchema = {
     searchField: {
-      'ui:col-md': 5,
-      'ui:placeholder': 'Search By:',
-      'ui:options': {
-        label: false
-      }
+      'ui:col-md': 5
     },
     searchValue: {
-      'ui:col-md': 7,
-      'ui:placeholder': 'Search Term',
-      'ui:options': {
-        label: false
-      }
+      'ui:col-md': 7
     }
   };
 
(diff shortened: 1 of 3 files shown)


kriscooke (commented):
Looks like a Cypress test needs to be fixed (perhaps a CSS selector was changed and needs to be updated in test?):
```
1) The Create Product modal
       Creates & displays a new product:
     AssertionError: Timed out retrying: Expected to find content: 'New Product' but never did.
```

app/components/SearchBox.tsx
```diff
@@ -47,10 +47,18 @@ const SearchBox: React.FunctionComponent<Props> = ({
 
   const uiSchema = {
     searchField: {
-      'ui:col-md': 5
+      'ui:col-md': 5,
+      'ui:placeholder': 'Search By:',
+      'ui:options': {
+        label: false
+      }
```
kriscooke, line 54: The search inputs' `<label>`s were explicitly added as part of a recent accessibility upgrade. Screenreaders etc. cannot comprehend placeholder text as semantically being the label for that input - only the `<label>`

Regardless of the accessibility context, we're trying to lock down v1.x by not unnecessarily introducing UI changes such as this until v2.

app/components/SearchBox.tsx
```diff
@@ -47,10 +47,18 @@ const SearchBox: React.FunctionComponent<Props> = ({
...
     },
     searchValue: {
-      'ui:col-md': 7
+      'ui:col-md': 7,
+      'ui:placeholder': 'Search Term',
+      'ui:options': {
+        label: false
+      }
```
kriscooke, line 61: Ditto - can you find an alternative to accomplish the goal of alignment without changing the search inputs' labelling UI? (remove placeholder and add label back)

app/pages/admin/products-benchmarks.tsx
```diff
@@ -132,6 +114,11 @@ class ProductsBenchmarks extends Component<Props> {
             <br />
           </Col>
         </Row>
+        <Row>
+          <Col>
+            <Button onClick={this.toggleShowCreateForm}>New Product</Button>
+          </Col>
+        </Row>
```
kriscooke, line 121: See my previous comment about unnecessarily introducing further change to UI post-v1 user launch. We have plenty of opportunity to co-design change through an intentional team process with v2.

Our current admins are used to looking for this button in the top right. (and have given positive feedback in the recent user studies about the ease of finding it in that location :)

If you're looking for the location it 'should' be aligned to, it was right-aligned within the grey panel at the top of the screen which says "Manage Products" - so only the original button's vertical location needed adjusting.

kriscooke (changes requested):

app/components/SearchBox.tsx
```diff
@@ -64,8 +72,8 @@ const SearchBox: React.FunctionComponent<Props> = ({
         ObjectFieldTemplate={FormObjectFieldTemplate}
         onSubmit={handleSubmit}
       >
-        <div className="search-actions">
-          <ButtonGroup style={{marginLeft: '20px'}}>
+        <div className="search-actions" style={{textAlign: 'right'}}>
+          <ButtonGroup>
```
kriscooke, line 76: @pbastia You could add some `marginTop` to `.search-actions` here to bring the vertical alignment into line with the other inputs. 

re: my offline comment in standup, it's best practise to stick with CSS only to adjust appearance instead of adding new, empty HTML DOM elements just for the purpose of spacing (and which won't be used for content).

app/pages/admin/products-benchmarks.tsx
pbastia: This one I made the change with @rajpersram 's advice, since he's had user feedback saying that people expected the button on the left instead, it was an easy change to reduce user confusion. I'm happy to put it anywhere else!

dleard (dismissed):

app/layouts/default-layout.tsx
```diff
@@ -49,10 +51,15 @@ const DefaultLayout: React.FunctionComponent<Props> = ({
...
-              )}
+              <Row>
+                <Col>
+                  <h1>{title}</h1>
+                  {help && (
+                    <Help title={help.title} helpMessage={help.helpMessage} />
+                  )}
+                </Col>
+                <Col xs="auto">{titleControls}</Col>
+              </Row>
```
kriscooke, line 62: Nice one! 👌

kriscooke (approved):

kriscooke:
Also just a note that we extend our commit prefix conventions to branch names as well, so a similar branch in future might be named `fix/search-button-alignment`. It's not terribly important this time around - just something for staying organized. 😄

PR #1306, opened 2020-11-25
"Revise Application" button should be "Resume Latest Draft" if one exists

kriscooke:
In viewing a submitted application (for example from an old email link requesting changes to the application) at `/reporter/view-application`, if a newer draft has already been created, the "Revise" button will instead show:

a) "Resume latest draft" if the latest draft's version number is greater than the latest submitted version
b) "View most recent submission" if they're viewing a really old version, where a new version has already been revised and submitted
  * even if this newer submission has had changes requested, they can first navigate to the newer submission to view analyst comments before clicking either "Revise" or "Resume latest draft" on that version.

There is no change in appearance or behavior for the case where the submitted revision is still the most recent revision.

[GGIRCS-2047](https://youtrack.button.is/issue/GGIRCS-2047)

a) Resume latest draft

b) View most recent submission

### DIFF
--- a/app/containers/Applications/ReviseApplicationButtonContainer.tsx
+++ b/app/containers/Applications/ReviseApplicationButtonContainer.tsx
@@ -15,6 +15,14 @@ export const ReviseApplicationButton: React.FunctionComponent<Props> = ({
   relay
 }) => {
   const router = useRouter();
+  const newerDraftExists =
+    application.latestDraftRevision.versionNumber > application.latestSubmittedRevision.versionNumber;
+  const newerDraftURL = `/reporter/application?applicationId=${
+    encodeURIComponent(application.id)
+  }&version=${
+    application.latestDraftRevision.versionNumber
+  }`;
+
   const handleClick = async () => {
     const variables = {
       input: {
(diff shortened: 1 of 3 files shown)


app/containers/Applications/ReviseApplicationButtonContainer.tsx
```diff
@@ -15,6 +15,42 @@ export const ReviseApplicationButton: React.FunctionComponent<Props> = ({
...
+    encodeURIComponent(application.id)
+  }&version=${application.latestSubmittedRevision.versionNumber}`;
+  const viewLatestSubmissionButton = (
+    <>
+      <p style={{margin: '1rem 0'}}>
+        <strong>Note:</strong> There is a more recently submitted version of this application.
+      </p>
+      <a href={latestSubmissionURL}>
+        <Button>View most recent submission</Button>
+      </a>
```
matthieu-foucault, line 31: You should use a nextjs `Link` here so that the navigation is client-side

kriscooke: Ah - thanks for the reminder!

matthieu-foucault (approved):

PR #1307, opened 2020-11-25
feat: Protect reporting_year date ranges with trigger

dleard:
- Throws an exception if a new or updated date range overlaps with other entries in the table

### DIFF
--- a/schema/deploy/trigger_functions/protect_date_overlap.sql
+++ b/schema/deploy/trigger_functions/protect_date_overlap.sql
@@ -1,0 +1,31 @@
+-- Deploy ggircs-portal:trigger_functions/protect_date_overlap to pg
+-- requires: schema_ggircs_portal_private
+
+begin;
+  create or replace function ggircs_portal_private.protect_date_overlap()
+    returns trigger as $$
+  declare
+    temp_row record;
+
+  begin
+
+    for temp_row in
+      select * from ggircs_portal.reporting_year
+    loop
+      if (new.reporting_year != temp_row.year) then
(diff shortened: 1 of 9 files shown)


schema/deploy/trigger_functions/protect_date_overlap.sql
```diff
@@ -0,0 +1,36 @@
...
+    loop
+      if (new.reporting_year != temp_row.reporting_year) then
+        if ((new.reporting_period_start between temp_row.reporting_period_start and temp_row.reporting_period_end)
+        or (temp_row.reporting_period_start between new.reporting_period_start and new.reporting_period_end)
+        or (new.reporting_period_end between temp_row.reporting_period_start and temp_row.reporting_period_end)
+        or (temp_row.reporting_period_end between new.reporting_period_start and new.reporting_period_end)
+        or (new.application_open_time between temp_row.application_open_time and temp_row.application_close_time)
+        or (temp_row.application_open_time between new.application_open_time and new.application_close_time)
+        or (new.application_close_time between temp_row.application_open_time and temp_row.application_close_time)
+        or (temp_row.application_close_time between new.application_open_time and new.application_close_time))
```
pbastia, line 23: For readability, we could use the `OVERLAPS` psql operator instead, in 9.6+
https://www.postgresql.org/docs/11/functions-datetime.html

dleard: Awesome Pierre, thanks! I was not aware of that operator. I'll re-write the function using `overlaps`.

pbastia (approved):

PR #1308, opened 2020-11-27
fix: reporting period dropdown alignment on bootstrap md width

pbastia:


kriscooke (approved):
Thanks for the extra test fixes! Just a note for the future - if you find unrelated things you'd like to fix along the way, it would be helpful if you could put them in a separate pull request from the task you're working on (in this case, [GGIRCS-1919](https://youtrack.button.is/issue/GGIRCS-1919)).

That way if there's any issue in review with the unrelated fixes, they won't delay the timely merge of the related work (for inclusion in a priority deployment, for example).

PR #1328, opened 2020-12-04
Refactor/remove calls to old mock timestamp method

pbastia:


schema/data/dev/certification_url.sql
```diff
@@ -23,11 +26,16 @@ update ggircs_portal.certification_url set certification_signature = 'signed' wh
...
 -- Update the status of application with id=1 to be 'submitted'
-  insert into ggircs_portal.application_revision_status(application_id, version_number, application_revision_status)
-    values (1,1,'submitted');
-  insert into ggircs_portal.application_revision_status(application_id, version_number, application_revision_status)
-  values (1,2,'submitted');
+insert into ggircs_portal.application_revision_status(application_id, version_number, application_revision_status)
+  values (1,1,'submitted');
+
+-- Moving timestamp one second forward to avoid revisions being submitted at the same time
+select mocks.set_mocked_time_in_transaction('2020-07-03 00:00:00.000000-07'::timestamptz + interval '1 second');
```
pbastia, line 35: @dleard  this is it!

schema/data/deploy-data.sh
```diff
@@ -205,6 +217,9 @@ createdb
 if [[ " ${actions[*]} " =~ " deploySwrs " ]]; then
   deploySwrs
 fi
+if [[ " ${actions[*]} " =~ " deployMocks " ]]; then
+  deployMocks
+fi
```
matthieu-foucault, line 222: This does not appear to be used. `deployMocks` should be added to the `actions` array in the `while` loop above.
This script's documentation should be updated to explain when the mocks schema is deployed, or better, make it an explicit option (`--deploy-mocks-schema`).

schema/data/dev/application-2018.sql
```diff
@@ -1,12 +1,7 @@
...
 
--- Set the current reporting year to 2018
-create or replace function ggircs_portal.current_timestamp() returns timestamptz as
-$$
-  select application_open_time
-  from ggircs_portal.reporting_year
-  where reporting_year = 2018
-$$ language sql;
+  -- Set the current reporting year to 2018, application open
+  select mocks.set_mocked_time_in_transaction('2019-04-01 14:49:54.191757-07'::timestamptz);
```
matthieu-foucault, line 4: any reason why the time is hardcoded here instead of 
```suggestion
  select mocks.set_mocked_time_in_transaction(
    select application_open_time
      from ggircs_portal.reporting_year
      where reporting_year = 2018
  );
```

schema/data/dev/certification_url.sql
matthieu-foucault, line 35: how about
```suggestion
select mocks.set_mocked_time_in_transaction(now() + interval '1 second');
```

schema/deploy/function_opened_reporting_year.sql
```diff
@@ -7,8 +7,7 @@ returns ggircs_portal.reporting_year
 as $function$
   select *
   from ggircs_portal.reporting_year as _reporting_year
-  where ggircs_portal.current_timestamp() >= application_open_time
-  and ggircs_portal.current_timestamp() <= application_close_time;
+  where now() between application_open_time and application_close_time;
```
matthieu-foucault, line 10: https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_BETWEEN_.28especially_with_timestamps.29

Usage of `between` is discouraged. This behaves as you would expect here since we're comparing timestamps with timestamps, but it could lead to bugs if we were to change `application_close_time` to `application_close_date`

matthieu-foucault (changes requested):

schema/deploy/function_opened_reporting_year.sql
pbastia: Oh, I had no idea. Thanks!

schema/data/deploy-data.sh
pbastia: the dev data now relies on the mocks schema being there, but the mocks schema needs to grant access to roles that are deployed with the portal schema I believe

schema/data/dev/application-2018.sql
pbastia: Good point, I pasted this similarly to our db unit tests

schema/data/dev/certification_url.sql
pbastia: Some tests rely on this data, so having a dependency on `now()` is not the best. We can have it like the previous comment, where we fetch the value from the `reporting_year` table

matthieu-foucault: Sounds good

schema/data/deploy-data.sh
matthieu-foucault: I'm thinking that `deployMocks` should be called for both the dev and test data. Is the "test" data even used for our tests, or are we still relying on the dev data being deployed for our tests?

you are calling `deployMocks` from within the `deployDevData` function above, so these lines are not really needed if we're not adding a `--deploy-mocks-schema` option

pbastia: I made it a little more flexible: dev and test data need the mocks schema to be deployed, so the script enforces it, but it's also accessible as a separate target if needed

matthieu-foucault (approved):

PR #1386, opened 2021-01-05
Chore/release

matthieu-foucault:


matthieu-foucault:
Temporary removing `sqitch-files-before-tag-are-immutable` check requirement, as I believe it's failing because I was too eager to push master (the other sqitch-related check wasn't required for `master`)

pbastia (approved):

PR #1388, opened 2021-01-06
docs: add rjsf documentation to stack.md

dleard:


dleard:
@matthieu-foucault did I miss anything?
@pbastia is this helpful / clear?

docs/stack.md
```diff
@@ -171,7 +171,26 @@ Some of the SQL search functions require a new type to be created in order for `
...
+Example: app/containers/Forms/FuelRowIdField.tsx
+
+Overriding a field is done the same way as a template, by creating a new component with the appropriate rjsf props (in the example above we extend FieldProps). In the example we have customized the behaviour of the RowId field. This field is a numeric ID that corresponds to an enumerated list of fuel names. Our custom field makes use of [useMemo()](https://reactjs.org/docs/hooks-reference.html#usememo) to only render values when necessary and matches the ID with the corresponding name, rendering the name of the fuel in the list rather than the numerical ID.
+
+A custom field is applied in the same way as the template, defining it in the props where the JsonSchemaForm is instantiated. This FuelRowId example is one of several custom fields we apply to this form, so we have wrapped all the customized fields in one component (see app/components/Application/ApplicationDetailsCardItemCustomFields.tsx) and passed that wrapper to be defined in the props (see app/containers/Applications/ApplicationDetailsCardItem for where this wrapper is applied).
```
pbastia, line 193: I think one of the confusing things I had a hard time grasping was how many times and in which order these custom fields/template get rendered when they pass through the different templates, depending on the data type. Something like Object->Array->CustomTemplate->CustomField->etc
Is there a way we can clarify this? or does it depend on the ad-hoc implementation every time?

dleard: Yes that is how things render from my understanding. If we look here: https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/fields/ArrayField.js#L22

The ArrayField takes TitleField as an argument, if we've customized TitleField, then everything in that custom Field will take precedence.

matthieu-foucault (dismissed):
LGTM, just needs a rebase

matthieu-foucault (approved):

PR #1389, opened 2021-01-07
feat: add data-migrate helm chart

matthieu-foucault:


helm/cas-ciip-portal-data-migrate/templates/job.yaml
```diff
@@ -0,0 +1,156 @@
...
+              rm -rf backup_label.old
+              cd ..
+              pg_ctl -D prod-bak start -w -o "-c log_statement='all' -c logging_collector=on -c archive_mode='off' -c archive_command='' -c ssl='off' -c hba_file='/home/postgres/prod-bak/pg_hba.conf' -c ident_file='/home/postgres/prod-bak/pg_ident.conf'"
+
+              echo "$PORTAL_HOST:$PORTAL_PORT:$PORTAL_DATABASE:$PORTAL_USER:$PORTAL_PASSWORD" >> .pgpass
+              chmod 0600 .pgpass
+              # we need to recreate users first as the ACL created in sqitch need to be dumped to DB
+              # no need to grant any privileges on schemas or set passwords here, this will be done in the app-users job
+              PGPASSWORD="$PORTAL_PASSWORD" PGHOST=$PORTAL_HOST PGUSER=$PORTAL_USER psql<<EOF
+                create user $PORTAL_APP_USER;
```
dleard, line 109: we have a tech debt card to remove this user as it is obsolete. Not sure if it should be removed here, or if for this migration job it doesn't matter.

dleard (approved):
One non-blocking comment on PORTAL_APP_USER above

helm/cas-ciip-portal-data-migrate/templates/job.yaml
matthieu-foucault: I don't think that this matters, unless we take care of the tech debt card before doing this migration

PR #1390, opened 2021-01-08
Feature/cypress uses mock login

pbastia:
Feature to bypass the login/hit to keycloak when running the e2e tests.
This reuses the same codepath as our `AS_ADMIN`, `AS_REPORTER`, ... targets and allows cypress to set a cookie to tell the server which user to take.

`app/server/postgraphile/authenticationPgSettings.js` checks for the AS_CYPRESS process argument, and a `mocks.auth` cookie to allow login to be bypassed as a certain test user.


app/cypress/integration/accessibility/admin-all-pages-accessibility.example.spec.js
```diff
@@ -4,10 +4,7 @@ describe('When logged in as an admin', () => {
...
     // The admin spec uses the same setup as the analyst
     cy.sqlFixture('fixtures/analyst-all-access-setup');
-    cy.login(
-      Cypress.env('TEST_ADMIN_USERNAME'),
-      Cypress.env('TEST_ADMIN_PASSWORD')
-    );
+    cy.mockLogin('admin');
   });
 
   afterEach(() => {
```
matthieu-foucault, line 10: The `logout` below should be removed?

app/server/postgraphile/authenticationPgSettings.js
```diff
@@ -17,11 +23,11 @@ const authenticationPgSettings = (req) => {
       'jwt.claims.sub': '00000000-0000-0000-0000-000000000000',
       'jwt.claims.user_groups': groups.join(','),
       'jwt.claims.priority_group': priorityGroup,
-      role: NO_AUTH_POSTGRES_ROLE
+      role: 'ciip_guest'
```
matthieu-foucault, line 26: Why this change? The NO_AUTH flag is intended for developers, and should be using whatever postgres superuser they have in their environment

app/cypress/integration/page-access.spec.js
```diff
@@ -2,10 +2,7 @@ describe('When logged in as a reporter', () => {
   beforeEach(() => {
     cy.cleanSchema();
     cy.sqlFixture('dev/user');
-    cy.login(
-      Cypress.env('TEST_REPORTER_USERNAME'),
-      Cypress.env('TEST_REPORTER_PASSWORD')
-    );
+    cy.mockLogin('reporter');
```
matthieu-foucault, line 5: This file can be the one testing the real login

app/server/postgraphile/authenticationPgSettings.js
pbastia: Oooh I completely misunderstood what this did, I missed it when I refactored this method out of index.js. Fix incoming

matthieu-foucault (approved):
:heart_eyes: 
Before: 
> All specs passed!                        28:24

After: 
> All specs passed!                    14:47

kriscooke:
✨ 🌈

PR #1401, opened 2021-01-13
fix: cypress e2e race conditions

pbastia:
PR to fix the race condition when using mock logins


matthieu-foucault (approved):
Still a bit confused by some graphql errors we're seeing in the tests logs, but they don't seem to affect the tests, so let's merge this

PR #1429, opened 2021-01-26
Simplified admin dash

kriscooke:
- Re-organizes and simplifies groupings and language in the admin dash according to [GGIRCS-2185](https://youtrack.button.is/issue/GGIRCS-2185)
  - includes removing the Devops section
- Adds a Report a Problem link to the dash - [GGIRCS-2187](https://youtrack.button.is/issue/GGIRCS-2187)
- Enables the accessibility tests for the analyst and admin dashboards
  - tests for other pages in the admin/analyst scopes will be incrementally enabled with related work
- Updates related documentation (screenshots and descriptions of the dash) [here](https://github.com/bcgov/cas-ciip-portal/blob/0ddf8a6168a0fa0083216375dc98207152eeab4d/docs/admin-analyst-guide.md) and [here](https://github.com/bcgov/cas-ciip-portal/pull/1429/files#diff-ca4cb71eb5e8645017b057411c14de69d79af49d9cf89f6a742cab712b0e058a)
[GGIRCS-2186](https://youtrack.button.is/issue/GGIRCS-2186)
- Adds new environment variable `SUPPORT_EMAIL`; will be used temporarily for problem reports until an alternate service (fider replacement) is deployed.

dashboard


app/components/Dashboard/programAdministration.tsx
```diff
@@ -31,21 +31,26 @@ const ProgramDataManagement: React.FunctionComponent<Props> = ({
...
+        <Card.Title>Program Administration</Card.Title>
         <Card.Text>
-          Update program parameters, view program insights etc.
+          Review applications and administer program details:
         </Card.Text>
       </Card.Body>
       <ListGroup variant="flush">
-        {EditableItems}
+        <ListGroup.Item>
+          <Link href="/analyst/applications">
```
matthieu-foucault, line 41: I think this is one of these cases where you want to use the `passHref` prop: https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

app/components/Dashboard/reportingOperations.tsx
```diff
@@ -6,30 +6,23 @@ const ApplicationManagement: React.FunctionComponent = () => {
...
       <ListGroup variant="flush">
-        <ListGroup.Item>
-          <Link href="/analyst/applications">
-            <Card.Link href="#">View all applications</Card.Link>
-          </Link>
-        </ListGroup.Item>
         <ListGroup.Item>
           <Link href="/analyst/organisation-requests">
-            <Card.Link href="#">Organisation access requests</Card.Link>
+            <Card.Link href="#">Access Requests</Card.Link>
```
matthieu-foucault, line 15: Ditto regarding `passHref` (and same below)

app/pages/admin/index.tsx
```diff
@@ -32,14 +33,15 @@ class Admin extends Component<Props> {
...
       <DefaultLayout session={session} title="Administrator Dashboard">
         <div>
           <Row>
+            <ProgramAdministration />
+            <ReportingOperations />
             <UserManagement />
-            <ApplicationManagement />
-            <ProgramDataManagement />
-            <DevOpsManagement />
+            <ReportAProblem serviceUrl={feedbackUrl || '#'} />
```
matthieu-foucault, line 44: Eventually, we should either:
- not render `ReportAProblem` if there is no `feedbackUrl`
- require that `feedbackUrl` is defined

app/components/Dashboard/reportAProblem.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import {Card} from 'react-bootstrap';
+
+interface Props {
+  serviceUrl?: string;
```
matthieu-foucault, line 5:
```suggestion
  serviceUrl: string;
```

app/components/Dashboard/programAdministration.tsx
```diff
@@ -31,21 +31,26 @@ const ProgramDataManagement: React.FunctionComponent<Props> = ({
...
+            <Card.Link href="#">Submitted applications</Card.Link>
+          </Link>
+        </ListGroup.Item>
         <ListGroup.Item>
           <Card.Link
             target="_blank"
             href="https://cas-metabase.pathfinder.gov.bc.ca/"
```
matthieu-foucault, line 48: This link will need updating, and probably should be in an environment variable, if we want it to point to the dev/test/prod metabase instance.

matthieu-foucault (changes requested):

app/components/Dashboard/programAdministration.tsx
kriscooke: Interesting, I never noticed that before -

> Without this, the <a> tag will not have the href attribute, which might hurt your site’s SEO.

...not to mention the ability to right click and open the correct link in a new tab. Will fix.

app/pages/admin/index.tsx
kriscooke: I originally implemented this according to your first point, but immediately realized that makes the feature impossible to see and test until it's in production. We generally want things to be visible in dev and test if they are visible in prod, so we can verify.

I could move the `|| '#'` part up to the `feedbackUrl` assignment, but that seems basically equivalent.

matthieu-foucault: This is okay as-is. There's a code smell here given that in prod, if we forget to set the environment variable, then the application will deploy without any issue, but this link won't work. Ironically this is the link that should be used to report things that don't work

kriscooke: Yes. I guess the question is, how do you require in the yamls that a given value be defined?

matthieu-foucault: https://helm.sh/docs/howto/charts_tips_and_tricks/#know-your-template-functions

https://github.com/bcgov/cas-ciip-portal/blob/9b5ac0a772f67a48c6b30c1da94d513a29abd9ab/helm/cas-ciip-portal/templates/deploy.yaml#L63 would become something like

```yaml
value: {{ required "A valid .Values.feedbackServiceUrl entry is required!" .Values.helpDeskServiceUrl }}
```

kriscooke: Right - for some reason I got confused for a minute and thought we were talking about the Metabase URL. We can't make the fider URL required, because it only exists for production.

Therefore the existing solution correctly allows us to review the feature in dev/test using an `href="#"`, and fills in the correct URL for production.

matthieu-foucault: Well, it doesn't even exist for production anymore, given that Fider was not migrated to the silver cluster...
This would be replaced by the shared inbox email I suppose

app/next.config.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
   publicRuntimeConfig: {
     NO_MATHJAX: process.env.NO_MATHJAX,
     FEEDBACK_SITE_URL: process.env.FEEDBACK_SITE_URL,
+    SUPPORT_EMAIL: process.env.SUPPORT_EMAIL,
```
matthieu-foucault, line 26: Should `FEEDBACK_SITE_URL` be deleted (and all of its occurences replaced by `SUPPORT_EMAIL`)?

kriscooke: I had wondered that but left it as-is, since we intend to replace fider with a new solution that is distinct from a support email.

kriscooke: Perhaps the best compromise in the meantime (until we have implemented that solution) would be to keep the FEEDBACK_SITE_URL variable, but temporarily replace all usages in the app with SUPPORT_EMAIL.

matthieu-foucault (approved):

PR #1433, opened 2021-01-29
Docs/osx postgres install

pbastia:
Update to the docs to include more OSX postgres troubleshooting


kriscooke (approved):
I myself didn't need PG_SYSROOT, and don't have SIP as I'm not on a government computer. Therefore I think we should take your word for it in case this can help the next person :)

PR #1437, opened 2021-01-29
Refactor: remove portal_app user

dleard:
obsolete portal_app role in postgres removed via sqitch rework


matthieu-foucault (approved):

matthieu-foucault (commented):
So, I think that this might be the wrong refactor. We want a user to be created and granted the roles via our sqitch migrations.

dleard:
> So, I think that this might be the wrong refactor. We want a user to be created and granted the roles via our sqitch migrations.

I'm a little confused, I thought this user was obsolete and just needed to be removed. Should we be creating the `ciip_portal` role via sqitch instead?
https://youtrack.button.is/issue/GGIRCS-1787

matthieu-foucault:
Sorry, that card definitely needed more details. A user needs to be created and granted the `ciip_*` roles in order for the app to work, and `sqitch` is the best way to do it in a reusable way (the role needs to be created in the devs local environments, and in the e2e test environment).

When deploying the helm chart, to create the db with the correct permissions and to set a password for the user, the `db_init` job currently create the `ciip_portal` user, which is then used to deploy the sqitch schema, and used by the app to connect to the database.

For some reason, we ended up naming the role `portal_app` in the sqitch plan, but `ciip_portal` in the helm charts.

Updating the sqitch plan to create the `ciip_portal` user instead of `portal_app` makes sense, as the alternative (updating the helm charts) would require more work (an potentially trigger some short downtime during a deployment)

schema/deploy/database_functions/create_portal_app_user.sql
```diff
@@ -3,20 +3,6 @@
...
-    where  rolname = 'portal_app') then
-
-    create user portal_app;
-  end if;
-
-  grant ciip_administrator, ciip_analyst, ciip_industry_user, ciip_guest to portal_app;
-  execute format('grant create, connect on database %I to portal_app', current_database());
-end;
-$$;
+alter role portal_app rename to ciip_portal;
```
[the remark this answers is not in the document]
dleard: good point, I'll change it to do that instead

schema/verify/database_functions/create_portal_app_user.sql
```diff
@@ -2,6 +2,6 @@
 
 begin;
 
-select true from pg_roles where rolname='portal_app';
+select false from pg_roles where rolname='portal_app';
```
matthieu-foucault, line 5:
```suggestion
select false from pg_roles where rolname='portal_app';
select true from pg_roles where rolname='ciip_portal'; 
```

matthieu-foucault (changes requested):

schema/verify/database_functions/create_portal_app_user.sql
matthieu-foucault: ... although this might make sqitch ignore the result of the first statement now ...

dleard: indeed it does

matthieu-foucault (approved):

PR #1464, opened 2021-02-08
feat: replace search_applications with @filterable computed_column

dleard:
- Add postgraphile plugin for fuzzy filtering
- Create computed columns to allow replacing search_applications function with @sortable @filterable

Todo: explore what performance/security settings should be set on the postgraphile plugin so that it can do what we need to and protect against inadvertent/malicious expensive queries https://github.com/graphile-contrib/postgraphile-plugin-connection-filter


schema/deploy/computed_columns/application_operator.sql
```diff
@@ -0,0 +1,22 @@
+-- Deploy ggircs-portal:computed_columns/application_operator to pg
+-- requires: tables/application
+
+begin;
+
+create or replace function ggircs_portal.application_operator(app ggircs_portal.application)
+returns setof ggircs_portal.organisation
```
matthieu-foucault, line 7: Shouldn't this return a single organisation?

schema/deploy/computed_columns/application_operator.sql
```diff
@@ -0,0 +1,22 @@
...
+  begin
+
+    return query (
+      select o.* from ggircs_portal.organisation o
+        join ggircs_portal.facility f on o.id = f.organisation_id
+        join ggircs_portal.application on f.id = app.facility_id
+    );
+
+  end;
+$$ language 'plpgsql' stable;
```
matthieu-foucault, line 18: can it be an `sql` function instead of `plpgslq`? AFAIK they're more efficient as postgres treats plpgsql functions as a black box

app/server/postgraphile/postgraphileOptions.js
```diff
@@ -1,14 +1,15 @@
...
+const ConnectionFilterPlugin = require('postgraphile-plugin-connection-filter');
 
 const postgraphileOptions = () => {
   // Use consola for logging instead of default logger
   const pluginHook = makePluginHook([PostgraphileLogConsola]);
 
   let postgraphileOptions = {
     pluginHook,
-    appendPlugins: [PgManyToManyPlugin],
+    appendPlugins: [PgManyToManyPlugin, ConnectionFilterPlugin],
```
matthieu-foucault, line 12: can those go in the `.postgraphilerc.js` file? This would avoid having to maintain the same list in the `package.json`

schema/deploy/computed_columns/application_operator.sql
```diff
@@ -0,0 +1,22 @@
...
+
+begin;
+
+create or replace function ggircs_portal.application_operator(app ggircs_portal.application)
+returns setof ggircs_portal.organisation
+as $$
+  begin
+
+    return query (
+      select o.* from ggircs_portal.organisation o
```
pbastia, line 12: nice emoji 😁

schema/deploy/computed_columns/application_operator.sql
```diff
@@ -0,0 +1,22 @@
...
+        join ggircs_portal.facility f on o.id = f.organisation_id
+        join ggircs_portal.application on f.id = app.facility_id
```
pbastia, line 14: I'd specify `inner` for readability but it's the default behaviour

schema/deploy/computed_columns/application_operator.sql
dleard: `return query` requires plpgsql

schema/deploy/computed_columns/application_operator.sql
dleard: hm, I may not need application_operator or application_facility at all now that I think about it, maybe just the scalar-returning function is enough

matthieu-foucault: Indeed

dleard: Yeah, they're unnecessary, the logic can go directly in the scalar-returning function. I'll remove them.

PR #1466, opened 2021-02-08
feat: default layout expands to the bottom of the viewport

matthieu-foucault:


app/layouts/default-layout.tsx
```diff
@@ -69,6 +69,18 @@ const DefaultLayout: React.FunctionComponent<Props> = ({
...
+          .page-wrap {
+            display: flex;
+            flex-direction: column;
+            min-height: 100vh;
+          }
+          main {
+            flex-grow: 1;
+          }
+        `}
+      </style>
```
kriscooke, line 83: 👌 Nice move - any particular reason for separating these styles out from the rest?

matthieu-foucault: They don't need to be global. There's probably more below that could be moved to this local style, but I left that for a future refactor.

kriscooke: I thought so - just checking :)

kriscooke (approved):

PR #1472, opened 2021-02-12
Feat/search table multiple headers

pbastia:
Intermediate PR:
- refactors `_app.tsx` to paste router query params straight to relay
- refactors the `SearchTable` component to use query params when triggering the search function
- updates all search tables to not use react effects and relay refetch containers
- wires the Application search table to use the new computed columns for searching

@dleard  @matthieu-foucault


app/components/SearchTableLayout.tsx
```diff
@@ -52,7 +50,7 @@ export const SearchTableLayoutComponent: React.FunctionComponent<Props> = (
             />
           </Col>
         </Row>
-      </Container>
+      </Container> */}
```
dleard, line 53: Is this code to be removed?

pbastia: Yes 👍

PR #1473, opened 2021-02-12
Test: pgTap tests for application search

dleard:
Adds a pgTap test for each new computed column


dleard:
closed in favor of #1500

PR #1486, opened 2021-02-18
feat: filterable search UI components

dleard:


dleard (commented):
Looks pretty good to me @pbastia. Just a couple places where we could explicitly spell out some of the more abstract things going on.

pbastia (approved):

PR #1488, opened 2021-02-19
Fixed help button for admins/analysts

kriscooke:
[GGIRCS-2188](https://youtrack.button.is/issue/GGIRCS-2188)
- Help button is present on all logged-in views for analysts and admins
- "Report a problem to the development team" temporarily opens a mailto link (new email with subject line "Internal Support Request") directed to `SUPPORT_EMAIL`. This URL should be updated when/if we have a new issue reporting system.
- Documentation links to Github admin/analyst guide on master. It's not amazingly thorough, but we can update either the link destination or improve the docs as appropriate.

help button

### DIFF
--- a/app/layouts/default-layout.tsx
+++ b/app/layouts/default-layout.tsx
@@ -9,6 +9,8 @@ import Subheader from 'components/Layout/Subheader';
 import Help from 'components/helpers/Help';
 import SiteNoticeBanner from 'components/Layout/SiteNoticeBanner';
 import CookieDayPickerInput from 'components/helpers/CookieDayPickerInput';
+import HelpButton from 'components/helpers/HelpButton';
+import {ADMIN_GROUP, INCENTIVE_ANALYST} from 'data/group-constants';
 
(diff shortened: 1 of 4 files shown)


app/layouts/default-layout.tsx
```diff
@@ -33,6 +34,10 @@ const DefaultLayout: React.FunctionComponent<Props> = ({
   width = 'narrow',
   help
 }) => {
+  const isInternalUser =
+    session?.userGroups.includes('Incentive Administrator') ||
+    session?.userGroups.includes('Incentive Analyst');
```
matthieu-foucault, line 39:
```ts
import {INCENTIVE_ANALYST, ADMIN_GROUP} from 'data/group-constants';

const internalUserGroups = [INCENTIVE_ANALYST, ...ADMIN_GROUP];
```

You can retrieve the internal user groups like this (you're missing the "Realm Administrator" group), and loop through one array to see if there is an intersection between both groups.

matthieu-foucault (changes requested):
Just a small change, looks good otherwise

dleard (approved):
:+1:

PR #1508, opened 2021-03-02
Feat: relay-search pagination

dleard:
Start (no back)
noBack

Middle (back + forward)
noForward

End (no forward)
backForward


dleard:
Pagination bar at beginning of results (>9 pages, ellipses at end of bar indicates more results from end):
paginationStart

Pagination bar in middle of results (>9 pages, ellipses at beginning and end of bar indicates more results at each end)
paginationMiddle

Pagination bar at end of results (>9 pages, ellipses at beginning of bar indicates more results from start)
paginationEnd

dleard:
Sonarcloud is failing because the new file FilterableTablePagination is similar to PaginationBar. We will need both files until we can remove PaginationBar.

app/components/FilterableComponents/FilterableTablePagination.tsx
```diff
@@ -0,0 +1,93 @@
...
+        active={pageNumber === activePage}
+        onClick={() => paginate(pageNumber)}
+      >
+        {pageNumber}
+      </Pagination.Item>
+    );
+  }
+
+  return (
+    <>
```
matthieu-foucault, line 75: Don't need this fragment

app/pages/analyst/applications.tsx
```diff
@@ -50,22 +50,26 @@ class Applications extends Component<Props> {
     return {
       variables: {
         orderByField: 'operator_name',
-        direction: 'ASC'
+        direction: 'ASC',
+        max_results: 10,
```
matthieu-foucault, line 54: I think that we can set this to 20

app/pages/analyst/applications.tsx
```diff
@@ -50,22 +50,26 @@ class Applications extends Component<Props> {
...
     };
   }
 
   render() {
     const {query} = this.props;
+    /** maxResultsPerPage matches the value of max_results in getInitialProps().
+     *  max_results is passed to the initial query via the router
+     *  maxResultsPerPage is passed as a prop to aid in rendering the correct number of pages in the FilterableTablePagination component
+     **/
+    const maxResultsPerPage = 10;
```
matthieu-foucault, line 66: You could just call the static function above

app/components/FilterableComponents/FilterableTablePagination.tsx
```diff
@@ -0,0 +1,93 @@
...
+  maxResultsPerPage: number;
+  pageInfo: {
+    hasNextPage: boolean;
+    hasPreviousPage: boolean;
+  };
+}
+export const FilterableTablePaginationComponent: React.FunctionComponent<Props> = (
+  props
+) => {
+  const {hasNextPage, hasPreviousPage} = props.pageInfo;
```
matthieu-foucault, line 16: do you need those, given that you already have the number of pages, and the active page?

app/containers/Applications/ApplicationListContainer.tsx
```diff
@@ -76,6 +92,11 @@ export default createFragmentContainer(ApplicationList, {
             ...ApplicationRowItemContainer_application
           }
         }
+        pageInfo {
+          hasNextPage
+          hasPreviousPage
+        }
```
matthieu-foucault, line 98: as per the comment above, this is probably not needed

matthieu-foucault (changes requested):

app/pages/analyst/applications.tsx
dleard: I was thinking we could make this something the user could control, like you see sometimes see something like: `results **10** 20 50` but maybe that's an enhancement for later?

pbastia: Haha, every time I see one of these I set it to the maximum so I don't have to deal with too much pagination - at the expense of server work. I'd say if we leave it at 10 or 20 it's fine for our user's purposes?

app/pages/analyst/applications.tsx
dleard: call it in the FilterableTablePagination component?
Something like this?
`const [max, setMax] = useState(null);

  const getMaxResultsPerPage = async () => {
    const initProps = await Applications.getInitialProps();
    setMax(initProps.variables.max_results);
  }

  useEffect(() => {
    if (!max)
      getMaxResultsPerPage();
  }, []);`

dleard: ugh, 'insert code' option is ugly. Here's a screenshot of what I added to FilterableTablePagination to get the value out of the static async getInitialProps()
getmax

dleard: actually I guess it should go in ApplicationListContainer since this is a re-usable component, but same idea

matthieu-foucault: it probably doesn't have to be an async function. It's only async because it's a requirement of `getInitialProps`

dleard: without `await` it returns a promise rather than the object, which is why I wrapped it in an `async`

app/pages/_app.tsx
```diff
@@ -59,6 +59,10 @@ export default class App extends NextApp<AppProps> {
       ? JSON.parse(String(router.query.relayVars))
       : {};
 
+    const pageVars = router.query.pageVars
+      ? JSON.parse(String(router.query.pageVars))
+      : {};
+
```
matthieu-foucault, line 65: we don't need an extra `pageVars`, the pagination variables could be in `relayVars`. This would allow us to make the fact that `offset` and `max_results` are reserved variable names more explicit (pageVars overrides relayVars below)

matthieu-foucault (changes requested):

app/pages/_app.tsx
matthieu-foucault: Nevermind, relayVars are reset to an empty object when clearing search filters

app/components/FilterableComponents/FilterableTablePagination.tsx
```diff
@@ -0,0 +1,90 @@
...
+            activePage !== maxPages ? paginate(activePage + 1) : null
+          }
+        />
+        <Pagination.Last onClick={() => paginate(maxPages)} />
+      </Pagination>
+    );
+  return null;
+};
+
+export default FilterableTablePaginationComponent;
```
pbastia, line 90: this is exported twice?

app/components/FilterableComponents/FilterableTableHeaders.tsx
```diff
@@ -11,7 +11,9 @@ const FilterableTableHeaders: React.FunctionComponent<ISearchProps> = (
 ) => {
   const router = useRouter();
 
-  const [searchFilters, setSearchFilters] = useState({});
+  const [searchFilters, setSearchFilters] = useState(() =>
+    router.query.relayVars ? JSON.parse(String(router.query.relayVars)) : {}
```
matthieu-foucault, line 15:
```suggestion
    {
    try {
      return router.query.relayVars ? JSON.parse(router.query.relayVars as string) : {}
     }
      catch (e) {
        return {}
      }
    }
```

app/components/FilterableComponents/FilterableTablePagination.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  // Set the activePage by query string value. If no value exists, the activePage is the first page
+  let activePage = 1;
+  if (router?.query?.pageVars)
+    activePage =
+      JSON.parse(router?.query?.pageVars?.toString())?.activePage || 1;
+
+  const items = [];
+
+  const paginate = (pageNumber: number) => {
```
matthieu-foucault, line 25:
```suggestion
  const handlePageChange = (pageNumber: number) => {
```

dleard:
Pagination bar with Resuts Per Page selection
paginationResults

app/pages/_app.tsx
```diff
@@ -51,16 +51,25 @@ export default class App extends NextApp<AppProps> {
...
 
+    // Safely parse from URL query string, returns {} if query string is invalid JSON
+    const safeParse = (toParse) => {
+      try {
+        return JSON.parse(toParse as string);
+      } catch (e) {
+        return {};
+      }
+    };
+
```
pbastia, line 62: This function probably belongs to `app/lib` to make it reusable and independently testable

app/components/FilterableComponents/FilterableTablePagination.tsx
```diff
@@ -1,42 +1,65 @@
...
   const router = useRouter();
-  const maxPages = Math.ceil(totalCount / maxResultsPerPage);
 
-  // Set the activePage by query string value. If no value exists, the activePage is the first page
-  let activePage = 1;
-  if (router?.query?.pageVars)
-    activePage =
-      JSON.parse(router?.query?.pageVars?.toString())?.activePage || 1;
+  const getPageData = () => {
+    try {
```
pbastia, line 18: should we use the `safeParse()` function here?

dleard: this is functionally the same with the getPageData() using a try/catch block, but if we split safeParse() into a `/lib` then it could definitely be reused here instead of re-writing essentially the same functionality in this component

pbastia (dismissed):

app/components/FilterableComponents/FilterableTableHeaders.tsx
```diff
@@ -11,7 +11,15 @@ const FilterableTableHeaders: React.FunctionComponent<ISearchProps> = (
...
 
-  const [searchFilters, setSearchFilters] = useState({});
+  const [searchFilters, setSearchFilters] = useState(() => {
+    try {
+      return router.query.relayVars
+        ? JSON.parse(router.query.relayVars as string)
+        : {};
+    } catch (e) {
+      return {};
+    }
```
matthieu-foucault, line 21: use `safeJsonParse` here

app/components/FilterableComponents/FilterableTablePagination.tsx
```diff
@@ -0,0 +1,142 @@
...
+  props
+) => {
+  const {totalCount} = props;
+  const router = useRouter();
+
+  const pageData = router.query.pageVars
+    ? safeJsonParse(router.query.pageVars as string)
+    : {};
+
+  const maxResultsPerPage = pageData?.max_results || DEFAULT_MAX_RESULTS;
```
matthieu-foucault, line 22:
```suggestion
  const maxResultsPerPage = pageData.max_results || DEFAULT_MAX_RESULTS;
```

don't need optional chaining here

app/components/FilterableComponents/FilterableTablePagination.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const activePage = pageData?.offset / maxResultsPerPage + 1 || 1;
```
matthieu-foucault, line 23:
```suggestion
  const activePage = pageData.offset / maxResultsPerPage + 1 || 1;
```

same

app/containers/Applications/ApplicationListContainer.tsx
```diff
@@ -10,13 +10,14 @@ import {SortOnlyOption} from 'components/Search/SortOnlyOption';
...
 import {ApplicationListContainer_query} from 'ApplicationListContainer_query.graphql';
+import FilterableTablePagination from 'components/FilterableComponents/FilterableTablePagination';
 
 interface Props {
   query: ApplicationListContainer_query;
 }
 
 export const ApplicationList: React.FunctionComponent<Props> = (props) => {
-  const {edges} = props.query.allApplications;
+  const {edges, totalCount} = props?.query?.allApplications;
```
matthieu-foucault, line 20:
```suggestion
  const {edges, totalCount} = props.query.allApplications;
```
neither props nor query can be null

app/tests/unit/lib/safeJsonParse.test.ts
```diff
@@ -0,0 +1,13 @@
...
+describe('The safeJsonParse function', () => {
+  it('should parse valid JSON', async () => {
+    const jsonString = JSON.stringify({test: 'success'});
+    const testReturn = safeJsonParse(jsonString);
+    expect(testReturn.test).toBe('success');
+  });
+  it('should return {} when passed invalid JSON', async () => {
+    const testReturn = safeJsonParse(';;%%;234;');
+    expect(testReturn).toEqual({});
+  });
```
matthieu-foucault, line 12: add `it('should return {} when passed undefined'` which should work as-is

app/components/FilterableComponents/FilterableTablePagination.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+const FilterableTablePaginationComponent: React.FunctionComponent<Props> = (
+  props
+) => {
+  const {totalCount} = props;
+  const router = useRouter();
+
+  const pageData = router.query.pageVars
+    ? safeJsonParse(router.query.pageVars as string)
+    : {};
```
matthieu-foucault, line 20: safeJsonParse should handle `undefined` as is. No need to have a ternary

matthieu-foucault (changes requested):

pbastia (approved):

matthieu-foucault (approved):

PR #1509, opened 2021-03-03
Feat/filterable organisations

pbastia:
Reworking the organisation requests table for analysts to be searchable and sortable on all columns


app/cypress/integration/search-tables.spec.js
```diff
@@ -97,3 +101,127 @@ describe('When filtering applications', () => {
...
+    );
+    cy.get('body').happoScreenshot({
+      component: 'Organisation Requests List',
+      variant: 'Filtered By Status'
+    });
+    cy.contains('Clear').click();
+    cy.get('#page-content');
+  });
+
+  it('The application can be filtered by multiple search fields', () => {
```
dleard, line 198: I would suggest that we don't need to duplicate ALL the tests from the ApplicationList tests above. It really just tests all the Filterable components over and over. I think since all filtering types are tested above for the ApplicationList that for ciip_user_organisation, facility & product filtering we only really need this single "multiple search" test to ensure the filtering works for each page.

app/pages/analyst/organisation-requests.tsx
```diff
@@ -16,21 +16,27 @@ class OrganisationRequests extends Component<Props> {
...
-            direction: $direction
-            searchField: $searchField
-            searchValue: $searchValue
+            user_id: $user_id
+            first_name: $first_name
+            last_name: $last_name
+            email_address: $email_address
+            operator_name: $operator_name
+            status: $status
+            order_by: $order_by
```
dleard, line 39: A couple things can be refactored in the code below this line:
- remove orderByField & direction from getInitialProps. They are unused. Add a default order_by value if you think it necessary.
- Remove the SearchTable from the render(), just the OrganisationRequestsTable component is needed
- Add pagination (see analyst/applications.tsx for an example)

dleard (changes requested):

app/pages/analyst/organisation-requests.tsx
dleard: My bad, pagination is added at the ApplicationListContainer level (so OrganisationRequestsTable in this flow)

pbastia: I'm wondering if we should make the pagination part of the `FilterableTableLayout` component instead

matthieu-foucault: That's what I was thinking too. We could add a prop to toggle the pagination

matthieu-foucault: maybe wait until all of the search user stories are merged, and we can refactor. I am adding some features to FilterableTableLayout in #1523 , so I'd like to avoid conflicts

pbastia: Haha, I was already done ^^ sounds good, it's not much code and worth avoiding tangling things

dleard (approved):
LGTM :shipit:

PR #1511, opened 2021-03-04
Feat: filterable products

dleard:
- [x] Computed columns for filter/sort
- [x] Refactor fragments
- [x] Add pagination (after #1508 is merged)
- [x] Deprecate SearchProducts function
- [x] pgTap tests
- [x] unit tests
- [x] e2e tests?


app/containers/Products/ProductListContainer.tsx
```diff
@@ -1,140 +1,104 @@
...
-      searchField,
-      searchValue,
-      orderByField,
-      direction,
-      productCount
-    };
-    relay.refetch(refetchVariables);
-  });
+export const ProductList: React.FunctionComponent<Props> = ({query}) => {
+  if (query?.allProducts?.edges) {
```
pbastia, line 23: I would invert that `if` statement for readability
```
if(!query?.allProducts?.edges) return <LoadingSpinner />
...
```

app/containers/Products/ProductRowItemContainer.tsx
```diff
@@ -209,15 +179,13 @@ export const ProductRowItemComponent: React.FunctionComponent<Props> = ({
...
-        console.log(response);
+        await createLinkedProduct(id);
       }
     });
 
     previousLinks.forEach(async (id, index) => {
       if (!newLinks.includes(id)) {
-        const response = await removeLinkedProduct(linkData[index]);
-        console.log(response);
+        await removeLinkedProduct(linkData[index]);
```
pbastia, line 188: is there a way we can bundle up these in a single mutation? Or is it too much work since there aren't many links updated at the same time?

pbastia (changes requested):

app/containers/Products/ProductRowItemContainer.tsx
dleard: It's unlikely there are many links updated at one time. Looking in the database the most links one product has is 4, so probably unnecessary to put the work in to bundle them. Investigation batch mutations could be a tech-debt card, but one we probably don't need to tackle until we have a need for it.

matthieu-foucault: Mutations are already batched in a single HTTP request, so it the gain would be marginal (if a single mutation can even update more than one record?)

matthieu-foucault (approved):

pbastia (approved):

PR #1512, opened 2021-03-05
chore: update cas-postgres chart to 0.8.0

pbastia:


dleard (dismissed):

helm/cas-ciip-portal-data-migrate/values.yaml
```diff
@@ -5,6 +5,6 @@ image:
   psql:
     repository: gcr.io/ggl-cas-storage/cas-postgres
     pullPolicy: Always
-    tag: "0.2.0"
+    tag: "0.8.0"
```
pbastia, line 8: oops, I don't think this appliess here

dleard (dismissed):

matthieu-foucault (approved):

PR #1522, opened 2021-03-08
fix: reset offset when applying new search filters

dleard:
The offset value was not being reset, leaving the user on a non-existent page when the results got pared down via a filter.


matthieu-foucault (approved):

matthieu-foucault:
@dleard the security fix was merged, you can rebase this branch

matthieu-foucault (approved):

PR #1523, opened 2021-03-09
feat: update facilities application search to use filter plugin

matthieu-foucault:
Known bug: when clearing the search, the application status filter does not reset if it is set to "Not started". This might require a bit of a refactor to fix properly, so I'll do that in the next PR, with https://youtrack.button.is/issue/GGIRCS-2302


.circleci/config.yml
```diff
@@ -205,25 +205,7 @@ jobs:
...
-      - run:
-          name: regenerate schema via introspection
-          command: |
-            source ~/.bashrc
-            pushd app
-            yarn build:schema
-            popd
-      - run:
-          name: check for inconsistency between schema ddl & graphql
-          command: git diff --exit-code || (echo 'The contents of the `schema` folder are out of sync with `app/server/schema.graphql`. Did you forget to \"yarn build:relay\"?' && false)
```
matthieu-foucault, line 220: not sure what was the intended test by doing a rebase and regenerating the schema a second time

schema/verify/types/application_search_result.sql
```diff
@@ -4,9 +4,11 @@ begin;
 
 do $$
   begin
-    assert (
-      select true from pg_catalog.pg_type where typname = 'application_search_result'
-    ), 'type "application_search_result" is not defined';
+    if (select true from pg_catalog.pg_type where typname = 'application_search_result') then
+      raise 'type "application_search_result" is not defined';
```
dleard, line 8:
```suggestion
      raise 'type "application_search_result" is defined when it should not be';
```

app/pages/reporter/facilities.tsx
```diff
@@ -18,26 +17,32 @@ class FacilitiesList extends Component<Props> {
...
-      $reportingYear: Int
+      $operatorName: String
+      $facilityName: String
+      $applicationStatus: CiipApplicationRevisionStatus
+      $applicationIdIsNull: Boolean
+      $applicationId: Int
+      $organisationRowId: Int
+      $offset: Int
+      $max_results: Int
+      $reportingYear: Int = 2019
```
dleard, line 28: 2019 hardcoded? Could we init this to the value of openedReportingYear() || 2019 instead?

dleard (changes requested):
Making a computed column off reporting_year with a custom return type was a really good idea!

app/pages/reporter/facilities.tsx
matthieu-foucault: That is to be fixed when working on https://youtrack.button.is/issue/GGIRCS-2135. 
Maybe I should hardcode it to 2020 though.
I don't think that we can do this initialization in graphql. I was thinking of creating a `default_searched_reporting_year()` function (either the currently opened one or the last one that was opened), and creating another `facility_application_by_reporting_year()` function with no parameter, which would use `default_searched_reporting_year()`. I want it to be another function so that they can remain `sql` instead of `plpgsql` (so that the query planner can optimize things)

.circleci/config.yml
pbastia: it might be because relay's `schema.graphql` and `schema.json` are committed, so we're making sure they match our database schema from the same commit? But this should be caught by the e2e at least

app/components/FilterableComponents/FilterableTableLayout.tsx
```diff
@@ -1,29 +1,81 @@
...
+export const FilterableTableLayoutComponent: React.FunctionComponent<Props> = ({
+  searchOptions,
+  body,
+  isLoading,
+  extraFilters = []
+}) => {
+  const router = useRouter();
+  const relayVars = useMemo(
+    () => safeJsonParse(router.query.relayVars as string),
+    [router]
```
pbastia, line 24: I'm not really familiar with how useMemo detects changes, isn't the router object going to be the same every time, and this won't trigger when the relayVars change?

pbastia, line 24: I'm not really familiar with how useMemo detects changes, isn't the router object going to be the same every time, and this won't trigger when the relayVars change?

matthieu-foucault: No, the router should be immutable (otherwise it breaks react conventions) and be a new object every time it changes

app/components/FilterableComponents/FilterableTableLayout.tsx
```diff
@@ -1,29 +1,81 @@
...
+    });
+
+    const queryString = JSON.stringify(newQuery);
+
+    const url = {
+      pathname: router.pathname,
+      query: {
+        ...router.query,
+        relayVars: queryString,
+        pageVars: JSON.stringify({offset: 0})
```
pbastia, line 48: This is where we lose the pagination 'result per page' setting on reset. It is also breaking SRP, the filtering should not know about pagination specifics

app/containers/Facilities/FacilitiesListContainer.tsx
```diff
@@ -1,180 +1,159 @@
...
-      selectedReportingYear={selectedReportingYear}
-    />
-  );
+  const selectableReportingYears = [];
+  reportingYears.forEach(({node}) => {
+    if (node.reportingYear < new Date().getFullYear())
+      selectableReportingYears.push(node.reportingYear);
+  });
+
+  const reportingPeriodFilter: ISearchExtraFilter = {
```
pbastia, line 70: This could probably go in its own class/file

app/components/FilterableComponents/FilterableTableLayout.tsx
matthieu-foucault: I am planning to move the page change handling in this component (so that all interactions with the router are constrained to the one file)

pbastia: Should we break it in 2 pieces then?
- a small non-UI, pure TS, router interaction service
- the `FilterableTableLayout` that deals with react rendering

pbastia (approved):

dleard (approved):

PR #1525, opened 2021-03-10
Feat: review step database entities

dleard:
Database setup work for the front-end review process refactoring here https://youtrack.button.is/issue/GGIRCS-2293

- [x] Add review_step_name table
- [x] Add review_step table
- [x] Add create/refresh trigger function
- [x] Apply trigger to application_revision_status
- [x] pgTap tests


dleard:
There is a wrinkle in adding the trigger on submission of the application. I've given the reporter no RLS privileges to the review_step table, but since the trigger fires on submission of an application (when the reporter is using the app) it borked because they don't have insert privileges. 
Three options: 
1) give the reporter insert privileges on review_step 
2) find somewhere else to apply the trigger
3) make the trigger function `security definer`
I went with option 3 & added `security definer` to the function, which elevates their privileges only when running it. But I'm up for discussion on whether another option is better.

dleard:
I also created a tech debt card to look at & rework the _100_timestamps function (https://youtrack.button.is/issue/GGIRCS-2321). It was written a long time ago and does some things I don't really like:
1) if you want to use it, you must have all 3 sets of [created_at/by, updated_at/by, deleted_at/by]) regardless of if you need all 3 or it barfs.
2) it only sets the deleted_by/at when an 'is_deleted' column is present where I'd like to be set it automatically when other column names like 'is_active' is changed.

It might be beneficial for me to to do the function rework before merging this PR.

schema/deploy/trigger_functions/create_or_refresh_review_step.sql
```diff
@@ -0,0 +1,47 @@
...
+    select id, step_name from ggircs_portal.review_step_name where is_active = true
+  loop
+    if exists(select * from ggircs_portal.review_step where application_id = new.application_id and review_step_name_id = temp_row.id) then
+      update ggircs_portal.review_step set is_complete = false where application_id = new.application_id and review_step_name_id = temp_row.id;
+    else
+      insert into ggircs_portal.review_step(application_id, review_step_name_id)
+        values (new.application_id, temp_row.id);
+    end if;
+  end loop;
+
```
pbastia, line 24: Wild suggestion, this might be completely out of scope:
We could use an upsert instead if we create a unique index on both columns - this would have the benefit of guaranteeing that we only have at most one instance of each review_step_name per application.

```SQL
create unique index ggircs_portal_review_step_on_app_id_and_review_step_name_id on ggircs_portal.review_step(application_id, review_step_name_id);

insert into ggircs_portal.review_step(application_id, review_step_name_id)
        values (new.application_id, temp_row.id)
        on conflict (application_id, review_step_name_id)
        do update set is_complete=false;

File: schema/deploy/trigger_functions/create_or_refresh_review_step.sql [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+  for temp_row in
+    select id, step_name from ggircs_portal.review_step where is_active = true
+  loop
+    if exists(select * from ggircs_portal.application_review_step where application_id = new.application_id and review_step_id = temp_row.id) then
+      update ggircs_portal.application_review_step set is_complete = false where application_id = new.application_id and review_step_id = temp_row.id;
+    else
+      insert into ggircs_portal.application_review_step(application_id, review_step_id)
+        values (new.application_id, temp_row.id);
+    end if;
```
pbastia, line 22:
```suggestion
    insert into ggircs_portal.review_step(application_id, review_step_name_id) 
        values (new.application_id, temp_row.id) 
        on conflict (application_id, review_step_name_id) 
        do update set is_complete=false;
```
Would this work, and remove the need for the `if` statement?

dleard: definitely, I've updated it

pbastia (approved):

pbastia (approved):

PR #1526, opened 2021-03-10
Reporter-facing side of application review

kriscooke:
On `/reporter/view-application`:
- See [screenshots](https://youtrack.button.is/issue/GGIRCS-2297#focus=streamItem-4-4629.0-0) of all the possible decision/revision/submission states that an application can be in.
- Moves the 'View latest submission' and 'Resume latest draft' buttons into the main page for DOM reasons

On `/reporter/application`:
- Replaces ApplicationCommentsContainer with new revamped ApplicationDecision component (re-used from the above page) - see [screenshot](https://youtrack.button.is/issue/GGIRCS-2297#focus=streamItem-4-4692.0-0)

[GGIRCS-2297](https://youtrack.button.is/issue/GGIRCS-2297)

### DIFF
--- a/app/components/Application/ApplicationDecision.tsx
+++ b/app/components/Application/ApplicationDecision.tsx
@@ -1,0 +1,56 @@
+import React from 'react';
+import {Alert} from 'react-bootstrap';
+
+interface Props {
+  changesRequested: Boolean;
+  decision?: string;
+  reviewComments: string[];
+}
+
+const BS_ALERT = {
+  REJECTED: 'danger',
+  APPROVED: 'success'
+};
+
+const DECISION_TEXT = {
+  REJECTED:
+    'Thank you for applying to the CIIP. Unfortunately this submission did not meet program requirements and a payment cannot be granted.',
+  APPROVED:
+    'This application has been approved. Your incentive payment is being processed. No further action is required from you at this time.'
+};
+
+const ApplicationDecision: React.FunctionComponent<Props> = ({
+  changesRequested,
+  decision,
+  reviewComments,
+  children
+}) => {
+  const heading = changesRequested
+    ? 'Action Required'
+    : `Decision: ${decision[0].toUpperCase()}${decision
+        .toLowerCase()
+        .slice(1)}`;
+
+  // eslint-disable-next-line react/no-array-index-key
+  const items = reviewComments.map((c, i) => <li key={`comment-${i}`}>{c}</li>);
+
+  return (
+    <Alert variant={changesRequested ? 'danger' : BS_ALERT[decision]}>
+      <Alert.Heading>{heading}</Alert.Heading>
+      <p>
+        {changesRequested
+          ? 'Changes to your application have been requested. Please read these review notes and gather the relevant information before revising your application.'
+          : DECISION_TEXT[decision]}
+      </p>
+      {reviewComments.length > 0 && (
+        <>
+          <p className="h5">Reviewer Notes:</p>
+          <ul>{items}</ul>
+        </>
+      )}
+      {children}
+    </Alert>
+  );
+};
+
+export default ApplicationDecision;
--- a/app/components/Application/ApplicationDecision.tsx
+++ b/app/components/Application/ApplicationDecision.tsx
@@ -1,18 +1,21 @@
 import React from 'react';
 import {Alert} from 'react-bootstrap';
+import {getUserFriendlyStatusLabel} from 'lib/text-transforms';
 
 interface Props {
-  changesRequested: Boolean;
-  decision?: string;
+  actionRequired: Boolean;
+  decision: string;
   reviewComments: string[];
 }
 
 const BS_ALERT = {
+  REQUESTED_CHANGES: 'warning',
   REJECTED: 'danger',
   APPROVED: 'success'
 };
 
 const DECISION_TEXT = {
+  REQUESTED_CHANGES: '',
   REJECTED:
     'Thank you for applying to the CIIP. Unfortunately this submission did not meet program requirements and a payment cannot be granted.',
   APPROVED:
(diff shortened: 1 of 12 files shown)


app/pages/reporter/view-application.tsx
```diff
@@ -59,7 +77,54 @@ class ViewApplication extends Component<Props> {
...
+      (result) => result.node.description
+    );
+    const status =
+      query?.application?.applicationRevisionStatus?.applicationRevisionStatus;
+    const changesRequested = status === 'REQUESTED_CHANGES';
+    const hasBeenReviewed = status !== 'SUBMITTED' && status !== 'DRAFT';
+
+    const thisVersion = Number(this.props.router.query.version);
+    const latestSubmittedRevision =
+      query.application.latestSubmittedRevision.versionNumber;
```
dleard, line 90:
```suggestion
      query.application.latestSubmittedRevision?.versionNumber;
```

app/pages/reporter/view-application.tsx
```diff
@@ -59,7 +77,54 @@ class ViewApplication extends Component<Props> {
...
+    const latestDraftRevision =
+      query.application.latestDraftRevision.versionNumber;
```
dleard, line 92:
```suggestion
      query.application.latestDraftRevision?.versionNumber;
```

latestSubmittedRevision and latestDraftRevision can be null and need optional chaining here. This is what is breaking the cypress test.

app/components/Application/ApplicationDecision.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import {Alert} from 'react-bootstrap';
+
+interface Props {
+  actionRequired: Boolean;
```
matthieu-foucault, line 5:
```suggestion
  actionRequired: boolean;
```
using the boolean type instead of the class is preffered

app/components/Application/ApplicationDecision.tsx
```diff
@@ -0,0 +1,65 @@
...
+    'Thank you for applying to the CIIP. Unfortunately this submission did not meet program requirements and a payment cannot be granted.',
+  APPROVED:
+    'This application has been approved. Your incentive payment is being processed. No further action is required from you at this time.'
+};
+
+const DECISION_PRETTY = {
+  REQUESTED_CHANGES: 'Changes Requested',
+  REJECTED: 'Rejected',
+  APPROVED: 'Approved'
+};
```
matthieu-foucault, line 28: You can use `getUserFriendlyStatusLabel` instead here

app/components/Application/ApplicationDecision.tsx
```diff
@@ -0,0 +1,65 @@
...
+  actionRequired,
+  decision,
+  reviewComments,
+  children
+}) => {
+  const heading = actionRequired
+    ? 'Action Required'
+    : decision === 'REQUESTED_CHANGES'
+    ? 'Changes Requested'
+    : `Decision: ${DECISION_PRETTY[decision]}`;
```
matthieu-foucault, line 40: Not a big fan of double ternary operators, can you use if/else instead?

app/components/Application/ApplicationDecision.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  // eslint-disable-next-line react/no-array-index-key
```
matthieu-foucault, line 42: It is okay to disable it here because the comments array will not be updated while this component is mounted. Can you add a comment to justify disabling the linter?

app/containers/Applications/ApplicationWizard.tsx
```diff
@@ -42,6 +43,11 @@ export const ApplicationWizardComponent: React.FunctionComponent<Props> = ({
     latestSubmittedRevision,
     applicationRevisionByStringVersionNumber: applicationRevision
   } = application;
+  const reviewComments = query.application.reviewCommentsByApplicationId?.edges.map(
+    ({node: description}) => description
```
matthieu-foucault, line 47:
```suggestion
    ({node: {description}}) => description
```
what you wrote above renames `node` to `description`

app/containers/Applications/ApplicationWizard.tsx
```diff
@@ -42,6 +43,11 @@ export const ApplicationWizardComponent: React.FunctionComponent<Props> = ({
...
+  );
+  const revisionInProgress =
+    latestSubmittedRevision?.versionNumber < latestDraftRevision.versionNumber;
```
matthieu-foucault, line 50: would `latestDraftRevision.versionNumber > 1`  work too?

Side not. I got very confused because "review" and "revision" are very similar words.

matthieu-foucault (changes requested):

app/components/Application/ApplicationDecision.tsx
kriscooke: You'd hope I'd remember a utility function that I wrote myself 🤣

app/containers/Applications/ApplicationWizard.tsx
kriscooke: What I was probably thinking is that the above would cover an edge case where they were viewing this page (from a pasted link) with `version` equal to the latest submitted version. 

But of course this page actually redirects them to view-application, so that wouldn't happen. I suppose both comparisons are about equal in that case.

app/components/Application/ApplicationDecision.tsx
kriscooke: I guess I gravitated that way because the if/else if/else alternative requires re-assignment so `heading` can't be a constant, which it should be ideally. 🤔 

I broke it into separate lines for readability. I'm not sure if the below is actually more readable?

```
  let heading;
  if (actionRequired) {
    heading = 'Action Required';
  } else if (decision === 'REQUESTED_CHANGES') {
    heading = getUserFriendlyStatusLabel(decision);
  } else {
    heading = `Decision: ${getUserFriendlyStatusLabel(decision)}`;
  }
```

matthieu-foucault: It's a bit more verbose, but I think that the order of operations is more explicit

matthieu-foucault (approved):

PR #1527, opened 2021-03-10
Feat: remove certifier UI

pbastia:
Removing certifier UI elements
- removes certification wizard steps and prompts for application submission right away
- removes certification-specific pages
- updates tests and snapshots to reflect that

trigger on `application_revision_status` has already been removed by @dleard in #1524


matthieu-foucault (changes requested):
`react-signature-canvas` can be removed from our dependencies

matthieu-foucault (approved):

PR #1529, opened 2021-03-12
fix: don't drop ciip_portal user during revert

dleard:
dropping the user during revert caused an issue during `make test` when both the local _dev and _test databases existed. The deploy does a `create if not exists...` anyway.

### DIFF
--- a/.bin/sqitch-check-immutable-files.sh
+++ b/.bin/sqitch-check-immutable-files.sh
@@ -4,8 +4,8 @@
 # usage: sqitch-immutable-files.sh schema_dir base_branch
 
 set -euo pipefail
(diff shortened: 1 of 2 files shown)


matthieu-foucault:
One issue if we merge this is that `sqitch-files-before-tag-are-immutable` will not pass until we make a release, which means that Shipit will not do any continuous deployment in our dev environment.

matthieu-foucault:
Maybe we want `sqitch-files-before-tag-are-immutable` to only look at the `deploy` scripts?

dleard:
> Maybe we want `sqitch-files-before-tag-are-immutable` to only look at the `deploy` scripts?

This makes sense to me. The point of the check is really to make sure that whatever is being deployed is not changed post-release. I don't see any reason to be checking the revert/verify files

matthieu-foucault (approved):

PR #1536, opened 2021-03-15
Feat/search filters refactor

matthieu-foucault:


app/components/FilterableTable/FilterableTablePagination.tsx
```diff
@@ -87,19 +72,18 @@ const FilterableTablePaginationComponent: React.FunctionComponent<Props> = (
...
-              100
-            </Dropdown.Item>
+            {PAGE_SIZE_OPTIONS.map((option) => (
+              <Dropdown.Item
+                ref={option}
+                onClick={() => onPageSizeChange(option)}
+              >
+                {option}
+              </Dropdown.Item>
+            ))}
```
pbastia, line 86: Since this is static we should compute it outside of the rendering loop

matthieu-foucault: `onPageSizeChange` isn't static though

app/components/FilterableTable/FilterableTable.tsx
```diff
@@ -1,41 +1,50 @@
...
-    () => safeJsonParse(router.query.relayVars as string),
+  const filterArgs = useMemo<FilterArgs>(
+    () => safeJsonParse(router.query.filterArgs as string),
+    [router]
+  );
+
+  const {offset, pageSize} = useMemo<PageArgs>(
+    () => safeJsonParse(router.query.pageArgs as string),
     [router]
   );
```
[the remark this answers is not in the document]
matthieu-foucault: We actually can rerender when the data changes, when adding products for instance. The impact either way is marginal

app/components/FilterableTable/FilterableTablePagination.tsx
pbastia: oh right!

pbastia (approved):

PR #1538, opened 2021-03-16
Application Review: New UI and deprecate unnecessary comment types

dleard:
- New UI for viewing application review comments (contained in #1538)
- DB components of both [GGIRCS-2193](https://youtrack.button.is/issue/GGIRCS-2193) and [GGIRCS-2293](https://youtrack.button.is/issue/GGIRCS-2293).)


dleard:
@kriscooke Handoff notes:
- There is a very rough ApplicationReviewStep component in "components/Applications/"
- The pages/application-review fragment has been refactored to return the comments based on comment type. These queries still need connections defined (there is an example in Organisations.tsx). It MAY be necessary to pull the comments into the component and make the component a container with its own fragment, I've never tried to create a connection on a 'root' page before.
- The create/update/delete mutations for reviewComments will need updating with the connection keys (I left 'stand-in' keys in the range_add / updater functions)

app/components/Application/ApplicationReviewStep.tsx
```diff
@@ -0,0 +1,84 @@
...
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE COMMENTS GO HERE
+            COMMENTS GO HERE
```
matthieu-foucault, line 53: so where do comments go?

kriscooke (commented):
@matthieu-foucault For your work removing query params and moving pages: I noticed when I was trying to fix unit tests for pages broken by the migration, that the jest test for `pages/reporter/application/[etc]/version/[etc]/view` won't run and seems to be unrecognized (not matched) by jest as a test file. Here I've moved the test to a simplified path so I could run the test.

app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx
```diff
@@ -84,9 +90,17 @@ class ViewApplication extends Component<Props> {
...
+      application?.applicationReviewStepsByApplicationId.edges;
+    // Merge review comments from all applicationReviewSteps into one list:
+    const reviewComments = reviewSteps?.reduce((mergedStepComments, step) => {
+      const stepComments =
+        step.node.reviewCommentsByApplicationReviewStepId.edges;
+      return [
+        ...mergedStepComments,
+        ...stepComments.map((step) => step.node.description)
+      ];
+    }, []);
```
Comment (pbastia) [inline app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx:103]: This should probably be abstracted away since it's the same as in `ApplicationWizard.tsx:47`

schema/verify/computed_columns/form_result_has_unresolved_comments.sql
```diff
@@ -2,6 +2,16 @@
...
+  begin
+
+    if (select exists(select * from pg_proc where proname='ggircs_portal.form_result_has_unresolved_comments')) then
+      raise exception 'ggircs_portal.form_result_has_unresolved_comments exists when it should not';
+    else
+      perform true;
+    end if;
+
+  end;
+$$;
```
pbastia, line 15: we can use the `ggircs_portal_private.verify_function_not_present( ... )` helper for all these sqitch verify procedures

schema/verify/computed_columns/form_result_internal_general_comments.sql
```diff
@@ -2,6 +2,16 @@
...
+  begin
+
+    if (select exists(select * from pg_proc where proname='ggircs_portal.form_result_internal_general_comments')) then
+      raise exception 'ggircs_portal.form_result_internal_general_comments exists when it should not';
+    else
+      perform true;
+    end if;
+
+  end;
+$$;
```
pbastia, line 15: same

schema/verify/computed_columns/form_result_requested_change_comments.sql
```diff
@@ -2,6 +2,16 @@
...
+  begin
+
+    if (select exists(select * from pg_proc where proname='ggircs_portal.form_result_requested_change_comments')) then
+      raise exception 'ggircs_portal.form_result_requested_change_comments exists when it should not';
+    else
+      perform true;
+    end if;
+
+  end;
+$$;
```
pbastia, line 15: same

schema/verify/mutations/create_review_comment_mutation_chain.sql
```diff
@@ -2,6 +2,16 @@
...
+  begin
+
+    if (select exists(select * from pg_proc where proname='ggircs_portal.create_review_comment_mutation_chain')) then
+      raise exception 'ggircs_portal.create_review_comment_mutation_chain exists when it should not';
+    else
+      perform true;
+    end if;
+
+  end;
+$$;
```
pbastia, line 15: same

pbastia (changes requested):

pbastia (commented):
Couple of very minor changes that could also be pushed to tech debt. Great work!

app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx
Comment (kriscooke) [reply app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx:103]: Yes it occurred to me... often I will let myself copy/paste something ONCE and then abstract once I need it a third time. 
Not sure where would be appropriate to put that? `lib/`? I see we now also have a `functions/` folder (but not sure what is the difference)?

Comment (matthieu-foucault) [reply app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx:103]: Looks like this needs to be moved down to `ApplicationDecision`, with a fragment on `Application`

Comment (pbastia) [reply app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx:103]: it would be some sort of polyfill for `array.flatMap()` which has IE compatibility issues from what I understand, I'd say `lib/` is great for that

Comment (kriscooke) [reply app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx:103]: > Looks like this needs to be moved down to ApplicationDecision, with a fragment on Application

Was kind of trying to avoid that, as I felt that surely we could merge some arrays without needing to create a new fragment container? The ApplicationDecision doesn't want any of the original data about a `ReviewComment` - all it wants is an array of strings to put in a bulleted list.

schema/deploy/tables/review_comment_001.sql
```diff
@@ -0,0 +1,35 @@
...
+  returns void as $$
+declare
+  temp_row record;
+  temp_id int;
+begin
+    for temp_row in
+      select id from ggircs_portal.application
+    loop
+      insert into ggircs_portal.application_review_step(application_id, review_step_id)
+      values (temp_row.id, 1)
```
matthieu-foucault, line 21: Where is this review step created? I was expecting to see it in `tables/application_review_step` but I don't

dleard: This function does two things: creates legacy application_review_steps for all existing applications and attaches all comments for the application to that legacy step. I suppose it could be two functions, one in the `application_review_step` migration to create the legacy steps and another here in this migration to connect all comments to the legacy `application_review_step`. I can split this function up to make sure the data is migrated more clearly.

All new application_review_steps are created via a trigger on application submit.

matthieu-foucault: I was confused where the `1` was coming from, but that's from the `review_step` table, not `application_review_step`.
Maybe replace the `1` with `(select id from ggircs_portal.review_step where step_name = 'legacy')` to make it explicit?

schema/verify/computed_columns/form_result_has_unresolved_comments.sql
dleard: Thanks Pierre, I wrote this before your verify_function_not_present was merged. I'll refactor.

schema/deploy/tables/review_comment_001.sql
matthieu-foucault: In view of @kriscooke 's comment in https://github.com/bcgov/cas-ciip-portal/pull/1574#issuecomment-816927193, maybe the `step_name` should be `other` instead of `legacy`

dleard: :+1:  I can make both those changes

app/tests/unit/containers/Applications/ApplicationWizard.test.tsx
```diff
@@ -1,8 +1,50 @@
 import React from 'react';
 import {shallow} from 'enzyme';
 import {ApplicationWizardComponent} from 'containers/Applications/ApplicationWizard';
+import {ApplicationWizard_query} from 'ApplicationWizard_query.graphql';
 import ViewApplication from 'pages/reporter/application/[applicationId]/version/[versionNumber]/view';
 
+const getTestQuery = () => {
```
kriscooke, line 7: I've started this pattern in #1574 as well - using a factory function to define test queries because it was such a pain to update them all.

I want to ideally update in one place next time!  🤪

schema/verify/computed_columns/form_result_has_unresolved_comments.sql
dleard: ahh, I can't refactor. My files exist before that function is created

pbastia: Hard to keep track of which branch started when, it's definitely not an essential improvement

matthieu-foucault: Reordering the changes in the sqitch plan is an option, as long as it's all after the latest tag

kriscooke (changes requested):
Just "requesting changes" as a precaution to prevent this from being merged without #1574, now that this PR is no longer marked as Draft.

pbastia (dismissed):

kriscooke:
🤔 Interesting - my f3f5bad change doesn't seem to have helped the flakiness of the `updateApplicationReviewStepMutation` test

app/tests/unit/mutations/application_review_step/updateApplicationReviewStepMutation.test.ts
```diff
@@ -0,0 +1,61 @@
...
+        id: 'abc',
+        applicationReviewStepPatch: {
+          isComplete: true
+        }
+      }
+    });
+    expect(test).toBeDefined();
+    expect(
+      typeof test.data.updateApplicationReviewStep.applicationReviewStep
+        .isComplete
```
matthieu-foucault, line 58:
```suggestion
      test.data.updateApplicationReviewStep.applicationReviewStep.isComplete
```

kriscooke (approved):

PR #1539, opened 2021-03-17
fix: Make the 'Resume application' button a real anchor link

kriscooke:
One last button to make into a real link, following [GGIRCS-2226](https://youtrack.button.is/issue/GGIRCS-2226).

### DIFF
--- a/app/cypress/integration/reporter-facilities-list.spec.js
+++ b/app/cypress/integration/reporter-facilities-list.spec.js
@@ -15,11 +15,7 @@ describe('When the reporter is viewing a list of facilities & applications', () 
       'have.value',
       2019
     );
-    cy.get('tbody > tr:nth-child(1) .btn').should(
-      'have.prop',
-      'disabled',
-      false
-    );
+    cy.get('tbody > tr:nth-child(1) a.btn').should('have.attr', 'href');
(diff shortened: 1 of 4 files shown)


matthieu-foucault (changes requested):
I think that there is a bunch of `button` selectors that need to be updated/removed in the Cypress tests

app/cypress/integration/reporter-facilities-list.spec.js
```diff
@@ -15,11 +15,7 @@ describe('When the reporter is viewing a list of facilities & applications', ()
       'have.value',
       2019
     );
-    cy.get('tbody > tr:nth-child(1) .btn').should(
-      'have.prop',
-      'disabled',
-      false
-    );
```
kriscooke, line 22: @matthieu-foucault It appears that the Resume button is never disabled anywhere anyway, so it's a bit funny that we were using this assertion.

Anchor links don't have a `disabled` prop, which is why it was failing. Since the intention of this test was to ensure that the button/link is active, I've replace with an assertion that it has an `href` although I'm not even sure that's necessary

matthieu-foucault: That assertion might have been relevant at some point in the past

matthieu-foucault (approved):

PR #1540, opened 2021-03-17
Feat: naics codes table

dleard:
Adds a naics table & a custom mutation that upserts the row on conflict


schema/deploy/mutations/custom_create_naics_mutation.sql
```diff
@@ -0,0 +1,20 @@
+-- Deploy ggircs-portal:mutations/custom_create_naics_mutation to pg
+-- requires: tables/naics
+
+begin;
+
+create or replace function ggircs_portal.custom_create_naics_mutation(
+  naics_code_input int,
```
matthieu-foucault, line 7: A naics code is a varchar made of numbers, not an int.
I has to follow a lexicographic order, e.g. 11 is closer to 1111 than to 21.
In general, all of these "code" fields should be a varchar, because you never know when someone is going to decide that a new `11-a`  code will be created

schema/deploy/mutations/custom_create_naics_mutation.sql
```diff
@@ -0,0 +1,20 @@
+-- Deploy ggircs-portal:mutations/custom_create_naics_mutation to pg
+-- requires: tables/naics
+
+begin;
+
+create or replace function ggircs_portal.custom_create_naics_mutation(
```
matthieu-foucault, line 6: can we name it `create_naics_mutation`? I guess we don't want to use the standard postgraphile mutation, so it would be good if our function was to override it

schema/deploy/tables/naics.sql
```diff
@@ -0,0 +1,42 @@
+-- Deploy ggircs-portal:tables/naics to pg
+-- requires: schema_ggircs_portal
+
+begin;
+
+create table ggircs_portal.naics (
+  naics_code integer not null primary key,
```
matthieu-foucault, line 7: same comment regarding naics_code being a varchar. Feel free to add a generated integer id and a unique index.

matthieu-foucault (changes requested):

schema/deploy/mutations/custom_create_naics_mutation.sql
dleard: I think there's maybe value in making the name different because then it's easy to see in the front-end mutation definition that we are using a custom function from our database rather than the default. I suppose comments in the front-end mutation would be enough, as long as we remember to write them. Thoughts?

matthieu-foucault: Yeah but we never want to use the default mutation, right? So if we can override it (or omit the generation of the default create mutation for this table), then there is no ambiguity when looking at the graphql API

dleard: Cool, makes sense. I'll make a note in the card to add some comments so it's still clear that we're using a custom mutation

schema/verify/policies/naics_policies.sql
```diff
@@ -0,0 +1,7 @@
+-- Verify ggircs-portal:policies/naics_policies on pg
+
+BEGIN;
+
+-- XXX Add verifications here.
+
```
matthieu-foucault, line 6: any verify needed here?

matthieu-foucault (changes requested):

schema/deploy/tables/naics.sql
```diff
@@ -0,0 +1,52 @@
...
+perform ggircs_portal_private.grant_permissions('select', 'naics', 'ciip_analyst');
+
+-- Grant ciip_analyst permissions
+perform ggircs_portal_private.grant_permissions('select', 'naics', 'ciip_industry_user');
+
+end
+$grant$;
+
+-- Enable row-level security
+alter table ggircs_portal.naics enable row level security;
```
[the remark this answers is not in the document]
dleard: you're right, not needed here since the grants take care of the permissions. I will remove the policy stuff

schema/verify/policies/naics_policies.sql
dleard: policy verify files (and really, all non-schema related verifies) are...odd & should maybe just be in the scope of the unit tests. @pbastia and I ran into an issue with verifying policies where we either need to update all the verify files (including previous migrations where we used the `_001` method & not a `rework`) when we remove policies or agree that the policies are out of scope of the verify & should be the job of the tests (All tables have RLS tests that will fail if a necessary policy is missing). Sequentially verifying the files during a deploy works fine, but the `sqitch verify` command only works over the final state of the db and will kerplode if a non-reworked verify file asserts that a policy should exist and it doesn't in the final state.

example:
**policy_a_verify**
- assert x exists
- asset y exists
**policy_a_001_verify** (deploy file removes x)
- assert y exists

Because `policy_a` is different than `policy_a_001` in sqitch's mind, it runs both verify files & when it sees x does not exist, throws an exception in the `policy_a_verify` file. We may want to rethink what goes in a verify file (schema related only?) and what should be checked in tests, especially when using the `_001` to perform migrations.

dleard: This particular verify file is going away though, since we don't need policies for this table :P

matthieu-foucault (approved):

PR #1543, opened 2021-03-19
refactor: order by id when retrieving application status

dleard:
ordering by created_at is causing issues with timetravel on dev and is unnecessary since our id primary key is sequential and can be used in the order by to get the latest created row.


pbastia (approved):

pbastia:
Beautiful

matthieu-foucault (approved):

PR #1557, opened 2021-03-22
fix: default layout top padding should only be 60px

matthieu-foucault:


app/layouts/default-layout.tsx
```diff
@@ -153,7 +153,7 @@ const DefaultLayout: React.FunctionComponent<Props> = ({
 
           @media screen and (min-width: 992px) {
             #page-content {
-              padding-top: 110px;
+              padding-top: 60px;
```
kriscooke, line 156: This should have one exception so we can keep the existing top padding on the homepage. You can do that by adding the following rule in `pages/index`. (before this can work you'll need to replace the first `<Row>` with `<div className="row">`):

```
          @media screen and (min-width: 992px) {
            .row:first-child {
              padding-top: 50px;
            }
          }
```

kriscooke (changes requested):

app/layouts/default-layout.tsx
matthieu-foucault: why do we want to keep the existing 110px padding on the home page?

kriscooke: It's for aesthetic reasons: The current whitespace balance creates a centering effect - thus emphasis - on the welcome content (in addition to consistency).

The whitespace shown here between the 3 main content blocks repeats throughout this page, including between the lower content and the footer, and between the top block and the header. The 60px default changes that and creates a mild crowding effect by squishing the top content block up into the header.
 
Current layout:
Screen Shot 2021-03-22 at 12 40 59 PM
currently lets the

kriscooke: (in other words, I suspect the original 110px setting may have been introduced specifically to balance the homepage, but unintentionally applied to the whole site via the default template).

app/pages/index.tsx
```diff
@@ -152,6 +152,9 @@ export default class Index extends Component<Props> {
           </Col>
         </div>
         <style jsx>{`
+          .row {
+            padding-top: 50px;
+          }
```
kriscooke, line 157: We only want to add the padding to the very first row (`:first-child`).

I did also forget to mention that we don't need or want the extra padding on smaller screens, which is why I put the suggestion above in a media query like so:
```
@media screen and (min-width: 992px) {
  .row:first-child {
    padding-top: 50px;
  }
}
```

matthieu-foucault: Right, sorry my coffee hasn't kicked in yet :sleepy:

kriscooke (approved):

PR #1564, opened 2021-03-24
Feat/naics code dropdown in admin form

pbastia:
NAICS code on the administration data form is a dropdown, populated with the official NAICS list approved by CAS.

- [x] functionality
- [x] validation if NAICS is not in allowed values ?
- [x] unit tests
- [x] e2e


app/containers/Forms/NaicsField.tsx
```diff
@@ -0,0 +1,62 @@
...
+  );
+
+  const fieldProps = {
+    ...props,
+    schema: {
+      ...props.schema,
+      enum: naicsList,
+      enumNames: props.query.naicsCodes.edges.map((edge) => {
+        const naicsCode = edge.node;
+        return `${naicsCode.naicsCode} - ${naicsCode.ciipSector} - ${naicsCode.naicsDescription}`;
```
matthieu-foucault, line 23:
```suggestion
        return `${naicsCode.naicsCode} - ${naicsCode.naicsDescription}`;
```

I think that the CIIP sector is more for internal use

app/containers/Forms/NaicsField.tsx
```diff
@@ -0,0 +1,62 @@
...
+  query: NaicsField_query;
+}
+
+export const NaicsFieldComponent: React.FunctionComponent<Props> = (props) => {
+  const naicsList = props.query.naicsCodes.edges.map(
+    (edge) => edge.node.naicsCode
+  );
+
+  const fieldProps = {
+    ...props,
```
matthieu-foucault, line 17: the `query` prop doesn't need to be passed down to the field.
```suggestion
    ...props,
    query: undefined,
```

app/containers/Forms/NaicsField.tsx
```diff
@@ -0,0 +1,62 @@
...
+    schema: {
+      ...props.schema,
+      enum: naicsList,
+      enumNames: props.query.naicsCodes.edges.map((edge) => {
```
matthieu-foucault, line 21: You could move this `map` call in a `useMemo`, so that it's not called on every render

app/containers/Forms/NaicsField.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const invalidNaicsAlert =
+    props.formData && !naicsList.includes(props.formData) ? (
+      <Alert variant="danger" className="mt-3 mb-0">
+        <strong>Warning:</strong> The naics code submitted to SWRS is not in the
+        list of valid codes for this program. Please refer to the guidance
+        documents and select an appropriate NAICS code.
+      </Alert>
+    ) : (
+      <></>
```
matthieu-foucault, line 36:
```suggestion
      null
```
null us fine, no need for an empty fragment

app/containers/Forms/NaicsField.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {Alert} from 'react-bootstrap';
+import {createFragmentContainer, graphql} from 'react-relay';
+import {NaicsField_query} from '__generated__/NaicsField_query.graphql';
+
+interface Props extends FieldProps<string> {
+  query: NaicsField_query;
+}
+
+export const NaicsFieldComponent: React.FunctionComponent<Props> = (props) => {
+  const naicsList = props.query.naicsCodes.edges.map(
```
matthieu-foucault, line 12: use a `useMemo` to avoid creating a new list on every render

app/containers/Forms/NaicsField.tsx
```diff
@@ -0,0 +1,62 @@
...
+export default createFragmentContainer(NaicsFieldComponent, {
+  query: graphql`
+    fragment NaicsField_query on Query {
+      naicsCodes: allNaicsCodes {
+        edges {
+          node {
+            rowId
+            naicsCode
+            naicsDescription
+            ciipSector
```
matthieu-foucault, line 56:
```suggestion
            naicsCode
            naicsDescription
```

app/tests/unit/containers/Forms/NaicsField.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const props = {
+      ...initialProps,
+      formData: ''
+    };
+
+    const componentUnderTest = mount(<NaicsFieldComponent {...props} />);
+    expect(componentUnderTest).toMatchSnapshot();
+  });
+
+  it('Displays (<code> - <industry> - <description>) when the form data contains a 6-digit NAICS code', () => {
```
matthieu-foucault, line 47:
```suggestion
  it('Displays (<code>  - <description>) when the form data contains a 6-digit NAICS code', () => {
```
Update when removing the industry

matthieu-foucault (changes requested):

matthieu-foucault (approved):

PR #1565, opened 2021-03-25
refactor: naics UI improvements

dleard:
- [x] More descriptive language in the delete confirmation dialog
- [x] Two action buttons for clarity: "Delete" (red) and "Cancel" ('secondary' dark grey)
- [x] Acronyms like NAICS should be capitalized ("New Naics Code" button -> "New NAICS Code")
- [x] When a new code is added, on re-opening the New dialog again to add a subsequent one, the form validation styles should be cleared (not showing red/green).
- [x] Warning & block mutation when adding a NAICS code that already exists AND is currently active (follow the pattern: delete -> re-add)

### DIFF
--- a/app/containers/Admin/NaicsCode/NaicsCodeTable.tsx
+++ b/app/containers/Admin/NaicsCode/NaicsCodeTable.tsx
@@ -16,11 +16,23 @@ export const NaicsCodeTableContainer: React.FunctionComponent<Props> = (
 ) => {
   const [validated, setValidated] = useState(false);
   const [showCreateModal, setShowCreateModal] = useState(false);
+  const [showActiveCodeError, setShowActiveCodeError] = useState(false);
   const {query} = props;
 
   const handleClose = () => {
     setShowCreateModal(false);
     setValidated(false);
+    setShowActiveCodeError(false);
+  };
+
+  const naicsCodeIsActive = (naicsCode) => {
+    let codeExists = false;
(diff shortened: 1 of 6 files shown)


app/containers/Admin/NaicsCode/NaicsCodeTable.tsx
```diff
@@ -16,15 +16,34 @@ export const NaicsCodeTableContainer: React.FunctionComponent<Props> = (
...
+  };
+
+  const naicsCodeIsActive = (naicsCode) => {
+    let codeExists = false;
+    query.allNaicsCodes.edges.forEach((edge) => {
+      if (edge.node.naicsCode === naicsCode) {
+        codeExists = true;
+      }
+    });
+    return codeExists;
```
pbastia, line 35: You can use `some(...)`, it returns on the first found occurrence instead of looping through the entire array

```suggestion
    return query.allNaicsCodes.edges.some((edge) => 
      edge.node.naicsCode === naicsCode
    );
```

pbastia (changes requested):

kriscooke (commented):
It's looking great! However while the below is no longer an issue, the red/green validation doesn't seem to be happening at all anymore 🤔 

> When a new code is added, on re-opening the New dialog again to add a subsequent one, the form validation styles should be cleared (not showing red/green).

pbastia (approved):

PR #1566, opened 2021-03-25
Add array support to diff visualization

matthieu-foucault:


app/hooks/useJsonSchemaDiff.ts
```diff
@@ -0,0 +1,92 @@
...
+    const keys = new Set<string>();
+
+    if (lhs) Object.keys(lhs).forEach((k) => keys.add(k));
+    if (rhs) Object.keys(rhs).forEach((k) => keys.add(k));
+
+    keys.forEach((k) =>
+      addObjectToDiffMap(`${path}_${k}`, lhs?.[k], rhs?.[k], idDiffMap)
+    );
+  }
+};
```
pbastia, line 18: Just a thought - I wonder if it would be useful to add a "max depth" parameter to avoid crash with circular references?

app/hooks/useJsonSchemaDiff.ts
```diff
@@ -0,0 +1,92 @@
...
+ * @returns An object with the formData to display, and the idDiffMap, containing the field-level diffs
+ */
+const useJsonSchemaDiff = (
+  formData: any,
+  showDiff: boolean,
+  formIdPrefix: string,
+  lhs?: any,
+  rhs?: any
+) => {
+  if (!showDiff) return {formData};
```
pbastia, line 36: I would handle the diffing toggle outside this function, make that the responsibility of the caller

app/hooks/useJsonSchemaDiff.ts
```diff
@@ -0,0 +1,92 @@
...
+            idDiffMap[`${arrayElementPath}_${key}`] = {
+              lhs: null,
+              rhs: difference.item.rhs[key]
+            };
+          });
+        } else if (difference.item.kind === 'D') {
+          const deletedItem = {};
+          // Deleted an element from the array
+          Object.keys(difference.item.lhs).forEach((key) => {
+            deletedItem[key] = null;
```
pbastia, line 63: `deletedItem` doesn't seem to be used anywhere?

app/hooks/useJsonSchemaDiff.ts
matthieu-foucault: Since the data comes from a JSON object, we can't have any circular reference in practice

app/hooks/useJsonSchemaDiff.ts
matthieu-foucault: That would break the first rule of hooks: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

Technically this doesn't need to be a hook though, since it doesn't use any state/effect

app/hooks/useJsonSchemaDiff.ts
matthieu-foucault: Indeed, that was from a previous implementation that didn't work. This injects `difference.item.lhs` below instead, so that the conditional schemas render properly

app/hooks/useJsonSchemaDiff.ts
pbastia: Ah, I didn't see that part about it being a hook.
Maybe the hook could just be the toggle - but the function could live in a library by itself?

pbastia (approved):

PR #1567, opened 2021-03-26
Feat: refresh swrs version

dleard:
updates the swrs (version 0) of an application if the corresponding swrs.report is newer.


schema/test/unit/swrs_functions/refresh_swrs_version_data_test.sql
```diff
@@ -0,0 +1,189 @@
...
+select is(
+  (select facility_name from ggircs_portal.ciip_admin where application_id=2 and version_number=0),
+  (select facility_name from ggircs_portal.get_swrs_facility_data(2, (select reporting_year from ggircs_portal.opened_reporting_year()))),
+  'Refresh function added facility data to the admin form for application with id = 2'
+);
+
+select is(
+  ((select updated_at from ggircs_portal.form_result where application_id=3 and version_number=0 and form_id=1) < now() - interval '12 hours'),
+  true::boolean,
+  'Application with id 3 was not updated by the refresh function because form_result.updated_by was after swrs.imported_at'
```
matthieu-foucault, line 130:
```suggestion
  'Application with id 3 was not updated by the refresh function because form_result.updated_at was after swrs.imported_at'
```
same typo below

matthieu-foucault (changes requested):
just a minor typo in the pg_tap test, looks good otherwise

matthieu-foucault (approved):

PR #1569, opened 2021-03-29
Feat: product naics associations UI

pbastia:


app/mutations/product_naics_code/updateProductNaicsCodeMutation.ts
```diff
@@ -0,0 +1,60 @@
...
+        ...optimisticProductNaicsCode
+      }
+    }
+  };
+
+  const updater = (store: RecordSourceSelectorProxy) => {
+    const record = store.get(naicsCodeId);
+    const connection = ConnectionHandler.getConnection(record, connectionKey);
+    ConnectionHandler.deleteNode(connection, variables.input.id);
+  };
```
matthieu-foucault, line 46: We're assuming that we're doing a delete here, which isn't necessarily the case. maybe this file should be renamed to `deleteProductNaicsCodeMutation`, and we can hardcode the `productNaicsCodePatch` to set the `deletedAt` variable

pbastia: That's a really good point, it would also make the API a lot clearer for the caller

kriscooke (changes requested):
One of the AC for [GGIRCS-2273](https://youtrack.button.is/issue/GGIRCS-2273) was to add a tile to the admin dash:

> Add a new tile to the Adminstrator dashboard, between "Reporting Operations" and "Users":

> ### Form Configuration
> Define and manage:

> NAICS codes
> Allowed products

kriscooke (commented):
Also needs a (passing) accessibility test added to `cypress/integration/accessibility/admin-all-pages-accessiblity.spec.js` to ensure we start from a baseline of accessiblity with new views.

app/components/Admin/NaicsCodeList.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  return (
+    <ListGroup variant="flush">
+      {naicsCodes.map(({code, description, id}) => (
+        <Link
+          key={id}
+          prefetch={false}
+          href={{pathname: router.pathname, query: {naicsCodeId: id}}}
+        >
+          <ListGroup.Item active={router.query?.naicsCodeId === id}>
```
kriscooke, line 21: Adding the `action` prop to `ListGroup.Item` ([docs example](https://react-bootstrap.github.io/components/list-group/#actionable-items)) improves the accessibility:
- makes the items render as `<button>` which enables interaction with them via keyboard focus navigation
- adds background hover style and `cursor: pointer;` to make it very clear that the list items are meant to be interacted with.

app/components/Admin/NaicsCodeList.tsx
```diff
@@ -0,0 +1,30 @@
...
+            <b>{code}</b>
+            <br />
+            <small>{description}</small>
```
kriscooke, line 24: Per the [design](https://youtrack.button.is/_persistent/Screen%20Shot%202021-03-22%20at%208.11.46%20AM.png?file=7-625&c=false&updated=1616427224019) could we swap the code number and description so the text label is above and in bold instead of the number?
I'd also recommend rendering the code below as regular body font-sized text instead of <small> because, while the user is likely scanning visually for the text, the code number still has similar importance to the description:

```
          <ListGroup.Item ...>
            <b>{description}</b>
            <br/>
            {code}
          </ListGroup.Item>
```

app/containers/Admin/ProductNaicsCode/AllowableProductsTableRow.tsx
```diff
@@ -0,0 +1,88 @@
...
+    await deleteProductNaicsCodeMutation(
+      environment,
+      naicsCodeId,
+      productNaicsCode,
+      'AllowableProducts_productNaicsCodesByNaicsCodeId'
+    );
+  };
+
+  const deleteModalContent = {
+    deleteName: 'allowed Product',
```
kriscooke, line 35:
```suggestion
    deleteName: 'allowed product',
```


Screen Shot 2021-04-06 at 10 22 52 AM

app/components/Admin/NaicsCodeList.tsx
```diff
@@ -0,0 +1,30 @@
+import Link from 'next/link';
```
kriscooke, line 1: Might it be better to name this file/component something more specific like `SelectableNaicsCodeList`? A few times I wondered if it was at all related or used in the list of NAICS codes at `/admin/naics-codes`.

app/containers/Products/LinkedProductModal.tsx
```diff
@@ -58,6 +58,8 @@ export const LinkedProductModalComponent: React.FunctionComponent<Props> = ({
   query,
   linkData
 }) => {
+  console.log(linkData);
+
```
kriscooke, line 62: Probably no need for this anymore :)

app/components/Admin/NaicsCodeSelectionComponent.tsx
```diff
@@ -0,0 +1,46 @@
...
+  naicsCodes: {code: string; description: string; id: string}[];
+}
+
+export const NaicsCodeSelectionComponent: React.FunctionComponent<Props> = ({
+  naicsCodes
+}) => {
+  const router = useRouter();
+
+  return (
+    <ListGroup variant="flush">
```
kriscooke, line 16: If you make the containing rectangle a height-constrained scrollable area, you'll want to ensure to set `tab-index="0"` so it can be interacted with using keyboard focus:

```
<div className="scrollable" tabIndex={0}>
```

I don't think you can target the scrollbar pseudoselectors using inline styles (for ex. on the Bootstrap ListGroup), so you may need to "line" it with a container div and make that the height-constrained scrollable area as I've shown.

app/components/Admin/NaicsCodeSelectionComponent.tsx
```diff
@@ -0,0 +1,63 @@
...
+              <ListGroup.Item action active={router.query?.naicsCodeId === id}>
+                {description}
+                <br />
+                <b>{code}</b>
+              </ListGroup.Item>
+            </Link>
+          ))}
+        </ListGroup>
+      </div>
+      <style jsx global>
```
matthieu-foucault, line 40: not all of the styles below need to be global, only `.list-group-item.active` (at least until we use a bootstrap theme)

app/containers/Admin/ProductNaicsCode/AllowableProductsSearch.tsx
```diff
@@ -0,0 +1,142 @@
...
+      .map((e) => {
+        return {
+          id: e.node.rowId,
+          name: e.node.productName
+        };
+      })
+      .filter((option) => !existingProductIds.includes(option.id));
+  }, [query]);
+
+  const [selectedProducts, setSelectedProducts] = useState([]);
```
matthieu-foucault, line 33: Can you select more than one product? If not this doesn't need to be an array

app/containers/Admin/ProductNaicsCode/AllowableProductsTable.tsx
```diff
@@ -0,0 +1,80 @@
...
+          {props.naicsCode.productNaicsCodesByNaicsCodeId.edges.map((e) => (
+            <AllowableProductsTableRow
+              naicsCodeId={props.naicsCode.id}
+              key={e.node.id}
+              productNaicsCode={e.node}
+            />
+          ))}
+        </tbody>
+      </Table>
+      <style jsx global>{`
```
matthieu-foucault, line 45: Those styles don't need to be global (if you remove the `.table` selector)

kriscooke (commented):
Looking great! Just a couple more things I was reminded of from our design:
- text links in top left to navigate to "Manage Industry (NAICS) Codes" and "Manage Products"
- your idea to show the list of allowed products on the right in alphabetical order.
  - it probably wouldn't hurt to do the same for the NAICS code descriptions on the left
  - if this is too much work, it's not that important

app/containers/Admin/ProductNaicsCode/ProductNaicsCodeAssociation.tsx
```diff
@@ -0,0 +1,107 @@
...
+          {!currentNaics && (
+            <Row>
+              <Col>
+                Please select a NAICS code on the left to add allowable products
+              </Col>
+            </Row>
+          )}
+        </Col>
+      </Row>
+      <style jsx global>{`
```
matthieu-foucault, line 74:
```suggestion
      <style jsx>{`
```

matthieu-foucault (changes requested):

app/components/Admin/NaicsCodeSelectionComponent.tsx
kriscooke: (you can have two `<style>` tags in a component - one global and one not 🤓)

app/containers/Admin/ProductNaicsCode/AllowableProductsSearch.tsx
```diff
@@ -0,0 +1,145 @@
...
+          ) {
+          edges {
+            node {
+              id
+              productId
+            }
+          }
+        }
+      }
+      allProducts(filter: {productState: {equalTo: PUBLISHED}}) {
```
matthieu-foucault, line 135: We should probably order this by `productName`

matthieu-foucault (changes requested):

matthieu-foucault (dismissed):
:shipit:

matthieu-foucault:
@kriscooke will need to approve it too

kriscooke (dismissed):
👌

matthieu-foucault (approved):

PR #1573, opened 2021-04-01
fix: application open/close overlap issue

dleard:


pbastia (approved):
We'll have to be aware of the side-effects on the non-2019 reporting years for this one

matthieu-foucault (approved):
This will do for now, as it is blocking our CI. I would like to add a parameter to that function in the future, so that we can choose which window to mock, and test the ciip incentive calculation for different reporting years

PR #1574, opened 2021-04-01
Simplified view of comments for analyst application review

kriscooke:
Note: The base for this PR is against #1538 rather than develop so we can handle review separately.
#1538 contains the necessary DB work for the new UI to build on here.

(The temporary "Click to toggle review comments" button will be removed along with the application header/title area update in the subsequent GGIRCS-2293.)

#### New components:
- `ReviewSidebar` lists comments for a given review step (until 2293, this just shows all the comments for the application).
- `ReviewComment` shows an individual timestamped comment in the list, with Resolve/Delete buttons

#### Incidental changes:

Added options to the default layout:
- `fixedHeader` allows individual pages to configure the header to be fixed-positioned (aka. "sticky"). This was necessary to support the layout of the fixed-position sidebar on the application review page.
- `disableHelpButton` allows disabling the automatic inclusion of the help button by individual pages to better manage their own layouts. This was necessary to let the application review page manage the display of the help button in combination with the interactivity of opening/closing the sidebar.
    * since there isn't a good/responsive CSS-based way to shift the help button when the sidebar is opened/closed, instead it is hidden while open, and shown closed once again when the sidebar is closed.

[GGIRCS-2193](https://youtrack.button.is/issue/GGIRCS-2193)


app/tests/unit/mutations/application_review_step/updateApplicationReviewStepMutation.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    } catch (error_) {
+      error = error_;
+    }
+
+    expect(error.message).toEqual(
+      'Variable "$input" got invalid value { id: "abc" }; Field applicationReviewStepPatch of required type ApplicationReviewStepPatch! was not provided.'
+    );
+  });
+  it('Should return isComplete(boolean) if valid', () => {
+    const test = tester.mock(mutation, {
```
kriscooke, line 56: I keep getting a failure with `Cannot query field "id" on type "UpdateApplicationReviewStepPayload".` here, which I don't understand.

app/tests/unit/mutations/application_review_step/updateApplicationReviewStepMutation.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  'utf8'
+);
+
+const mutation = `
+        mutation updateApplicationReviewStepMutation(
+          $input: UpdateApplicationReviewStepInput!
+        ) {
+          updateApplicationReviewStep(input: $input) {
+            id
+            isComplete
```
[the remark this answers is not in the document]
matthieu-foucault: Instead of copying the mutation (and not actually testing the application code), you can import it. It doesn't seem to be well documented, or typed with typescript, but this works:
```ts
import {mutation} from 'mutations/application_review_step/updateApplicationReviewStepMutation';
const mutationText = (mutation as any).default.params.text;
```

and then you use `mutationText` in the tests below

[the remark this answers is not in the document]
matthieu-foucault: Of course this requires exporting the `mutation` variable in `mutations/application_review_step/updateApplicationReviewStepMutation`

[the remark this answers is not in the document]
kriscooke: > Instead of copying the mutation (and not actually testing the application code), you can import it.

That does seem unwise, doesn't it! I'll admit I based the test on the one for another mutation since I wasn't too sure how to go about it - thanks for the tip, that sounds like a much better idea 😄 

We should probably also update our other mutation tests to do the same.

app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
```diff
@@ -0,0 +1,343 @@
...
+                    viewOnly={applicationReviewStep.isComplete}
+                    isResolved={resolved}
+                    onResolveToggle={resolveComment}
+                    onDelete={(id) => deleteComment(id, 'general')}
+                  />
+                );
+              }
+            )}
+          </ul>
+        )}
```
matthieu-foucault, line 167: Sonarcloud is complaining that this block is duplicated by the one below. It's a bit tricky because of the comment type in the `deleteComment` call. If we can't remove that parameter from `deleteComment`, we might need to write something like
```tsx
   const makeRenderReviewComment(commentType) => ({
                node: {
                  id,
                  description,
                  createdAt,
                  ciipUserByCreatedBy,
                  resolved
                }
              }) => {
                const showComment = showingResolved || !resolved;
                return !showComment ? null : (
                  <ReviewComment
                    key={id}
                    id={id}
                    description={description}
                    createdAt={createdAt}
                    createdBy={`${ciipUserByCreatedBy.firstName} ${ciipUserByCreatedBy.lastName}`}
                    viewOnly={applicationReviewStep.isComplete}
                    isResolved={resolved}
                    onResolveToggle={resolveComment}
                    onDelete={(id) => deleteComment(id, 'general')}
                  />
                );
              }
```
and use it like
```tsx
generalComments.map(makeRenderReviewComment('general'))
internalComments.map(makeRenderReviewComment('internal'))

```

kriscooke:
Currently this just reads the review step's name from the DB, but I think from a UI perspective we had been going to call deprecated review steps "Other" - as in, "Other Review".  Thoughts?

Screen Shot 2021-04-09 at 12 41 10 PM

app/mutations/application_review_step/deleteReviewCommentMutation.ts
```diff
@@ -12,16 +12,16 @@ import {
 import {ConnectionHandler} from 'relay-runtime';
 
 const mutation = graphql`
-  mutation deleteReviewCommentMutation(
-    $input: UpdateReviewCommentInput!
-    $applicationId: ID!
-    $version: String!
-  ) {
+  mutation deleteReviewCommentMutation($input: UpdateReviewCommentInput!) {
```
matthieu-foucault, line 15:
```suggestion
  mutation deleteReviewCommentMutation($input: DeleteReviewCommentInput!) {
```

app/pages/analyst/application-review.tsx
```diff
@@ -99,16 +113,48 @@ class ApplicationReview extends Component<Props> {
...
+                }}
+              >
+                <FontAwesomeIcon size="lg" icon={faArrowUp} />
+              </Button>
+            </OverlayTrigger>
+          </div>
+          {this.state.isSidebarOpened && (
+            <ReviewSidebar
+              applicationReviewStep={
+                query.application.applicationReviewStepsByApplicationId.edges[0]
```
kriscooke, line 150: This is where we set the currently selected review step to display in the sidebar. In future this will be set based on what they've selected in the review step menu, but for now it displays the first review step retrieved.

All existing applications' comments have been migrated in #1538 to a unique review step named "Legacy" review, and there are no comments yet referencing the new review steps - so I think that it's okay to temporarily set the displayed step in this way until the selector menu is implemented.

app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
```diff
@@ -0,0 +1,343 @@
...
+    await updateReviewCommentMutation(environment, variables);
+  };
+  const deleteComment = async (id, commentType) => {
+    const {environment} = relay;
+    const variables = {
+      input: {
+        id,
+        reviewCommentPatch: {
+          deletedAt: nowMoment().format('YYYY-MM-DDTHH:mm:ss')
+        }
```
matthieu-foucault, line 87: I don't think that this patch is needed. There is an issue with `deleteReviewCommentMutation` where the input type is wrong. The only input should be the `id`

app/mutations/application_review_step/deleteReviewCommentMutation.ts
```diff
@@ -12,16 +12,16 @@ import {
...
-          ...ApplicationDetailsContainer_application
-            @arguments(version: $version)
+      reviewComment {
+        id
+        description
+        createdAt
+        resolved
+        ciipUserByCreatedBy {
+          firstName
+          lastName
```
matthieu-foucault, line 24: Is all of this needed, or just the `id`?

matthieu-foucault (approved):

app/components/Admin/ReviewComment.tsx
```diff
@@ -0,0 +1,103 @@
...
+              <Tooltip id="delete" {...props}>
+                Delete comment
+              </Tooltip>
+            )}
+          >
+            <Button
+              size="sm"
+              aria-label="Delete"
+              variant="secondary"
+              onClick={() => onDelete(id)}
```
matthieu-foucault, line 59: Do we want to show a confirmation modal here? `DeleteConfirmationModal` could be reused (with a bit of tweaking)

kriscooke: The display of comment list items is borrowed from the previous design that didn't use that pattern, and comment deletion didn't really come up as an important point in our design process + feedback from our SME. However - requested by her, there **will** be one shown if you try to mark a review completed when there are outstanding unresolved comments.

All of the review step-related work was supposed to come in as part of [GGIRCS-2293](https://youtrack.button.is/issue/GGIRCS-2293) next up, so if necessary that may be an opportunity to do the same for comment deletion.

app/mutations/application_review_step/deleteReviewCommentMutation.ts
kriscooke: Just realized there's also no test for this mutation

kriscooke: About the suggested snippet - if I do that, I get
```
ERROR:
Variable '$input' was defined as type 'DeleteReviewCommentInput!' but used in a location expecting the type 'UpdateReviewCommentInput!'
```

Because ultimately it's an updateReviewComment mutation

app/mutations/application_review_step/deleteReviewCommentMutation.ts
kriscooke: I don't think it is.

app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
kriscooke: > There is an issue with deleteReviewCommentMutation where the input type is wrong.

Could you elaborate?

app/mutations/application_review_step/deleteReviewCommentMutation.ts
matthieu-foucault: Right, the line below, nevermind then

app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
matthieu-foucault: I was referring to my comment in `deleteReviewCommentMutation`, but it turns out I was wrong, so nevermind :)

PR #1575, opened 2021-04-02
feat: update view-application page to use next dynamic routes

matthieu-foucault:


app/cypress/integration/reporter-access-all-pages.spec.js
```diff
@@ -33,7 +33,19 @@ describe('When logged in as a reporter', () => {
...
     cy.get('.admin');
+
+    // test permanent redirection from the v1 view-application page
+    cy.visit(
+      `/reporter/view-application?applicationId=${applicationId}&version=1`
+    );
+    cy.url().should(
+      'include',
+      `/reporter/application/${applicationId}/version/1/view`
+    );
```
pbastia, line 49: We could add a `cy.get('#page-content');` in between to force cypress to wait and avoid timing issues, not sure how much of an issue this is in this case

docs/all-views-inventory.md
```diff
@@ -127,65 +127,15 @@ Interstitial page shown after submitting an application, which confirms the appl
...
-View requests made to certify applications addressed to the user's email address, including links to view the individual application details. List is searchable and sortable by facility, operator, request status, certifier and certification date; certifiers with many requests can select multiple requests to show the signing interface and certify all selected applications at once.
-
-Once certified, the requests do not disappear, serving as a historical record of applications the user has certified.
-
-#### Shared components:
-
-- `CertificationSignature`
-- `SignatureDisclaimerCard`
-  - `LegalDisclaimerText`
-
```
pbastia, line 188: nice catch on all the certification docs

pbastia (approved):
Love it! So much removed code!

app/cypress/integration/reporter-access-all-pages.spec.js
matthieu-foucault: I wouldn't expect any issue here. The assertion has a timeout, and our server should redirect to the new URL quickly enough

PR #1580, opened 2021-04-07
Feat: filter products by naics code

dleard:
- [x] computed column to return naicsCode on applicationRevision
- [x] update ProductRowIdField to filter choices by naics code
- [x] update ProductField to show a warning when the product in the application is not associated with the reported naics code
- [x] pgTap tests
- [x] jest unit tests
- [x] add dev data for product_naics_code table (or else we won't be able to add any products to applications)
- [x] update e2e tests broken by product-naics relationship requirement
- [x] refactor after #1584 is merged

PR #1581, opened 2021-04-08
Feat: removing flared vented ng

pbastia:
Removing 'Flared Natural Gas' and 'Vented Natural Gas' as options for fuel.
- [x] update the `FuelField`  to render an alert if the fuel is archived
- [x] update the `FuelRowIdField` to still render the fuel name when the fuel is deprecated
- [x] unit tests for FuelField/FuelRowIdField
- [x] updates to e2e tests if necessary


schema/data/prod/fuel.sql
```diff
@@ -92,9 +92,9 @@ values
   -- Flared Natural Gas was previously named Flared Natural Gas CH4
   -- It was renamed and the swrs_fuel_mapping_id was updated as applicants for the CIIP 2019 applications
   -- were instructed to report an amount of natural gas under this fuel
-  (86,'Flared Natural Gas','m3', 'active', 95),
+  (86,'Flared Natural Gas','m3', 'archived', 95),
```
matthieu-foucault, line 95: I am thinking that we should add a `comments` column to the `fuel` table, to describe why those fuels are archived (and that natural gas should be used instead). The comment above for Flared Natural Gas CH4 would also go in that column.
This wouldn't be intended to be displayed, but as a metadata so that anyone looking at this table in the database can see what's up with those archived fuels.

app/containers/Forms/FuelField.tsx
```diff
@@ -26,17 +33,32 @@ const FuelField: React.FunctionComponent<Props> = (props) => {
...
   };
 
-  return <ObjectField {...props} onChange={handleChange} />;
+  return (
+    <>
+      {isFuelArchived && (
+        <Alert variant="danger">
+          <strong>Warning:</strong> This fuel has been archived. Please remove
+          it and select an appropriate alternative.
+        </Alert>
```
matthieu-foucault, line 42: Will this alert be displayed to the reviewer when looking at an existing application? If so it might be confusing when they look at 2019 applications?

app/containers/Forms/ProductField.tsx
```diff
@@ -27,14 +27,13 @@ export const ProductFieldComponent: React.FunctionComponent<Props> = (
...
 
-  // TODO: Clean up this function with array.some() as noted in https://github.com/bcgov/cas-ciip-portal/pull/621
-  const productIsPublished = (formData, query) => {
-    const product = query.allProducts.edges.find(
-      ({node}) => node.rowId === formData.productRowId
-    )?.node;
-    if (product?.productState === 'PUBLISHED' || !product) return true;
-    return false;
-  };
+  const productIsPublished =
```
dleard, line 30: Can we undo these changes? This file is going through major changes in #1580 & this is going to cause conflicts

app/containers/Forms/ProductRowIdField.tsx
```diff
@@ -76,15 +76,11 @@ export const ProductRowIdFieldComponent: React.FunctionComponent<Props> = (
...
-    <div>
-      <InputGroup className="mb-3">
-        <InputGroup.Prepend>
-          <InputGroup.Text>
-            {allProducts.archivedProductNames[archivedIndex]}
-          </InputGroup.Text>
-        </InputGroup.Prepend>
-      </InputGroup>
-    </div>
+    <InputGroup>
```
dleard, line 79: ditto undo changes as this file also changeds a lot in #1580

app/containers/Forms/FuelField.tsx
pbastia: Yes, they show up on the summary page as well, I replicated the behaviour and look that was used for archived products. It could be a low impact improvement to clean this up, maybe a tech debt card?

schema/verify/tables/fuel_001.sql
```diff
@@ -0,0 +1,11 @@
+-- Verify ggircs-portal:tables/fuel_001 on pg
+
+BEGIN;
```
dleard, line 3: nitpicking here...toLowerCase()

app/containers/Forms/FuelRowIdField.tsx
```diff
@@ -17,26 +18,63 @@ export const FuelRowIdFieldComponent: React.FunctionComponent<Props> = (
...
+      </InputGroup.Text>
+    </InputGroup>
+  );
 };
 
 export default createFragmentContainer(FuelRowIdFieldComponent, {
   query: graphql`
     fragment FuelRowIdField_query on Query {
-      allFuels(condition: {state: "active"}) {
+      activeFuels: allFuels(condition: {state: "active"}) {
```
dleard, line 69: going forward do we want to use `filter` rather than `condition` for consistency? (This is a question, not necessarily a requested change). I've been replacing condition with filter when refactoring things.
Also: 
```suggestion
      activeFuels: allFuels(condition: {state: "active"} orderBy: NAME_ASC) {
```

dleard (changes requested):

app/containers/Forms/FuelField.tsx
matthieu-foucault: It can be cleaned up later, but it's not tech tebt, it's a feature

app/containers/Forms/FuelRowIdField.tsx
pbastia: Happy to indulge! I doubt there is any performance concern between using the postgraphile filter plugin and the default graphQL syntax

dleard (approved):
Awesome! :shipit:

matthieu-foucault (approved):

PR #1586, opened 2021-04-09
chore: update cypress

matthieu-foucault:


matthieu-foucault:
Cypress 7 is having significant performance issues (https://github.com/cypress-io/cypress/issues/15853), downgrading to 6

matthieu-foucault:
Also https://github.com/cypress-io/cypress/issues/15779, so maybe 6.6 is the best option for now

matthieu-foucault:
v6.9.1 seemed to be good after all. I had to fix some tests in `override-justification.spec.js` which should never have passed, but somehow did

pbastia (dismissed):

app/cypress/support/commands.js
```diff
@@ -34,8 +34,8 @@ Cypress.Commands.add('login', (username, password) => {
 });
 
 Cypress.Commands.add('logout', () => {
-  cy.clearCookie('mocks.auth');
-  cy.request('/logout');
+  cy.get('header').contains('Logout').click();
+  cy.url().should('equal', Cypress.config().baseUrl + '/');
```
pbastia, line 38: shouldn't we still clear the cookie?

matthieu-foucault: no, `cy.logout` is only used when using the real login. Cypress will clear its own cookies automatically when using the mock login

pbastia (dismissed):

pbastia (approved):

PR #1589, opened 2021-04-12
Feat/header userinfo

naomiaro:
https://youtrack.button.is/issue/GGIRCS-2276

> As a reporter, I want to see my name in the header so that I know which account I'm logged into.

LOGGED IN

Mobile View:
Screenshot 2021-04-19 at 10 53 08 AM

Desktop View:
Screenshot 2021-04-19 at 10 53 38 AM

NOT LOGGED IN

Mobile View:
Screenshot 2021-04-19 at 10 56 13 AM

Desktop View:
Screenshot 2021-04-12 at 1 10 53 PM


app/layouts/default-layout.tsx
```diff
@@ -197,6 +198,9 @@ export default createFragmentContainer(DefaultLayout, {
     fragment defaultLayout_session on JwtToken {
       ciipUserBySub {
         __typename
+        firstName
+        lastName
+        emailAddress
```
matthieu-foucault, line 203: This will generate a warning, as Relay expects values to be used in the same file as the query.

app/components/Layout/Header.tsx
```diff
@@ -80,19 +92,64 @@ const HeaderLayout = ({
...
+                                {user.emailAddress}
+                              </div>
+                            </a>
+                          </Dropdown.Item>
+                          <Dropdown.Item
+                            as={Form}
+                            action="/logout"
+                            method="post"
+                            tabIndex={0}
+                            onKeyPress={submitForm}
```
kriscooke, line 122: The HTML form appears to submit fine for me without an imperative handler - was there some issue with form submission?

app/components/Layout/Header.tsx
```diff
@@ -80,19 +92,64 @@ const HeaderLayout = ({
...
+                      </Link>
+                    </li>
+                    <li className="d-lg-none">
+                      <Form action="/logout" method="post">
+                        <button type="submit" className="nav-button text-right">
+                          Logout
+                        </button>
+                      </Form>
+                    </li>
+                  </>
```
matthieu-foucault, line 151: I would extract all of this in its own container, which should be a relay `fragmentContainer`, so that the relay fragment fetching the user info is colocated with its rendering, preventing the warning I mentioned in my other comment.

Instead of a `user` prop, the `Header` component would have a `userProfileDropdown` prop, and the `UserProfileDropdown` component would be used in `default-layout`.

matthieu-foucault (changes requested):

app/components/Layout/Header.tsx
naomiaro: I was trying keyboard navigation on the new dropdown, but it didn't submit without this.

app/components/Layout/Header.tsx
naomiaro: Ah ok, yes I need to get used to Relay. I don't think I noticed a warning tbh. So you're saying here that `userProfileDropdown` is a prop that would pass down a component instead? I'll try that out...

matthieu-foucault: That warning may be at the very top of the file, as there is (was? It seemed to have been fixed with the PR that updated relay things) a bug with the linter

app/components/Layout/Header.tsx
```diff
@@ -180,6 +238,11 @@ const HeaderLayout = ({
               background-color 0.15s ease-in-out, border-color 0.15s ease-in-out,
               box-shadow 0.15s ease-in-out;
           }
+          #menu-toggle.logged-in {
+            background: #f8f9fa;
+            border-radius: 100%;
+            padding: 0.375rem 0.7rem;
```
kriscooke, line 244: If you set an explicit 40px height and width on this, the circular background would no longer be squished and occupy the same width as the hamburger icon 🙂 (to complement, the icon's size can be fine-tuned to fit well within by setting its `style` prop inline instead of the `size` prop)

Screen Shot 2021-04-12 at 2 56 50 PM

app/components/Layout/Header.tsx
```diff
@@ -80,19 +92,64 @@ const HeaderLayout = ({
...
-                  </li>
+                  <>
+                    <li className="d-none d-lg-block">
+                      <Dropdown alignRight>
+                        <Dropdown.Toggle variant="primary">
+                          <FontAwesomeIcon
+                            color="white"
+                            icon={faUser}
+                            size="1x"
+                          />
```
kriscooke, line 103: Could we find a way to have the user icon button appear the same (circular background, inverse colors - dropdown caret would be the only difference) as on narrower viewports?

kriscooke (changes requested):

app/tests/unit/components/Layout/__snapshots__/Header.test.tsx.snap
```diff
@@ -2,48 +2,48 @@
 
 exports[`matches the last accepted Snapshot 1`] = `
 <header
-  className="jsx-2785271651 "
+  className="jsx-2264600173 "
```
kriscooke, line 5: We're trying to improve our unit testing to have more functional assertions, as compared to mainly snapshots.

Since this work is turning the Header into more of a container that interacts with data (the user object), it might be a good time to add some functional assertions based on whether or not the user is logged in.

pbastia (dismissed):
just some extra brackets we can remove.
Really good looking menu! I love it it!

app/components/Layout/Header.tsx
```diff
@@ -78,22 +87,7 @@ const HeaderLayout = ({
...
-                )}
-                <li>
-                  <Form action="/logout" method="post">
-                    <button type="submit" className="nav-button">
-                      Logout
-                    </button>
-                  </Form>
-                </li>
-              </>
+              <>{isRegistered && <>{userProfileDropdown}</>}</>
```
pbastia, line 90:
```suggestion
              isRegistered && userProfileDropdown
```

app/components/Layout/Header.tsx
```diff
@@ -1,16 +1,16 @@
...
 
 const DESKTOP_BREAKPOINT_QUERY = '(min-width: 992px)';
 
 const HeaderLayout = ({
   isLoggedIn = false,
   isRegistered = false,
   fixed = false,
+  userProfileDropdown = null,
   children
 }) => {
```
matthieu-foucault, line 15: Look like this component is missing some typed Props 
```suggestion
interface Props {
  isLoggedIn?: boolean;
  isRegistered?: boolean;
  fixed?: boolean;
  userProfileDropdown?: React.ReactNode;
}
const HeaderLayout: React.FunctionComponent<Props> = ({
  isLoggedIn = false,
  isRegistered = false,
  fixed = false,
  userProfileDropdown,
  children
}) => {
```

app/components/Layout/Header.tsx
```diff
@@ -237,6 +216,27 @@ const HeaderLayout = ({
           }
         `}
       </style>
+      <style jsx global>{`
```
matthieu-foucault, line 219: Instead of creating a global style, these should be moved to the `UserProfileDropdown`'s styles

app/layouts/default-layout.tsx
```diff
@@ -4,6 +4,7 @@ import {graphql, createFragmentContainer} from 'react-relay';
 import {defaultLayout_session} from '__generated__/defaultLayout_session.graphql';
 import getConfig from 'next/config';
 import Header from 'components/Layout/Header';
+import UserProfileDropdownContainer from 'containers/User/UserProfileDropdown';
```
matthieu-foucault, line 7:
```suggestion
import UserProfileDropdown from 'containers/User/UserProfileDropdown';
```

app/layouts/default-layout.tsx
```diff
@@ -49,6 +50,11 @@ const DefaultLayout: React.FunctionComponent<Props> = ({
         fixed={fixedHeader}
         isLoggedIn={Boolean(session)}
         isRegistered={Boolean(session?.ciipUserBySub)}
+        userProfileDropdown={
+          <UserProfileDropdownContainer
```
matthieu-foucault, line 54:
```suggestion
          <UserProfileDropdown
```

app/tests/unit/containers/User/UserProfileDropdown.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+import {shallow} from 'enzyme';
+import {UserProfileDropdownComponent} from 'containers/User/UserProfileDropdown';
+
+// https://github.com/facebook/relay/issues/2394#issuecomment-379590645
+const mockRefType: any = null;
+const user = {
+  firstName: 'Hamza',
+  lastName: 'Javed',
+  emailAddress: 'hamza@button.is',
+  ' $refType': mockRefType
```
matthieu-foucault, line 11:
```suggestion
  ' $refType': 'UserProfileDropdown_user'
```

this should just be the name of the relay fragment/query

app/tests/unit/layout/default-layout.test.tsx
```diff
@@ -3,8 +3,29 @@ import {shallow} from 'enzyme';
...
   });
+
+  it('when logged in matches the last accepted Snapshot', () => {
+    const mockRefType: any = null;
+    const user = {
+      __typename: 'qq',
+      firstName: 'Test',
+      lastName: 'Tester',
+      emailAddress: 'test-tester@testing.com',
+      ' $fragmentRefs': mockRefType
```
matthieu-foucault, line 18:
```suggestion
      ' $fragmentRefs': {
        UserProfileDropdown_user: true
      }
```
This is just listing the fragments that are in the query. Its only purpose when unit testing is to make typescript happy (same as `' $refType'`)

app/layouts/default-layout.tsx
```diff
@@ -196,6 +202,7 @@ export default createFragmentContainer(DefaultLayout, {
     fragment defaultLayout_session on JwtToken {
       ciipUserBySub {
         __typename
```
matthieu-foucault, line 204: You don't need `__typename` anymore. Its only purpose is when you don't actually need any field in a graphql node, but just want to check its existence. Now that you have a fragment in this node you can remove it.
```suggestion
```

app/tests/unit/containers/User/UserProfileDropdown.test.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import {shallow} from 'enzyme';
+import {UserProfileDropdownComponent} from 'containers/User/UserProfileDropdown';
+
+// https://github.com/facebook/relay/issues/2394#issuecomment-379590645
+const mockRefType: any = null;
+const user = {
+  firstName: 'Hamza',
+  lastName: 'Javed',
+  emailAddress: 'hamza@button.is',
```
matthieu-foucault, line 10: Feel free to use a fake name for the tests

matthieu-foucault (changes requested):
If you want to write unit tests for the behavior of `UserProfileDropdown`, you might need to use `mount` instead of `shallow`, so that you can open the dropdown.

app/components/Layout/Header.tsx
kriscooke: I can probably offer some CSS suggestions later on to help trim this down to streamline the dropdown toggle's containing box and equate it with the mobile hamburger button size.

Screen Shot 2021-04-15 at 10 48 14 AM

app/tests/unit/layout/default-layout.test.tsx
naomiaro: Type '{ firstName: string; lastName: string; emailAddress: string; ' $fragmentRefs': { UserProfileDropdown_user: boolean; }; }' is not assignable to type '{ readonly " $fragmentRefs": FragmentRefs<"UserProfileDropdown_user">; }'.
  The types of '" $fragmentRefs".UserProfileDropdown_user' are incompatible between these types.
    Type 'boolean' is not assignable to type 'true'.

I was getting this before, not sure why it's complaining for this component since I saw others similar to this.

app/layouts/default-layout.tsx
naomiaro: Cool thanks for mentioning what it was doing!

app/components/Layout/Header.tsx
naomiaro: It's used in both components because Dashboard button is still in here. Is there another way to avoid copying all the styles?

app/tests/unit/layout/default-layout.test.tsx
naomiaro: just having to cast everything with `as` in the end for these tests, seems like other areas of the codebase are doing that as well.

app/tests/unit/layout/default-layout.test.tsx
```diff
@@ -1,10 +1,35 @@
...
 
 describe('The DefaultLayout component', () => {
-  it('matches the last accepted Snapshot', () => {
+  it('when logged out matches the last accepted Snapshot', () => {
     const wrapper = shallow(<DefaultLayoutComponent session={null} />);
     expect(wrapper).toMatchSnapshot();
   });
+
+  it('when logged in matches the last accepted Snapshot', () => {
+    const session = {
```
matthieu-foucault, line 13:
```suggestion
    const session: defaultLayout_session = {
```

This is a better option than casting. This way if you have an issue with the query, the correct line will be highlighted.

app/components/Layout/Header.tsx
matthieu-foucault: The `css` template method should allow you to have reusable css: https://github.com/vercel/styled-jsx#styles-outside-of-components

app/tests/unit/containers/User/UserProfileDropdown.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+    );
+    expect(render).toMatchSnapshot();
+  });
+
+  it('matches desktop snapshot', () => {
+    render = shallow(
+      <UserProfileDropdownComponent user={user as UserProfileDropdown_user} />
+    );
+    expect(render).toMatchSnapshot();
+  });
```
matthieu-foucault, line 26: I don't see the difference between those two tests. If you're looking for visual differences, they will be in the happo screenshots

matthieu-foucault (changes requested):

app/tests/unit/layout/default-layout.test.tsx
naomiaro: yeah I get typescript errors just doing this as well, so this object needs to be changed, but I'm not quite sure how.

app/tests/unit/containers/User/UserProfileDropdown.test.tsx
naomiaro: yeah forgot I had this in there, was harder to change the jsdom window size than I thought. A better test for cypress I guess.

app/tests/unit/layout/default-layout.test.tsx
```diff
@@ -1,10 +1,40 @@
...
+
+    const userProfileDropdown: ReactElement = wrapper
+      .find('HeaderLayout')
+      .prop('userProfileDropdown');
+
+    expect(userProfileDropdown.props.user).toEqual({
+      ' $fragmentRefs': {
+        UserProfileDropdown_user: true
+      }
+    });
```
matthieu-foucault, line 38: You can compare object references here
```suggestion
    expect(userProfileDropdown.props.user).toBe(session.ciipUserBySub);
```

app/components/Layout/Header.tsx
kriscooke: We have an example of that in `containers/Forms/FormSharedStyles`

matthieu-foucault:
@naomiaro you will likely see some e2e test failure, because of a change that got recently merged here: 
https://github.com/bcgov/cas-ciip-portal/blob/a2fd0cab5c6286f1a6e3bde3909f526d0f2ce024/app/cypress/support/commands.js#L37

This will likely need to be changed to:
```js
cy.get('header .user-icon').click();
cy.get('header').contains('Logout').click(); 
```

app/components/Layout/Header.tsx
naomiaro: Looks like I might be able to use resolve here since it's a scoped parent/child thing https://github.com/vercel/styled-jsx#the-resolve-tag

app/components/Layout/Header.tsx
naomiaro: Bootstrap menus seem to be calculated internally to a position that just looks slightly off:

Screenshot 2021-04-16 at 6 19 37 PM

On mobile the hamburger and user icon are in a button with the same height and width. Was there something else needed there?

naomiaro: I can remove padding from the desktop dropdown if you want, the hamburger icon will never be shown in the menu like this.

kriscooke (commented):
You might have noticed some weird behavior in the mobile emulator where it "expands" the page width when the dropdown opens. That seemed to only happen in the emulator, but not if you narrowed the actual window.

However it does seem to be real when there is content overflow in the dropdown (eg: if name or email text is long - inspired by one of your tests). It appears cut off, but in reality has expanded the page width and you'd have to scroll right and read white-on-white text to see the rest of it. 

As solution, I don't think it would be unreasonable to use `text-overflow: ellipsis` and just cut off excessively long names/emails because the user will still be able to see enough to know what email they're logged in with.

Screen Shot 2021-04-19 at 11 59 40 AM

app/containers/User/UserProfileDropdown.tsx
```diff
@@ -0,0 +1,115 @@
...
+              }}
+            />
+          </Dropdown.Toggle>
+          <Dropdown.Menu>
+            <Dropdown.Item as={Link} href="/user/profile">
+              <a className="dropdown-item text-right">
+                <div>
+                  <span>{user.firstName}</span> <span>{user.lastName}</span>
+                </div>
+                <div className="small text-muted text-nowrap">
```
kriscooke, line 57: Removing `.text-nowrap` is an easy fix to the "long email" issue as it just permits the text to wrap to the next line without creating overflow:
```suggestion
                <div className="small text-muted">
```

naomiaro: OK let's let it wrap if the emails are really long for now.

kriscooke (approved):

matthieu-foucault (approved):

kriscooke:
🎉

PR #1592, opened 2021-04-12
Chore: remove certifier mentions in code

pbastia:
Removing items in code related to the certifier role:
- testing and test helpers
- e2e fixtures


kriscooke (changes requested):
Since all of the UI copy where you've identified certifier mentions is text that had to be written, reviewed and approved by both the business area and by GCPE, I'm not comfortable with us editing it ad-hoc until we receive their updated copy. In particular, some of it it is regulatory language referring to definitions found in legislation that's independent of this app.  

They're currently revising these texts, and may even find it helpful to refer to the old text in-app during that process. Everything else looks great - it should be fine if you interactively rebase to just remove 74b4dac.

kriscooke (approved):

kriscooke (approved):

PR #1594, opened 2021-04-13
Feat: allow deprecation of emission categories

pbastia:
Deprecating the 'other, non carbon-taxed' emission category

This feature will make sure that any emission category with the 'deleted_at' field set will render as read-only in existing applications, and won't be shown as an option for new applications.

### DIFF
--- a/schema/test/unit/tables/ciip_user_organisation_test.sql
+++ b/schema/test/unit/tables/ciip_user_organisation_test.sql
@@ -126,7 +126,7 @@ select results_eq(
   $$
     select count(*) from ggircs_portal.ciip_user_organisation
   $$,
-  ARRAY[5::bigint],
+  ARRAY[8::bigint],
   'Analyst can select all from table ciip_user_organisation'
 );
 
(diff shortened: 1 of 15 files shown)


matthieu-foucault (commented):
can you add submitted applications in our dev data with deleted fuel/emission categories, for easier testing/PO approval?

schema/test/unit/tables/ciip_user_organisation_test.sql
```diff
@@ -126,7 +126,7 @@ select results_eq(
   $$
     select count(*) from ggircs_portal.ciip_user_organisation
   $$,
-  ARRAY[5::bigint],
+  ARRAY[8::bigint],
```
dleard, line 129: Not a change request, just some useful info for future tests: Using ARRAY[...] in these results_eq() assertions was giving me some trouble when the array of results contained different types (not an issue here). I've recently started using values(...) instead of ARRAY[...] which appears to work better.

dleard (approved):

PR #1606, opened 2021-04-16
Review step selector on application review page

kriscooke:
- Review step selector shows selected review step in sidebar
- Shows completion state of each step
   * marking a step complete shows a confirmation modal when there remain unresolved comments, requested by SME
- "Make a decision or request changes" button remains disabled until all steps are marked completed
    * once a decision is made, text/color/icon indicates the decision
- "Change decision" button preserves current functionality for admins (but not analysts, per SME) to change the decision
    * actual making/changing of decision will be implemented in [GGIRCS-2294](https://youtrack.button.is/issue/GGIRCS-2294) when the status dropdown will finally be removed.
    * disabled when a newer draft exists
    * instead, disabled button indicates the existence of newer draft
- Once a decision is made, review becomes view-only (reversible if the decision is reversed)
    * disables marking steps incomplete, creation/modification of comments
- Minor accessibility fixes for the page
- Implements a reusable `GenericConfirmationModal` for use in patterns requiring confirmation to continue

[GGIRCS-2293](https://youtrack.button.is/issue/GGIRCS-2293)

Screen Shot 2021-04-23 at 8 00 31 AM
Screen Shot 2021-04-23 at 8 01 14 AM
Screen Shot 2021-04-23 at 7 59 40 AM
Screen Shot 2021-04-23 at 8 05 38 AM

### DIFF
--- a/app/components/GenericConfirmationModal.tsx
+++ b/app/components/GenericConfirmationModal.tsx
@@ -1,0 +1,82 @@
+import React from 'react';
+import {Modal, Button} from 'react-bootstrap';
+
+interface Props {
+  show: boolean;
+  title: string;
+  onConfirm: () => void;
+  onCancel?: () => void;
+  confirmButtonText?: string;
+  cancelButtonText?: string;
+  confirmButtonVariant?: string;
+  cancelButtonVariant?: string;
+  size?: 'sm' | 'lg' | 'xl';
+}
+
+export const GenericConfirmationModal: React.FunctionComponent<Props> = ({
+  children,
+  show,
+  title,
+  onConfirm,
+  onCancel,
+  confirmButtonVariant,
+  confirmButtonText = 'Confirm',
+  cancelButtonVariant = 'secondary',
+  cancelButtonText = 'Cancel',
+  size
+}) => {
+  return (
+    <Modal onHide={onCancel} show={show} size={size} centered>
+      <Modal.Header>
+        <h2>{title}</h2>
+      </Modal.Header>
+      <Modal.Body>{children}</Modal.Body>
+      <Modal.Footer>
+        <Button id="cancel" onClick={onCancel} variant={cancelButtonVariant}>
+          {cancelButtonText}
+        </Button>
+        <Button id="confirm" onClick={onConfirm} variant={confirmButtonVariant}>
+          {confirmButtonText}
+        </Button>
+      </Modal.Footer>
+      <style jsx>{`
+        h2 {
+          margin-bottom: 0;
+        }
+      `}</style>
+      <style jsx global>{`
+        .modal-header {
+          background: #036;
+          color: #fff;
+        }
+        .btn-primary#confirm {
+          background: #036;
+        }
+        .btn-primary#confirm:hover,
+        .btn-primary#cancel:hover {
+          background: #002040;
+          border-color: #002040;
+        }
+        .btn-outline-primary#confirm,
+        .btn-outline-primary#cancel {
+          color: #036;
+          border-color: #036;
+        }
+        .btn-outline-primary#confirm:hover,
+        .btn-outline-primary#cancel:hover {
+          color: #fff;
+          background: #002040;
+        }
+        .btn-success#confirm {
+          background: #24883e;
+        }
+        .btn-outline-success#confirm {
+          color: #036;
+          border-color: #036;
+        }
+      `}</style>
+    </Modal>
+  );
+};
+
+export default GenericConfirmationModal;
--- a/app/components/GenericConfirmationModal.tsx
+++ b/app/components/GenericConfirmationModal.tsx
@@ -67,13 +67,6 @@ export const GenericConfirmationModal: React.FunctionComponent<Props> = ({
           color: #fff;
           background: #002040;
         }
-        .btn-success#confirm {
-          background: #24883e;
-        }
-        .btn-outline-success#confirm {
-          color: #036;
-          border-color: #036;
-        }
       `}</style>
     </Modal>
   );
--- a/app/components/GenericConfirmationModal.tsx
+++ b/app/components/GenericConfirmationModal.tsx
@@ -5,7 +5,7 @@ interface Props {
   show: boolean;
   title: string;
   onConfirm: () => void;
-  onCancel?: () => void;
+  onCancel: () => void;
   confirmButtonText?: string;
   cancelButtonText?: string;
   confirmButtonVariant?: string;
--- a/app/containers/Admin/ApplicationReview/ApplicationReviewStepSelector.tsx
+++ b/app/containers/Admin/ApplicationReview/ApplicationReviewStepSelector.tsx
@@ -21,7 +21,7 @@ export const ApplicationReviewStepSelector: React.FunctionComponent<Props> = ({
   return (
     <Row>
       <Col md={5}>
-        <ListGroup as="ul" role="listbox">
+        <ListGroup as="ul">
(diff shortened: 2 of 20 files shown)


app/components/GenericConfirmationModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+          color: #fff;
+          background: #002040;
+        }
+        .btn-success#confirm {
+          background: #24883e;
+        }
+        .btn-outline-success#confirm {
+          color: #036;
+          border-color: #036;
+        }
```
kriscooke, line 76: I want to pull these overrides out into the default layout

app/containers/Applications/ApplicationRevisionStatusContainer.tsx
```diff
@@ -103,78 +103,59 @@ export const ApplicationRevisionStatusComponent: React.FunctionComponent<Props>
 
   return (
     <>
-      <Row>
-        <Col md={3}>
```
kriscooke, line 107: Eventually this component will be removed once "make a decision" is implemented in GGIRCS-2294. Until then, this just removes the `Row` and `Col` layout wrappers to make it inline with the page heading.

app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
```diff
@@ -10,53 +10,75 @@ import deleteReviewCommentMutation from 'mutations/application_review_step/delet
...
+  );
   const noGeneralCommentsToShow =
-    generalComments.length === 0 ||
-    (!showingResolved && generalComments.every((c) => c.node.resolved));
+    generalComments?.length === 0 ||
+    (!showingResolved && allGeneralCommentsAreResolved);
   const noInternalCommentsToShow =
-    internalComments.length === 0 ||
-    (!showingResolved && internalComments.every((c) => c.node.resolved));
+    internalComments?.length === 0 ||
```
kriscooke, line 52: I added a bunch of these optional `?` indicators during an intermediate stage, but I think I can probably remove them
```suggestion
    internalComments.length === 0 ||
```

app/containers/Incentives/IncentiveCalculatorContainer.tsx
```diff
@@ -17,7 +17,7 @@ export const IncentiveCalculator: React.FunctionComponent<Props> = ({
     <>
       <Jumbotron>
         <div style={{marginBottom: '30px'}}>
-          <h5>Incentive by Product:</h5>
+          <h2>Incentive by Product:</h2>
```
kriscooke, line 20: accessibility heading requirement

app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
matthieu-foucault: We can stay on the safe side and leave them. A stricter typescript compiler config would probably be needed to be confident that we don't need those

matthieu-foucault (commented):
Code looks good. Testing in local dev environment...

app/containers/Admin/ApplicationReview/ApplicationReviewStepSelector.tsx
```diff
@@ -0,0 +1,203 @@
...
+      disabled
+      style={{opacity: 1, color: '#343a40', textAlign: 'left'}}
+    >
+      <FontAwesomeIcon icon={faLock} style={{marginRight: 8}} />
+      There is a newer draft of this application.
+    </Button>
+  );
+  return (
+    <Row>
+      <div className="col-xxl-6 col-xl-7 col-lg-8 col-md-10">
```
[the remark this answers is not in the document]
kriscooke: Yes - I've found react-bootstrap has a lot of quirks that don't always let you access the full features of each underlying Bootstrap component (I've found this in the `Modal` as well). 

In this case, [the react-bootstrap component](https://react-bootstrap.github.io/layout/grid/#col-props) doesn't implement the [underlying Bootstrap `xxl` breakpoint](https://getbootstrap.com/docs/5.0/layout/breakpoints/), which for context equates to a desktop monitor (or for me, Macbook Pro width and up).

This prevents the component from taking up a huge width on desktop far beyond the scale of the actual text content items.

app/components/GenericConfirmationModal.tsx
kriscooke: Yes! For now, I think it would be best to do a separate PR to pull these out, because it would update buttons and badges all over the site.

app/components/GenericConfirmationModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+        .btn-outline-primary#cancel:hover {
+          color: #fff;
+          background: #002040;
+        }
+        .btn-success#confirm {
+          background: #24883e;
+        }
+        .btn-outline-success#confirm {
+          color: #036;
+          border-color: #036;
```
kriscooke, line 75: Just noticed this one was wrong anyway (that color is blue, not green) - I'm going to remove the not-yet-implemented variants (`.btn-success`) from here for now, and rely on overriding these globally in a separate PR.

app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
```diff
@@ -109,14 +149,48 @@ export const ReviewSidebar: React.FunctionComponent<Props> = ({
...
+      title={`Confirm ${applicationReviewStep?.reviewStepByReviewStepId?.stepName} review is completed`}
+      onConfirm={() => {
+        markReviewStepComplete(true);
+        setShowUnresolvedCommentsModal(false);
+      }}
+      size="lg"
+      onCancel={() => setShowUnresolvedCommentsModal(false)}
+      confirmButtonVariant="outline-primary"
+      cancelButtonVariant="primary"
+      cancelButtonText="Cancel and resolve comments"
```
matthieu-foucault, line 172: Clicking on this button only cancels, it doesn't resolve comments. Should the text just be `Cancel`?

kriscooke: That button is intended to do nothing - analysts should manually resolve individual comments to ensure they are indeed Resolved. The text is meant to instruct the user what to do next (1. Cancel 2. Make sure your comments are resolved).

kriscooke: It's supposed to mean "I want to cancel, I will resolve comments before proceeding" but that was too long 😛

matthieu-foucault: I did have a second where I was wondering if it was going to automatically resolve the comments when I click the button.
 If you think it's clear as-is then I'm happy to leave it like that.

kriscooke: I could see how one might take it the other way though - perhaps just Cancel is safer.

**Side note**: It is still possible to proceed and mark a step completed without resolving all comments - this allows them to write comments about why an application was Rejected for example. It's just a checkpoint they requested to remind them to ensure they've addressed any unresolved comments.

matthieu-foucault: re: Side note, the modal text was very clear about that :+1:

matthieu-foucault (approved):

matthieu-foucault (changes requested):
holding off on adding margin to validation override notification

matthieu-foucault (approved):

PR #1610, opened 2021-04-19
fix: update contact us page email locations

naomiaro:
https://youtrack.button.is/issue/GGIRCS-1557
https://youtrack.button.is/issue/GGIRCS-2015

- Update email contact depending on help location
- cleans up all env vars related to emails.

MOBILE:
Screenshot 2021-04-19 at 4 30 17 PM

DESKTOP:
Screenshot 2021-04-19 at 4 28 54 PM

### DIFF
--- a/app/pages/resources/contact.tsx
+++ b/app/pages/resources/contact.tsx
@@ -3,9 +3,11 @@ import {Container, Alert} from 'react-bootstrap';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import {faInfoCircle} from '@fortawesome/free-solid-svg-icons';
 import {graphql} from 'react-relay';
+import getConfig from 'next/config';
 import {CiipPageComponentProps} from 'next-env';
 import {contactQueryResponse} from 'contactQuery.graphql';
 import DefaultLayout from 'layouts/default-layout';
+const getEmailShortForm = require('app/server/helpers/getEmailShortForm.js');
 
 interface Props extends CiipPageComponentProps {
   query: contactQueryResponse['query'];
@@ -26,6 +28,10 @@ class Contact extends Component<Props> {
   render() {
     const {query} = this.props;
     const {session} = query || {};
+    const supportEmail = getConfig()?.publicRuntimeConfig.SUPPORT_EMAIL;
+    const adminEmail = getEmailShortForm(
+      getConfig()?.publicRuntimeConfig.ADMIN_EMAIL
+    );
     return (
       <DefaultLayout session={session} title="Contact Us">
         <Container>
@@ -34,15 +40,15 @@ class Contact extends Component<Props> {
             <span className="pl-2">
               For help with your CIIP application, or questions about the
               CleanBC Industrial Incentive Program, please email{' '}
-              <a href="mailto:GHGRegulator@gov.bc.ca?subject=Support Request (CIIP)">
-                GHGRegulator@gov.bc.ca
+              <a href={`mailto:${adminEmail}?subject=Support Request (CIIP)`}>
+                {adminEmail}
               </a>
             </span>
           </Alert>
           <p className="px-4 py-3">
             To report a website error, contact the development team at{' '}
-            <a href="mailto:ggircs@gov.bc.ca?subject=Support Request (CIIP)">
-              ggircs@gov.bc.ca
+            <a href={`mailto:${supportEmail}?subject=Support Request (CIIP)`}>
+              {supportEmail}
             </a>
             .
           </p>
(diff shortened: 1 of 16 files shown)


app/pages/resources/contact.tsx
```diff
@@ -21,19 +23,28 @@ class Contact extends Component<Props> {
...
-              GHGRegulator@gov.bc.ca
-            </a>{' '}
-            with any questions or concerns.
+          <Alert variant="info">
+            <FontAwesomeIcon icon={faInfoCircle} />
+            <span className="pl-2">
+              For help with your CIIP application, or questions about the
+              CleanBC Industrial Incentive Program, please email{' '}
+              <a href="mailto:GHGRegulator@gov.bc.ca?subject=Support Request (CIIP)">
+                GHGRegulator@gov.bc.ca
```
kriscooke, line 38: We've been trying to replace any hardcoded constants like contact email with getting them from a runtime variable. 

Those are configured in [`values.yaml`](https://github.com/bcgov/cas-ciip-portal/blob/b37f88784871d92f7dbdc0679888a9e2b6b2c8a7/helm/cas-ciip-portal/values.yaml#L63-L64) (as well as in the corresponding `values-[env].yaml` if we want to override that in any particular env), and set in here the Next runtime in [`next.config.js`](https://github.com/bcgov/cas-ciip-portal/blob/b37f88784871d92f7dbdc0679888a9e2b6b2c8a7/app/next.config.js#L25). Usage example [here](https://github.com/bcgov/cas-ciip-portal/blob/b37f88784871d92f7dbdc0679888a9e2b6b2c8a7/app/components/helpers/HelpButton.tsx#L17)

It looks like `adminEmail` (for GHGRegulator) still needs to be set up in the Next config for use on the client side, even though we've done all the variable replacements for `SUPPORT_EMAIL`. 

This might be a good opportunity to remove any other hardcoded client-side instances of the admin email at the same time.

kriscooke (changes requested):
Just a few variable replacements to remove hardcoded values like email addresses.

As far as the page itself, looks great!

app/pages/resources/contact.tsx
naomiaro: Pulling in this ticket to address the issues: https://youtrack.button.is/issue/GGIRCS-2015

helm/cas-ciip-portal/values.yaml
```diff
@@ -59,8 +59,8 @@ dag:
 
 env:
   smtpConnectionString: smtp://apps.smtp.gov.bc.ca/?port=25&ignoreTLS=true&secure=false
-  senderEmail: BCCAS <no-reply.cas@gov.bc.ca>
-  adminEmail: GHG Regulator <GHGRegulator@gov.bc.ca>
+  senderEmail: no-reply.cas@gov.bc.ca
+  adminEmail: GHGRegulator@gov.bc.ca
```
kriscooke, line 63: Never noticed it before, but I wonder why we had the much more nicely spelled-out sender/receiver names in dev and test, but not in prod? 
ie: "CIIP Admin, BC Climate Action Secretariat" and "BC Climate Action Secretariat" in values-test and values-dev

Maybe we should be update [the lines in `sendMail`](https://github.com/bcgov/cas-ciip-portal/pull/1610/files#diff-a4b1bb3c67981694cbd9ddc2073beb8a68075b86310c6054cb26e778f4aad76dR15-R16) that hardcode the name part to this instead...

app/pages/resources/contact.tsx
```diff
@@ -21,19 +24,32 @@ class Contact extends Component<Props> {
...
   render() {
     const {query} = this.props;
     const {session} = query || {};
+    const supportEmail = getConfig()?.publicRuntimeConfig.SUPPORT_EMAIL;
+    const supportMailToUrl = supportEmail
+      ? `mailto:${supportEmail}?subject=Support Request (CIIP)`
+      : '#';
+    const adminEmail = getConfig()?.publicRuntimeConfig.ADMIN_EMAIL;
+    const adminMailToUrl = adminEmail
+      ? `mailto:${adminEmail}?subject=Support Request (CIIP)`
```
kriscooke, line 36: Since this is a general contact page and not specific to support requests, a subject line like "CIIP Inquiry" might be more appropriate for the `adminMailToUrl` here.

Conversely, we only ask them to email ggircs@gov specifically for support reasons, so that one makes sense.

helm/cas-ciip-portal/values.yaml
naomiaro: Yeah, maybe taking the information from values-test as that is what people saw when doing QA.

kriscooke (approved):

kriscooke (dismissed):

kriscooke (approved):

PR #1612, opened 2021-04-20
feat: facilities page defaults to the current reporting year

matthieu-foucault:
or the previous one if there is no open reporting year


dleard (approved):
LGTM

pbastia (approved):
👌

PR #1613, opened 2021-04-20
Feat: exclude biomass from emissions

dleard:
- add `is_ciip_emission` boolean column to gas table
- add unique index on gas.gas_type
- add computed column to retrieve total emissions (on application_revision)
- update ciip incentive function to use new computed column
- pgTap tests


matthieu-foucault (commented):
Code looks good but I see a CI failure with the graphql schema

schema/deploy/computed_columns/application_revision_total_ciip_emissions.sql
```diff
@@ -0,0 +1,24 @@
+-- Deploy ggircs-portal:computed_columns/application_revision_facility_emission to pg
+-- requires: tables/gas_001
+-- requires: views/ciip_emission
+
+begin;
+
+-- Computed column returns the sum of (ciip) emissions for an application revision.
+create or replace function ggircs_portal.application_revision_total_ciip_emissions(application_revision ggircs_portal.application_revision)
+returns table (total_ciip_co2e_emissions numeric, total_ciip_annual_emissions numeric)
```
matthieu-foucault, line 9: Just return a numeric with the `total_ciip_co2e_emissions`. You cannot do a `sum(annual_emission)`, since those are different gases.

matthieu-foucault (changes requested):
I actually found something on a second read

schema/deploy/computed_columns/application_revision_total_ciip_emissions.sql
```diff
@@ -0,0 +1,24 @@
...
+-- Computed column returns the sum of (ciip) emissions for an application revision.
+create or replace function ggircs_portal.application_revision_total_ciip_emissions(application_revision ggircs_portal.application_revision)
+returns numeric
+  as $$
+    select sum(annual_co2e) as total_ciip_co2e_emissions from ggircs_portal.ciip_emission ce
+    join ggircs_portal.gas
+    on ce.gas_type = gas.gas_type
+    and ce.version_number = application_revision.version_number
+    and ce.application_id = application_revision.application_id
+    and ce.gas_type != 'CO2'
```
matthieu-foucault, line 16: The CO2 gas type never actually made it to prod, so I don't think that we need to filter it out.
```suggestion
```

schema/test/unit/computed_columns/application_revision_total_ciip_emissions_test.sql
```diff
@@ -0,0 +1,73 @@
...
+select results_eq (
+  $$
+    with record as (select row(application_revision.*)::ggircs_portal.application_revision from ggircs_portal.application_revision where application_id = 1 and version_number = 1)
+    select ggircs_portal.application_revision_total_ciip_emissions((select * from record))
+  $$,
+  $$
+    select sum(annual_co2e)::numeric - (select sum(annual_co2e)::numeric from ggircs_portal.ciip_emission where version_number=1 and application_id=1 and gas_type ='CO2')::numeric from ggircs_portal.ciip_emission
+    where version_number = 1 and application_id = 1
+  $$,
+  'total_ciip_emissions function ignores the legacy catchall CO2 emissions when returning the sum of annual_emission'
```
matthieu-foucault, line 68: you can remove this test, there is no `CO2` gas

matthieu-foucault (changes requested):
Just a bit of unnecessary code that can be removed

matthieu-foucault (approved):

PR #1615, opened 2021-04-21
feat: rename "report a problem" field to "add a comment"

matthieu-foucault:


app/components/Forms/AddCommentField.tsx
```diff
@@ -52,21 +58,23 @@ const ProblemReportField: React.FunctionComponent<FieldProps<string>> = (
...
           <p>
-            Please leave a detailed message below explaining any issues
-            encountered on this page.
-            <br />
-            This message will be sent with your application when it is
-            submitted.
-            <br />A representative from the Climate Action Secretariat may
-            contact you for further clarification.
+            For any technical issue, or to review documentation, please click on
+            the help bubble in the bottom right of your screen.
```
kriscooke, line 77: Minor detail (what they'll see is a button):
```suggestion
            the help button in the bottom right of your screen.
```
I'm also wondering if we can delay adding this sentence and add it as part of the card implementing the help button (could cause confusion for QA testers)?

pbastia (approved):
Looks great!  👍 

The success message in the toast notification is a little hard for me to understand quickly - but it's probably my English that is at fault here

app/components/Forms/AddCommentField.tsx
matthieu-foucault: I would expect that help button to be in our next release to QA

kriscooke: OK. Either way, "button" is more correct - they will not see any bubble in the bottom of the screen.

kriscooke (approved):

kriscooke (approved):

PR #1617, opened 2021-04-21
Feat: operator access progress indicator

pbastia:
Introducing a reusable UI component to show process steps as a static progress bar.
it can be customized with:
- step #
- badge colour
- description

image

This refactors the reporter's organisation view with a clearer indication of the status of his access requests.

### DIFF
--- a/app/components/ProgressStepIndicator.tsx
+++ b/app/components/ProgressStepIndicator.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {Col, ProgressBar, Row} from 'react-bootstrap';
+import {Col, Row} from 'react-bootstrap';
 import {Variant} from 'react-bootstrap/esm/types';
 
 interface NumberCircleProps {
@@ -24,12 +24,11 @@ const NumberedCircle: React.FunctionComponent<NumberCircleProps> = ({
       <div className={`numberedCircle badge-${style}`}>{number}</div>
       <style jsx>{`
         .numberedCircle {
-          margin-top: -29px;
-          width: 38px;
-          line-height: 37px;
+          margin-top: -1.75rem;
+          width: 2.44rem;
+          line-height: 2.44rem;
           border-radius: 50%;
           text-align: center;
-          font-size: 16px;
           font-weight: bold;
         }
       `}</style>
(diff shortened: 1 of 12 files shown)


app/tests/unit/components/ProgressStepIndicator.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {mount} from 'enzyme';
+import React from 'react';
+
+describe('The Progress Step Indicator', () => {
+  it('Renders with no steps', () => {
+    const componentUnderTest = mount(<ProgressStepIndicator steps={[]} />);
+    expect(componentUnderTest).toMatchSnapshot();
+  });
+
+  it('Matches snapshot with 3 tests', () => {
```
kriscooke, line 12: 😉 
```suggestion
  it('Matches snapshot with 3 steps', () => {
```

app/tests/unit/components/ProgressStepIndicator.test.tsx
```diff
@@ -0,0 +1,53 @@
+import StatusBadgeColor from 'components/helpers/StatusBadgeColor';
+import ProgressStepIndicator from 'components/ProgressStepIndicator';
+import {mount} from 'enzyme';
```
kriscooke, line 3: I think you should be able to use `shallow` here instead, n'est-ce pas?

I often set up tests by copy pasting another, but you only need the full `mount` rendering sometimes for a fuller rendering of grandchildren and certain event simulations, which I don't see here.

pbastia: I like that `mount` expands the React objects in this case but you're right, maybe we don't need it here

app/pages/reporter/index.tsx
```diff
@@ -79,7 +78,7 @@ export default class Reporter extends Component<Props> {
     return (
       <DefaultLayout showSubheader session={session} title="My Operators">
         <Row>
```
kriscooke, line 80: I'm guessing you removed the 8-column width on the `Col` below in order to center it on the page? This will center it while maintaining the original page content width. I'd recommend sticking with that because once content gets too spread too widely apart it can be hard to associate related items (in this case, both the steps and the table row columns are quite far apart).

```suggestion
        <Row className="justify-content-md-center">
          <Col md={10} lg={8}>
```

kriscooke: Screen Shot 2021-04-22 at 7 21 41 AM

matthieu-foucault:
@kriscooke our `cypress-axe` dependency is out-of-date. I attempted to update it earlier this month, but it caused more failures in e2e tests, on components coming from our `bootstrap-typeahead` dependency, so I gave up on updating it at the time.

app/tests/unit/components/ProgressStepIndicator.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+        ]}
+      />
+    );
+    expect(componentUnderTest).toMatchSnapshot();
+    expect(componentUnderTest.find('.numberedCircle')).toHaveLength(3);
+    expect(componentUnderTest.find('div.stepDescription')).toHaveLength(3);
+  });
+
+  const stepCountTestCases = [1, 2, 3, 4, 5, 6];
+  test.each(stepCountTestCases)('Generates %p progress items', (stepCount) => {
```
kriscooke, line 40: Just a thought - but instead of 6 nearly identical test cases, might it be more valuable to have some assertions that the `wrapper.find('.numberedCircle').text()` equals the number (likewise with the descriptions) and that the `className` contains the status color?

app/tests/unit/components/ProgressStepIndicator.test.tsx
pbastia: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering reading this, I think it's better to stick to mount in this case - since we're testing rendering and not behaviour

app/pages/reporter/index.tsx
kriscooke: I'd recommend sticking with the scope of this feature (progress bar) and maintain the existing layout, as there are other appropriate venues to rethink page layout.

app/components/ProgressStepIndicator.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <div
+          role="progressbar"
+          className="progress-bar"
+          aria-valuenow={0}
+          aria-valuemin={0}
+          aria-valuemax={100}
+          aria-label="operator-progress-indicator"
+          style={{width: '0%'}}
+        />
+      </div>
```
pbastia, line 54: This could be refactored in a separate class - not sure if we need it at this moment

app/tests/unit/components/ProgressStepIndicator.test.tsx
kriscooke: (I didn't mean the new assertions also need to be repeated 6 times, but good enough for go time! 😛)

pbastia: Haha, this way it cycles through all the input possibilities, we've got full test coverage!

kriscooke (approved):

kriscooke (dismissed):

kriscooke (approved):

PR #1620, opened 2021-04-22
Feat: facility application progress

pbastia:
Adding a static indicator for the application process, reusing the `ProgressStepIndicator` component.

image

### DIFF
--- a/app/components/ProgressStepIndicator.tsx
+++ b/app/components/ProgressStepIndicator.tsx
@@ -42,15 +42,14 @@ export const ProgressStepIndicator: React.FunctionComponent<Props> = ({
   title
 }) => {
   return (
-    <>
+    <div className="progressStepIndicator">
       {title && (
         <>
-          <Row>
+          <Row className="mb-4">
             <Col>
               <h2>{title}</h2>
             </Col>
           </Row>
-          <br />
         </>
       )}
       <div className="progress ml-1 mr-1">
@@ -60,7 +59,7 @@ export const ProgressStepIndicator: React.FunctionComponent<Props> = ({
           aria-valuenow={0}
           aria-valuemin={0}
           aria-valuemax={100}
-          aria-label="progress-step-indicator"
+          aria-label="progress bar indicating the steps represented by this component"
           style={{width: '0%'}}
         />
       </div>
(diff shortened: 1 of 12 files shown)


app/pages/reporter/index.tsx
```diff
@@ -79,6 +81,30 @@ export default class Reporter extends Component<Props> {
...
+                    {
+                      description: 'Request approved',
+                      badgeStyle: StatusBadgeColor.APPROVED,
+                      number: 3
+                    }
+                  ]}
+                />
+              </Col>
+            </Row>
+            <br />
```
matthieu-foucault, line 107: `<br />` should not be used to create margins

app/pages/reporter/facilities.tsx
```diff
@@ -70,6 +73,37 @@ class FacilitiesList extends Component<Props> {
...
+                  description: 'Application approved or rejected',
+                  badgeStyle: StatusBadgeColor.APPROVED,
+                  number: 4
+                }
+              ]}
+            />
+          </Col>
+        </Row>
+        <br />
+        <br />
```
matthieu-foucault, line 106: `<br />` should not be used to create margins

app/components/ProgressStepIndicator.tsx
```diff
@@ -37,18 +38,29 @@ const NumberedCircle: React.FunctionComponent<NumberCircleProps> = ({
...
   return (
     <>
+      {title && (
+        <>
+          <Row>
+            <Col>
+              <h2>{title}</h2>
+            </Col>
+          </Row>
+          <br />
```
matthieu-foucault, line 53: `<br />` should not be used to create margins

app/components/ProgressStepIndicator.tsx
```diff
@@ -69,6 +81,9 @@ export const ProgressStepIndicator: React.FunctionComponent<Props> = ({
         ))}
       </Row>
       <style jsx global>{`
+        h2 {
+          font-size: 1.5rem;
+        }
```
matthieu-foucault, line 86: move this to a local `style jsx` to avoid side effects (and the ones below if possible)

app/components/ProgressStepIndicator.tsx
```diff
@@ -37,18 +38,29 @@ const NumberedCircle: React.FunctionComponent<NumberCircleProps> = ({
...
+      )}
       <div className="progress ml-1 mr-1">
         <div
           role="progressbar"
           className="progress-bar"
           aria-valuenow={0}
           aria-valuemin={0}
           aria-valuemax={100}
-          aria-label="operator-progress-indicator"
+          aria-label="progress-step-indicator"
```
matthieu-foucault, line 63: Given that `aria-label` is read by a screen reader, I would assume that it should be a readable label, not an id

matthieu-foucault (changes requested):

app/components/ProgressStepIndicator.tsx
pbastia: Because we use bootstrap `Row` and `Col` components we have to use `global` here - but there won't be side-effects since this component doesn't render any children

pbastia: ah nevermind, I got confused again with the scope of `global`

matthieu-foucault: Wouldn't a global style affect all components on the page where this component is rendered?
Given that the `h2` is rendered directly by this component, you should be able to use a local style

matthieu-foucault (dismissed):

matthieu-foucault (approved):

PR #1623, opened 2021-04-23
Update legacy review_step name from 'other' to 'application'

kriscooke:
...Again. Realized how it shows up as 'Other Review' in the review sidebar was awkward and confusing once I presented it in sprint review. But @matthieu-foucault had a brainwave to name it "application" so it shows up as "Application Review".

This is only relevant to pre-existing reviews created prior to this year.

### DIFF
--- a/schema/verify/tables/review_step_001.sql
+++ b/schema/verify/tables/review_step_001.sql
@@ -1,0 +1,7 @@
+-- Verify ggircs-portal:tables/review_step_001 on pg
+
+begin;
+
+select exists(select * from ggircs_portal.review_step where step_name='application');
+
+rollback;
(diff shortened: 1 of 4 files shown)


schema/verify/tables/review_step_001.sql
```diff
@@ -0,0 +1,7 @@
+-- Verify ggircs-portal:tables/review_step_001 on pg
+
+begin;
+
+select exists(select * from ggircs_portal.review_step where step_name='application');
```
dleard, line 5: Nothing wrong here, just a note on verify files because I had to think about this one for a second. 

We changed the immutable-before-tag check in .bin so that it does not check verify/revert files, only deploy files. That way verify files can be changed in the future regardless of their location in the plan. This is important because this verify file would fail `sqitch verify` in the future if we were to create a `review_step_00x` migration that changed the same step_name again. 

`sqitch verify` treats our files appended with `_00x` as separate (as opposed to reworks, which it knows are incremental), and runs them all on the final state of the database. If we changed the name again this verify file would look for step_name='application' where none exist and fail.

dleard (approved):
Looks good! Just a quick note on verify files if you're interested since we made a small change not too long ago that affects the files appended with `00x`.

PR #1624, opened 2021-04-26
Re-enable review comment creation with simplified modal

kriscooke:
- 'New Comment' button in review sidebar opens simplified modal to write a comment and add it to the selected review step
- Checkbox defines the comment type, aka. intended audience for the comment (`general` for applicants to see, or for `internal` use only)
- Visibility indicator confirms who can see the comment before completing the action
- Comment is saved and associated to the review step to which it was added
[GGIRCS-2194](https://youtrack.button.is/issue/GGIRCS-2194)

Screen Shot 2021-04-26 at 8 48 07 AM
Screen Shot 2021-04-26 at 8 48 17 AM

### DIFF
--- a/app/components/Admin/AddReviewCommentModal.tsx
+++ b/app/components/Admin/AddReviewCommentModal.tsx
@@ -6,7 +6,7 @@ import {faLock, faEye} from '@fortawesome/free-solid-svg-icons';
 interface Props {
   show: boolean;
   title: string;
-  onSubmit: (detail: object) => void;
+  onSubmit: ({commentText: string, isInternalComment: boolean}) => void;
   onHide: () => void;
 }
 
--- a/app/components/Admin/AddReviewCommentModal.tsx
+++ b/app/components/Admin/AddReviewCommentModal.tsx
@@ -1,0 +1,103 @@
+import React, {useState} from 'react';
+import {Modal, Button, Form, FormCheck} from 'react-bootstrap';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import {faLock, faEye} from '@fortawesome/free-solid-svg-icons';
+
+interface Props {
+  show: boolean;
+  title: string;
+  onSubmit: (detail: object) => void;
+  onHide: () => void;
+}
+
+export const AddReviewCommentModal: React.FunctionComponent<Props> = ({
+  show,
+  title,
+  onSubmit,
+  onHide
+}) => {
+  const [isInternalComment, setIsInternalComment] = useState(false);
+  const [commentText, setCommentText] = useState('');
+
+  const clearForm = () => {
+    setIsInternalComment(false);
+    setCommentText('');
+  };
+  const handleHide = () => {
+    onHide();
+    clearForm();
+  };
+
+  return (
+    <Modal id="add-comment-modal" size="lg" show={show} onHide={handleHide}>
+      <Modal.Header closeButton>
+        <h2>{title}</h2>
+      </Modal.Header>
+      <Modal.Body>
+        <Form.Label>Comment:</Form.Label>
+        <Form.Control
+          as="textarea"
+          rows={3}
+          value={commentText}
+          onChange={(e) => setCommentText(e.target.value)}
+        />
+        <FormCheck>
+          <FormCheck.Label>
+            <FormCheck.Input
+              checked={isInternalComment}
+              onChange={(e) => setIsInternalComment(e.target.checked)}
+            />
+            Visible for <strong>internal</strong> use only
+          </FormCheck.Label>
+        </FormCheck>
+      </Modal.Body>
+      <Modal.Footer>
+        <span>
+          This comment will {isInternalComment && <strong>not</strong>} be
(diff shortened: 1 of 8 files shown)


app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
```diff
@@ -191,6 +195,27 @@ export const ReviewSidebar: React.FunctionComponent<Props> = ({
...
+      }
+    };
+    await createReviewCommentMutation(
+      environment,
+      variables as createReviewCommentMutationVariables,
+      applicationReviewStep.id,
+      `ReviewSidebar_${commentType}Comments`
+    );
+    setShowAddCommentModal(false);
+  };
```
kriscooke, line 217: I could not quite find a way to make this testable in the ReviewSidebar test (asserting that the mutation is called) as it's quite internal to the component... 🤔

app/tests/unit/components/Admin/AddReviewCommentModal.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  });
+  it('closing the Bootstrap modal clears the form', () => {
+    const testComment = 'Typing a test comment';
+    const wrapper = mount(getTestElement({}));
+    const textarea = wrapper.find('Modal').first().find('textarea');
+
+    // Fill the form as the user would, before attempting to reset:
+    act(() => {
+      textarea.simulate('change', {target: {value: testComment}});
+    });
```
kriscooke, line 76: I tried many ways, but unfortunately could not seem to accurately simulate the checkbox in the same way to test it being cleared along with the textarea (it would not change its state to `checked`).

app/components/Admin/AddReviewCommentModal.tsx
```diff
@@ -0,0 +1,105 @@
+import React, {useState} from 'react';
+import {Modal, Button, Form, FormCheck} from 'react-bootstrap';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import {faLock, faEye} from '@fortawesome/free-solid-svg-icons';
+
+interface Props {
+  show: boolean;
+  title: string;
+  onSubmit: (detail: object) => void;
```
matthieu-foucault, line 9:
```suggestion
  onSubmit: (detail: {commentText: string, isInternalComment: boolean}) => void;
```

The `object` primitive type is only useful the `Object` API, otherwise not so much, as it doesn't allow you to access properties of the object

app/mutations/application_review_step/createReviewCommentMutation.ts
```diff
@@ -1,17 +1,28 @@
...
 
 const mutation = graphql`
-  mutation createReviewCommentMutation($input: CreateReviewCommentInput!) {
+  mutation createReviewCommentMutation(
+    $input: CreateReviewCommentInput!
+    $connections: [ID!]!
+  ) {
     createReviewComment(input: $input) {
-      reviewCommentEdge {
+      reviewCommentEdge @appendEdge(connections: $connections) {
```
kriscooke, line 16: @dleard FYI this is a handy [declarative alternative](https://relay.dev/docs/guided-tour/list-data/updating-connections/#using-declarative-directives) to the imperative updater function you were previously helping me with for removing/adding to connections (works with `@deleteEdge` as well).

app/containers/Admin/ApplicationReview/ReviewSidebar.tsx
matthieu-foucault: Do you mean that triggering `handleAddReviewComment` is the issue? 
The trick is to find it in the props of `AddReviewCommentModal` and call it directly. 
I did something similar here:
https://github.com/bcgov/cas-ciip-portal/blob/167365e41de8f02510a8873a7051d378ebf2f6e2/app/tests/unit/containers/Applications/ApplicationWizard.test.tsx#L210-L215

If you mean spying on the mutation I see that you've done that for another test already:

https://github.com/bcgov/cas-ciip-portal/blob/758dd771667548cb4f767496c093df5fcc268215/app/tests/unit/containers/Admin/ApplicationReview/ReviewSidebar.test.tsx#L220-L223

kriscooke: > Do you mean that triggering handleAddReviewComment is the issue?

Yes, that is the issue because `handleAddReviewComment` is not a prop (and it shouldn't be) - so there's no access it to call it.

matthieu-foucault: Isn't it the value of the `onSubmit` prop of the `AddReviewCommentModal` below?

kriscooke: Haha. Of course. It is. Tired eyes end of day when I wrote those tests yesterday!

matthieu-foucault (approved):
Looks all good to me. My comments do not **need** to be resolved if you want to merge this and move on :shipit:

matthieu-foucault (approved):

PR #1625, opened 2021-04-26
Feature/help menu

naomiaro:
https://youtrack.button.is/issue/GGIRCS-2298

- Enable the HelpButton component for reporters
- Update documentation and mailto links for reporters

Screenshot 2021-04-26 at 11 40 47 AM

### DIFF
--- a/app/components/helpers/HelpButton.tsx
+++ b/app/components/helpers/HelpButton.tsx
@@ -15,12 +15,17 @@ const HelpButton: React.FunctionComponent<Props> = ({
   isInternalUser = true
 }) => {
   const supportEmail = getConfig()?.publicRuntimeConfig.SUPPORT_EMAIL;
-  const mailToUrl = supportEmail
+  const adminEmail = getConfig()?.publicRuntimeConfig.ADMIN_EMAIL;
+  const supportMailToUrl = supportEmail
     ? `mailto:${supportEmail}?subject=Internal Support Request`
     : '#';
+  const adminMailToUrl = adminEmail
+    ? `mailto:${adminEmail}?subject=Support Request (CIIP)`
+    : '#';
+  const mailToUrl = isInternalUser ? supportMailToUrl : adminMailToUrl;
   const docsUrl = isInternalUser
     ? 'https://github.com/bcgov/cas-ciip-portal/blob/master/docs/admin-analyst-guide.md'
-    : '#';
+    : 'https://github.com/bcgov/cas-ciip-portal/blob/master/docs/reporting-guide.md';
 
   const [isOpened, setIsOpened] = useState(false);
   const toggleHelpBubble = () => {
@@ -66,7 +71,25 @@ const HelpButton: React.FunctionComponent<Props> = ({
               </li>
             </ul>
           ) : (
-            ''
+            <ul>
+              <li>
+                <span role="img" aria-label="open book emoji">
+                  📖
+                </span>
+                <a href={docsUrl} target="_blank" rel="noopener noreferrer">
+                  See the help documentation
+                </a>
+              </li>
+              <li>OR</li>
+              <li>
+                <span role="img" aria-label="letter emoji">
+                  ✉️
+                </span>
+                <a href={mailToUrl}>
+                  Contact us to request specific assistance
+                </a>
+              </li>
+            </ul>
           )}
           <span className="triangle">◥</span>
         </div>
--- a/app/components/helpers/HelpButton.tsx
+++ b/app/components/helpers/HelpButton.tsx
@@ -1,5 +1,6 @@
 import React, {useState} from 'react';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import Link from 'next/link';
 import getConfig from 'next/config';
 import {
   faQuestion,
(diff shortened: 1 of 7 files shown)


kriscooke (changes requested):
AC for 2298: For reporters, "Contact us" option should link to the `/resources/contact` page (instead of a mailto link), which you recently updated so they can identify who to address their question to :)

As the card describes, reporters will be reaching out to either of the two email contacts, based on whether they need program-related support or have problems with the website.

naomiaro:
Oh woops.. Monday morning reading skills 😁

app/components/helpers/HelpButton.tsx
```diff
@@ -66,7 +67,25 @@ const HelpButton: React.FunctionComponent<Props> = ({
               </li>
             </ul>
           ) : (
-            ''
```
kriscooke, line 69: Perhaps better than duplicating the help bubble content for reporters (which could  facilitate accidental template divergence) would be to extract constants, `option1` and `option2`:
```
const option1 = (
  <>
    span (based on user)
    anchor link (based on user)
  </>
);

...

<ul>
  <li>{option1}</li>
  <li>OR</li>
  <li>{option2}</li>
</ul>
```

naomiaro: Yeah if we're sticking to only 2 options each we could do this.

app/components/helpers/HelpButton.tsx
```diff
@@ -66,7 +67,25 @@ const HelpButton: React.FunctionComponent<Props> = ({
...
+                <a href={docsUrl} target="_blank" rel="noopener noreferrer">
+                  See the help documentation
+                </a>
+              </li>
+              <li>OR</li>
+              <li>
+                <span role="img" aria-label="letter emoji">
+                  ✉️
+                </span>
+                <Link href="/resources/contact">
```
kriscooke, line 84: It would probably be best for this to open in a new tab so it doesn't force them to navigate away (in case they're in the middle of an application, for example).

You can use the Link component to wrap an `<a>` link tag in order to specify the `target` and `rel` attributes as above.

matthieu-foucault: FYI, I believe that a Link should always have an `<a>` element (or a component that renders an `<a>` element) as a child, otherwise a warning will be printed in the console (I couldn't find why in the nextjs docs)

app/components/helpers/HelpButton.tsx
```diff
@@ -66,7 +67,25 @@ const HelpButton: React.FunctionComponent<Props> = ({
...
+                  📖
+                </span>
+                <a href={docsUrl} target="_blank" rel="noopener noreferrer">
+                  See the help documentation
+                </a>
+              </li>
+              <li>OR</li>
+              <li>
+                <span role="img" aria-label="letter emoji">
+                  ✉️
```
matthieu-foucault, line 82: Using font awesome icons would ensure that we have the same icon for all browsers/OS. This looks like a yellow rectangle to me (chromium/archlinux)

app/components/helpers/HelpButton.tsx
naomiaro: The Footer element doesn't use `<a>` I think it's more if you need to override something on the a tag like target etc?

app/components/helpers/HelpButton.tsx
naomiaro: Yeah I was wondering if I should switch from using the unicode icons here in both views. Fontawesome would look a lot flatter though as it's only one colour etc. Partly why I was wondering if they were custom icons by Raj before.

naomiaro: Screenshot 2021-04-26 at 12 55 43 PM
 There's another unicode icon at least in the analyst dashboard as well.

matthieu-foucault: image

This is what those look like on my machine. 
Maybe if we used a specific font those would be rendered the same way on all browsers. Not sure if BC Sans, which is the one we're supposed to use according to the [BC design system](https://developer.gov.bc.ca/Design-System/About-the-Design-System), has unicode characters in it.

naomiaro: Yeah we can try setting a specific font for unicode icons and then fallback to fontawesome if they still... don't look awesome 😬

kriscooke: @matthieu-foucault it's probably a problem of the specific emoji font you have installed, since ArchLinux leaves you more on your own to do many of the things we take for granted on other OS's: 

> Emojis should work without any configuration once you have at least one emoji font installed of supported format.
https://wiki.archlinux.org/index.php/fonts#Emoji_and_symbols

Emojis are rarely an issue on Mac and Windows these days, which are the OS's we expect the vast majority of users to be on. As a result many are now leaving behind SVG libs like FontAwesome to use the natively installed unicode icons. 

For the occasional character that's not rendered (say they're using ArchLinux), it simply underscores the importance of not using icons alone to convey meaning - in our case, the text link is well described and the icon is just decoration.

naomiaro: Yeah I haven't tried to use unicode symbols in years for a UI. Should we stick with what we have now then since they work alright on Windows/Mac?

kriscooke: Yes, it should be fine to stick with the unicode. We have teammates using Windows machines to help us QA them, and we can also use references like [Emojipedia](https://emojipedia.org/envelope/) which helpfully lists the cross-platform support/rendering of a specific character (in this case, scrolling down you can see a snapshot of how the envelope char that we're adding renders in MS products).

matthieu-foucault: FYI, it looks like I see the Windows emojis on CIIP, but the Google ones on Github, so the page's font does seem to have an impact on which emojis are rendered (at least for me). Somehow all of the Microsoft ones have a thick border, which is an odd design choice. The google envelope is the yellow rectangle one (with a slightly different yellow for the lines): https://emojipedia.org/envelope/

kriscooke: > I see the Windows emojis on CIIP, but the Google ones on Github, so the page's font does seem to have an impact on which emojis are rendered (at least for me)

Interestingly, since Github gets the interesting challenge of trying to render emojis on *goodness knows what arcane operating system* its users may be on, inspection shows they're using a clever progressive enhancement technique using web components / custom HTML elements to render a `.png` fallback image if unavailable:

Screen Shot 2021-04-26 at 2 47 25 PM

kriscooke (approved):

PR #1627, opened 2021-04-26
Fix: application swrs report

dleard:
- add swrs_report_id to application table
- populate swrs_report_id data for existing applications via swrs_facility_id and reporting_year
- fix refresh_swrs_version_data function
- update creat_application_mutation_chain to populate application table with swrs_report_id on create
- write/update pgTap tests


schema/deploy/tables/application_003.sql
```diff
@@ -0,0 +1,23 @@
+-- Deploy ggircs-portal:tables/application_003 to pg
+-- requires: tables/application_002
+
+begin;
+
+alter table ggircs_portal.application add column swrs_report_id integer;
+create unique index application_swrs_report_id_uindex on ggircs_portal.application(swrs_report_id);
```
matthieu-foucault, line 7: Are we sure that we want a unique constraint here? 
It makes sense that we would not see two applications with the same swrs_report_id, but I am wondering what situation could trigger an error here, and how our application would handle that error. We already have a constraint on the (facility_id, reporting_year) pair if I remember correctly

schema/deploy/tables/application_003.sql
```diff
@@ -0,0 +1,23 @@
...
+    from swrs.report r
+    join ggircs_portal.facility f
+    on r.swrs_facility_id = f.swrs_facility_id
+    join ggircs_portal.application a
+    on f.id = a.facility_id
+    and a.reporting_year = r.reporting_period_duration
+)
+update ggircs_portal.application a_outer
+set swrs_report_id = (select swrs_report_id from app_swrs where id = a_outer.id)
+where a_outer.id = (select id from app_swrs where id = a_outer.id);
```
matthieu-foucault, line 21: This looks about right. I assume you tested this manually in your local environment?

schema/deploy/swrs_functions/refresh_swrs_version_data.sql
```diff
@@ -53,30 +53,23 @@ begin
           raise notice 'Created new swrs version (version 0) for application ID: %', application_temp_row.id;
 
       -- Do nothing if a swrs report does not exist
-      elsif (report_imported_at is not null) then
+      elsif (generated_report_id is not null and generated_report_id != application_temp_row.report_id) then
```
matthieu-foucault, line 56: I am actually not quite sure that using the report_id is the best way to do that, since the `report_id` is defined by the ordering of the reports in the [materialized view, which has no `order by` clause](https://github.com/bcgov/cas-ggircs/blob/develop/deploy/swrs/transform/materialized_view/report.sql), and is thus unpredictable. I am not even sure that we should be using the `report_id` anywhere... Looking at our prod data, we have multiple applications with the same `report_id` (from different facility/organisation).

The `swrs.report` table does have a `version` column, which could be a better alternative

schema/deploy/tables/application_003.sql
dleard: We do have a uindex on the (facility_id, reporting_year) pair. I figured we'd probably want a unique constraint on the swrs_report_id since there should only ever be 1 per application. But I can take it out if you think it could be a problem somewhere?

schema/deploy/tables/application_003.sql
dleard: Yeah, I had to change some dates & stuff locally, but I managed to get it to update some existing applications (and not update others where the reporting year doesn't match)

schema/deploy/swrs_functions/refresh_swrs_version_data.sql
dleard: Would this mean adding a version column to the application table to compare with swrs.report.version? Is that not confusing with our use of `version` for application 'revisions' across the CIIP portal? I could fall back to the original imported_at <-> updated_at comparison to decide whether a set of form results needs updating.

matthieu-foucault: `application.swrs_report_version` should prevent any confusion

schema/deploy/mutations/create_application_mutation_chain.sql
```diff
@@ -18,10 +18,11 @@ begin
     raise exception 'The application window is closed';
   end if;
 
-  insert into ggircs_portal.application(facility_id, reporting_year, report_id)
+  insert into ggircs_portal.application(facility_id, reporting_year, report_id, swrs_report_id)
   values (
     facility_id_input, current_reporting_year,
-    (select report_id from ggircs_portal.facility where id = facility_id_input)
+    (select report_id from ggircs_portal.facility where id = facility_id_input),
```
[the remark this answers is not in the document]
matthieu-foucault: ...or don't get report_id at all, given the other conversation below

schema/deploy/tables/application_003.sql
matthieu-foucault: The `swrs_report_id` **should** be unique, but there is once instance of historical data (from 2011) where that is not the case. Given that this value is assigned by a third party service, it would probably be better to not create unique constraints in our DB based on it, so that we are more resilient to a potential bug in SWRS

schema/deploy/mutations/create_application_mutation_chain.sql
```diff
@@ -18,10 +18,11 @@ begin
     raise exception 'The application window is closed';
   end if;
 
-  insert into ggircs_portal.application(facility_id, reporting_year, report_id)
+  insert into ggircs_portal.application(facility_id, reporting_year, swrs_report_id, swrs_report_version)
   values (
     facility_id_input, current_reporting_year,
-    (select report_id from ggircs_portal.facility where id = facility_id_input)
+    (select swrs_report_id from swrs.report where swrs_facility_id = (select swrs_facility_id from ggircs_portal.facility where id = facility_id_input) and reporting_period_duration = current_reporting_year),
+    (select version from swrs.report where swrs_facility_id = (select swrs_facility_id from ggircs_portal.facility where id = facility_id_input) and reporting_period_duration = current_reporting_year)
```
matthieu-foucault, line 25: should you use a `with` statement to avoid doing the same query twice?

matthieu-foucault (changes requested):
just one minor change

matthieu-foucault (dismissed):
just needs a rebase

matthieu-foucault (approved):

PR #1628, opened 2021-04-26
Fix: product sort by date crash

dleard:
Product sorting bug:
The column to sort on was passed incorrectly to the filter component.
date_modified -> updated_at

Organisation request table bug:
Page was missing getInitialProps() so it wasn't setting the relay variables on first load

### DIFF
--- a/app/containers/Products/ProductListContainer.tsx
+++ b/app/containers/Products/ProductListContainer.tsx
@@ -23,7 +23,7 @@ interface Props {
 const filters: TableFilter[] = [
   new TextFilter('Product', 'product_name'),
   new DisplayOnlyFilter('Settings'),
-  new SortOnlyFilter('Modified (D/M/Y)', 'updated_at'),
+  new SortOnlyFilter('Modified (Y/M/D)', 'updated_at'),
   new NumberFilter('Benchmark', 'current_benchmark'),
   new NumberFilter('Eligibility Threshold', 'current_eligibility_threshold'),
   new YesNoFilter('Allocation of Emissions', 'requires_emission_allocation'),
--- a/app/containers/Products/ProductListContainer.tsx
+++ b/app/containers/Products/ProductListContainer.tsx
@@ -23,7 +23,7 @@ interface Props {
 const filters: TableFilter[] = [
   new TextFilter('Product', 'product_name'),
   new DisplayOnlyFilter('Settings'),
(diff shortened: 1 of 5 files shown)


app/containers/Products/ProductListContainer.tsx
```diff
@@ -23,7 +23,7 @@ interface Props {
 const filters: TableFilter[] = [
   new TextFilter('Product', 'product_name'),
   new DisplayOnlyFilter('Settings'),
-  new SortOnlyFilter('Modified (D/M/Y)', 'date_modified'),
+  new SortOnlyFilter('Modified (D/M/Y)', 'updated_at'),
```
pbastia, line 26: We're actually displaying the date in the DD-MM-YYYY format here. Should we reflect it in the column header?
`Modified (D-M-Y)` or `Modified (DD-MM-YY)` since there isn't all that much horizontal space?
@rajpersram @kriscooke any opinions? This is slightly outside the scope of this bugfix, we could also leave it as is

pbastia (approved):

app/containers/Products/ProductListContainer.tsx
kriscooke: > This is slightly outside the scope of this bugfix
Yes 🤓 

That said, I don't think what's listed here is a problem anyway. 

It still conveys that the first part is the day, then month, then year. Regular users aren't aware of technical implementation details that we often encounter (like the difference between the result if you either one of those into Moment, for example).

dleard: I don't think it's an issue since we're not actually asking for them to write anything here. In my mind this is just here to be explicit about what the displayed numbers are representing, ie: `2/4/2019` <-- is that Feb 4th or April 2nd? Day-Month-Year is more common, but Month-Day-Year is a thing, so that's what the (D-M-Y) is trying to disambiguate.

matthieu-foucault: >The Government of Canada recommends that all-numeric dates in both English and French use the YYYY-MM-DD format codified in ISO 8601

naomiaro: I've always liked that Canada does YYYY-MM-DD...

dleard: It's an easy change, I'll change it to Y/M/D

pbastia (approved):

PR #1629, opened 2021-04-26
fix: facilityBcghgid includes and enum sort false

naomiaro:
https://youtrack.button.is/issue/GGIRCS-2407
- enum table filters not sortable

https://youtrack.button.is/issue/GGIRCS-2409
- facilityBcghgid uses includes postgraphile filter

### DIFF
--- a/app/components/FilterableTable/Filters/EnumFilter.tsx
+++ b/app/components/FilterableTable/Filters/EnumFilter.tsx
@@ -1,14 +1,19 @@
 import {getUserFriendlyStatusLabel} from 'lib/text-transforms';
 import React from 'react';
 import {Form} from 'react-bootstrap';
-import TableFilter from './TableFilter';
+import TableFilter, {ISearchOptionSettings} from './TableFilter';
 import {FilterComponent} from './types';
 
 export default class EnumFilter<T> extends TableFilter {
   enumValues: T[];
 
-  constructor(display, argName, enumValues: T[]) {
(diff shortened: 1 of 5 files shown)


app/components/FilterableTable/Filters/EnumFilter.tsx
```diff
@@ -8,7 +8,7 @@ export default class EnumFilter<T> extends TableFilter {
   enumValues: T[];
 
   constructor(display, argName, enumValues: T[]) {
-    super(display, argName);
+    super(display, argName, {sortable: false});
```
pbastia, line 11: This is actually breaking the functional sort option on all the other tables using an `EnumFilter`, we just want to disable it for this particular field [here](https://github.com/bcgov/cas-ciip-portal/blob/develop/app/containers/Facilities/FacilitiesListContainer.tsx#L32)

pbastia (changes requested):

app/components/FilterableTable/Filters/EnumFilter.tsx
naomiaro: ah OK, wasn't sure if some enums should be sorted. I think I'll have to export this `ISearchOptionSettings` interface then since there's a few layers of inheritance going on here.

pbastia (approved):

matthieu-foucault (approved):
Approving after rebase

PR #1631, opened 2021-04-27
facility bcghgid import uses new swrs.facility.bc_ghg_id column

matthieu-foucault:
the new column was added in https://github.com/bcgov/cas-ggircs/pull/241


schema/revert/swrs_functions/import_swrs_organisation_facility.sql
```diff
@@ -28,9 +28,9 @@ begin
            business_legal_name, cra_business_number from swrs.organisation
     join latest_reports on report_id = latest_reports.id
     join swrs.report on organisation.report_id = report.id
-  ) on conflict(swrs_report_id) do update
+  ) on conflict(swrs_organisation_id) do update
```
dleard, line 31: What is the benefit of changing the `on conflict` here and below for facility? Is it the idea that swrs_organisation(facility)_id is a more reliably unique anchor than the swrs_report_id?

dleard (approved):
Looks good to me, just a question about the `on conflict` change

schema/revert/swrs_functions/import_swrs_organisation_facility.sql
matthieu-foucault: maybe? That in the revert file, so that was actually changed last time that function was reworked, so I don't remember

dleard: Ah I see, it looked like a change from this PR to me & looked random, my bad. And in fact git blame tells me that I changed that ~ a year ago. I assume I had a good reason :smile:

PR #1632, opened 2021-04-27
feat: migration removing orgs and facilities with no swrs equivalent

pbastia:
this is the CIIP part of the test swrs data making its way in the database,

the PR introduces a migration that deletes the facilities and organisations from the ggircs_portal schema, where the `swrs_..._id` value is set but doesn't exist in the swrs schema anymore.

If an application exist for any of these facilities, this will make the script blow up - it'll be up to us to see if that application is relevant and what to do with it.


schema/deploy/database_functions/prune_orphan_swrs_orgs_and_facilities.sql
```diff
@@ -0,0 +1,46 @@
+-- Deploy ggircs-portal:database_functions/prune_orphan_swrs_orgs_and_facilities to pg
+-- requires: schema_ggircs_portal_private
+
+-- This function looks at all the ciip portal organisations and facilities, and removes them if their swrs corresponding entry is not found.
+-- Items which don't have a swrs id were added in CIIP and should be kept.
```
matthieu-foucault, line 5: use a `comment on function` so that this is visible in the DB as well

pbastia: Excellent point 👌

matthieu-foucault (approved):

PR #1635, opened 2021-04-28
feat: add computed columns to retrieve form_result data

dleard:
Adds computed columns on application_revision for: emission, fuel & production form_result data.
Needed for several of the validation tasks coming up.


schema/verify/computed_columns/application_revision_fuel_form_data.sql
```diff
@@ -0,0 +1,7 @@
+-- Verify ggircs-portal:computed_columns/application_revision_fuel_form_data on pg
+
+BEGIN;
+
+-- XXX Add verifications here.
```
matthieu-foucault, line 5: missing verify

schema/verify/computed_columns/application_revision_production_form_data.sql
```diff
@@ -0,0 +1,7 @@
+-- Verify ggircs-portal:computed_columns/application_revision_production_form_data on pg
+
+BEGIN;
+
+-- XXX Add verifications here.
```
matthieu-foucault, line 5: missing verify

schema/verify/types/emission_form_data.sql
```diff
@@ -0,0 +1,13 @@
+-- Verify ggircs-portal:types/emission_form_data on pg
+
+begin;
+
+do $$
+  begin
+    assert (
+      select true from pg_catalog.pg_type where typname = 'emission_form_data'
```
matthieu-foucault, line 8: I had a doubt, but the column is indeed `typname`, and not `typename`. That's an interesting choice from postgres...

matthieu-foucault (changes requested):
Maybe add a TD card to refactor the ciip views so that they use these computed columns (not sure if our readonly user is allowed to execute functions, so that might be worth checking when doing that TD card)

dleard:
Tech debt card for re-using computed columns in the ciip_* views: https://youtrack.button.is/issue/GGIRCS-2422

matthieu-foucault (approved):
:shipit:

PR #1646, opened 2021-04-29
feat: create admin-2020 form schema

matthieu-foucault:


schema/deploy/computed_columns/application_revision_naics_code.sql
```diff
@@ -13,7 +13,7 @@ create or replace function ggircs_portal.application_revision_naics_code(app ggi
         from ggircs_portal.form_result
         join ggircs_portal.form_json
         on form_result.form_id = form_json.id
-        and form_json.slug in ('admin', 'admin-2018')
+        and form_json.slug in ('admin', 'admin-2018', 'admin-2020')
```
pbastia, line 16: it looks like the `admin` slug is becoming the 2019 form for ever
I wonder if we should rename it as well, and maybe make "admin" behave like an alias through a computed column or something, to be the current year (or latest?) form in the system
Anyways, thoughts for tech debt :)

pbastia: actually, the slug could be a computed column built from form type + year

matthieu-foucault: I thought about it, but starting with this PR the slug is what is used in the `on conflict` clause when inserting the form schemas in the database, so we can't change it, unless we would change it live in prod just before deploying the release.

We could add some metadata to this table though, or update the form_json.description so that it actually says something useful :)

schema/test/unit/tables/ciip_application_wizard_test.sql
```diff
@@ -74,6 +50,15 @@ select lives_ok(
     'ciip_administrator can insert data in ciip_application_wizard table'
 );
 
+select results_eq(
+  $$
+    select count(*) from ggircs_portal.ciip_application_wizard;
+  $$,
+  ARRAY[1::bigint],
```
pbastia, line 57: we started using the `values ( )` syntax for these tests since `ARRAY[]` has compatibility issues with some types - although the only issues we reported where either with boolean/text conversion and multiple types. So this one should be fine regardless?

```suggestion
$$
    values (1::bigint)
$$
```

schema/test/unit/tables/form_json_test.sql
```diff
@@ -44,7 +49,7 @@ select results_eq(
   $$
     select count(*) from ggircs_portal.form_json
   $$,
-  ARRAY[4::bigint],
+  ARRAY[5::bigint],
```
pbastia, line 52: ditto in this file

pbastia (approved):
One comment about `ARRAY` that doesn't really matter, looks good to me otherwise!

schema/deploy/computed_columns/application_revision_naics_code.sql
pbastia: true - we do have a 'description' column in that table that [we don't use for anything useful](https://github.com/bcgov/cas-ciip-portal/blob/f4ebd3c7df002b302f925118ad186a00eda5f6af/schema/data/prod/form_json.sql#L32)...yet!

PR #1651, opened 2021-05-03
Feat: no reported ciip product validation

dleard:
validates that the application contains at least one product where:
`isCiipProduct && !isEnergyProduct && productionAmount > 0`


schema/deploy/application_validation_functions/no_reported_ciip_product.sql
```diff
@@ -0,0 +1,34 @@
...
+$$ language sql stable;
+
+grant execute on function ggircs_portal.no_reported_ciip_product to ciip_administrator, ciip_analyst, ciip_industry_user;
+
+comment on function ggircs_portal.no_reported_ciip_product(ggircs_portal.application_revision) is 'This validation function for a CIIP (CleanBC Industrial Incentive Program) application determines if it contains at least one valid ciip product reported';
+
+insert into ggircs_portal.application_revision_validation_function(validation_function_name, validation_description, validation_failed_message)
+values (
+  'no_reported_ciip_product',
+  'determines if an application has at least one valid ciip product reported',
```
matthieu-foucault, line 30: The intent of the `validation_description` is to be rendered on the analyst's side when an application is reviewed, so maybe "Application has at least one ciip product reported" would be more appropriate in that context. Other validation migrations might need an update to reflect that too.

schema/deploy/application_validation_functions/no_reported_ciip_product.sql
```diff
@@ -0,0 +1,34 @@
...
+      and pfd.product_amount > 0
+    );
+
+$$ language sql stable;
+
+grant execute on function ggircs_portal.no_reported_ciip_product to ciip_administrator, ciip_analyst, ciip_industry_user;
+
+comment on function ggircs_portal.no_reported_ciip_product(ggircs_portal.application_revision) is 'This validation function for a CIIP (CleanBC Industrial Incentive Program) application determines if it contains at least one valid ciip product reported';
+
+insert into ggircs_portal.application_revision_validation_function(validation_function_name, validation_description, validation_failed_message)
```
matthieu-foucault, line 27: Note that, as we discovered with @kriscooke while looking at #1644 , items that are inserted in the migrations will not be present in e2e tests, as the `test_helper.clean_ggircs_portal_schema` function will empty all the tables.
An sql file in `schema/data/prod` inserting those might be preferable, especially if we want to update those descriptions/failed messages later

kriscooke: Does this mean I should hold off on the proposed fix in my test fixture?

matthieu-foucault: I think so. Adding an script in `schema/data/prod` might be a better solution for cases where some table records are required for features that we are developing (advanced validation functions, application review steps, form schemas). Inserts in the sqitch migrations seem to be a better option for seed data (e.g. the list of naics codes) that can be modified later by a user of the application. 

I'll add a follow up to my comment in #1644

naomiaro:
I'm not really sure why Happo is adding 5 screenshots, but otherwise lgtm.

naomiaro (approved):

matthieu-foucault:
@naomiaro Happo is adding screenshots because of our flaky e2e test, which failed and got retried by cypress

schema/data/deploy-data.sh
```diff
@@ -192,6 +192,7 @@ deployProdData() {
   _psql -f "./prod/emission.sql"
   _psql -f "./prod/gas.sql"
   _psql -f "./prod/emission_gas.sql"
+  _psql -f "./prod/application_validation_function.sql"
```
matthieu-foucault, line 195: This needs to be added to `deployTestData` below, as that is the one that is actually run by cypress

matthieu-foucault (changes requested):
Just one small change and that should be good to go

matthieu-foucault (approved):
:shipi

PR #1652, opened 2021-05-03
Feat/incentive ratio

naomiaro:
https://youtrack.button.is/issue/GGIRCS-2328

- Incentive ratio column should be on the right of the 'payment allocation' column
- Threshold chart column is removed
- 'Calculated incentive' and 'Maximum incentive' columns use the English currency format ( $1,234,567.89 )

- Aggregate Total row is added for multiple products

Table with zero products:
Screenshot 2021-05-03 at 2 39 11 PM

Table with one product:
Screenshot 2021-05-03 at 2 39 50 PM

Table with two products:
Screenshot 2021-05-03 at 2 40 05 PM


app/tests/unit/containers/Incentives/IncentiveCalculatorContainer.test.tsx
```diff
@@ -27,6 +57,15 @@ describe('IncentiveCalculator', () => {
...
 
+  it('includes an aggregate row with multiple products', async () => {
+    const r = shallow(
+      <IncentiveCalculator
+        applicationRevision={applicationRevisionMultiProduct}
+      />
+    );
+    expect(r).toMatchSnapshot();
+  });
+
```
matthieu-foucault, line 68: There are a bunch of `todo` tests at the bottom of this file, which are all outdated. Let's remove them.

app/tests/unit/containers/Incentives/IncentiveCalculatorContainer.test.tsx
```diff
@@ -27,6 +57,15 @@ describe('IncentiveCalculator', () => {
     expect(r).toMatchSnapshot();
   });
 
+  it('includes an aggregate row with multiple products', async () => {
+    const r = shallow(
+      <IncentiveCalculator
+        applicationRevision={applicationRevisionMultiProduct}
+      />
+    );
+    expect(r).toMatchSnapshot();
```
matthieu-foucault, line 66: the snapshot matches are more of a "catch all" which would show us unexpected changes. Our unit tests should have some stronger assertions to complement them: in this case checking that the displayed totals match the expected value

matthieu-foucault (changes requested):

app/containers/Incentives/IncentiveCalculatorContainer.tsx
```diff
@@ -13,6 +14,30 @@ export const IncentiveCalculator: React.FunctionComponent<Props> = ({
...
+  );
+
+  // incentive ratio = calculated incentive / maximum incentive
+  // set the default to 1 to avoid division by zero.
+  let aggIncentiveRatio = 1;
+  if (aggTotals.incentiveProductMax > 0) {
+    aggIncentiveRatio =
+      aggTotals.incentiveProduct / aggTotals.incentiveProductMax;
+  }
+
```
matthieu-foucault, line 40: Double-checked the math to make sure that this is equivalent to a weighted average. It is.

matthieu-foucault (approved):

PR #1654, opened 2021-05-04
test: remove extra flaky instruction in products-benchmark e2e test

matthieu-foucault:


app/cypress/integration/product-benchmark.spec.js
```diff
@@ -59,26 +59,21 @@ describe('The products and benchmark page', () => {
     });
 
     cy.get('#root_benchmark').clear().type('12');
-    cy.get('.rjsf > .btn').contains('Save');
-    cy.get('.rjsf > .btn').click();
+    cy.get('.rjsf > .btn').contains('Save').click();
     cy.get('tbody > :nth-child(1) > :nth-child(4)').contains('12');
```
pbastia, line 63: should we check that the edit modal is closed, to avoid race conditions?
```javascript
cy.get('.modal').should('not.exist');
```

pbastia (approved):

app/cypress/integration/product-benchmark.spec.js
matthieu-foucault: https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Covering cypress should be checking that already, so I think we're good here

PR #1655, opened 2021-05-04
Feat: Missing indices on foreign keys

dleard:
- pgTap test which checks that all foreign keys have an index
- migration to create indices for all foreign keys that are currently missing one


schema/revert/database_functions/add_missing_indices.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert ggircs-portal:database_functions/add_missing_indices from pg
+
+BEGIN;
+
+-- XXX Add DDLs here.
+
+COMMIT;
```
matthieu-foucault, line 7: Do you want to write a revert?

dleard: No, I don't really want to delete all the indices that were created, the deploy file is idempotent anyway. I'll add a comment here and to the verify file that they are not needed for this migration

matthieu-foucault: sounds good to me

matthieu-foucault (approved):

matthieu-foucault (approved):

PR #1657, opened 2021-05-05
Feat: validation that mandatory products are reported

pbastia:
Validation function that makes sure that all mandatory products for an application's NAICS code are reported on the production form

image


schema/deploy/application_validation_functions/mandatory_products_are_reported.sql
```diff
@@ -0,0 +1,41 @@
...
+
+insert into ggircs_portal.application_revision_validation_function(
+  validation_function_name,
+  validation_description,
+  validation_failed_message)
+values
+  ('mandatory_products_are_reported',
+  'Validates that for an application revision, all mandatory products defined for the NAICS code of that application are reported',
+  'Some mandatory products have not been reported.'
+  );
```
matthieu-foucault, line 39: We established that those inserts should be in `schema/data/prod`, see https://github.com/bcgov/cas-ciip-portal/pull/1651#discussion_r625401294

matthieu-foucault: Also
- the `validation_description` will be shown to the analyst in the pass/failed checklist, so maybe something shorter, e.g. "All mandatory products are reported", would be better
- The warning message on the other hand could be more verbose, e.g. "Some products which are mandatory for this application's NAICS code have not been reported. Please go to the "Production" form to report any missing products."

matthieu-foucault (approved):
:shipit:

PR #1658, opened 2021-05-05
feat: use difference with march 2018 fuel charge to get CIIP CT

matthieu-foucault:


schema/test/unit/views/ciip_carbon_tax_calculation_test.sql
```diff
@@ -1,68 +1,89 @@
...
-      "fuelRowId": 11,
-      "quantity": 400,
-      "fuelUnits": "kL",
-      "emissionCategoryRowId": 2
-    }
-  ]'
-  where form_id = 3
-  and application_id = 1
-  and version_number = 1;
+set form_result = json_build_array(
```
dleard, line 26: :+1: nice. I hadn't thought to use json_build_* functions to create the form results

dleard (approved):
looks good!

PR #1663, opened 2021-05-06
Feat: validation that carbon-taxed fuels match in ciip and swrs

pbastia:
This PR creates a validation function that compares ciip and swrs reported fuel quantities, only on carbon-taxed fuels

to do this:
- retrieves the carbon-taxed fuels for both swrs and ciip (with the help of a new `fuel_is_carbon_taxed` computed column)
- groups by fuel ID and sums the fuel quantities reported
- "full outer join" these sums in a table on fuel ID
- makes sure we don't have null values on either side
- makes sure the difference in percentage doesn't exceed 0.1% on any fuel

the difference percentage is calculated with the absolute value of:
`diff(x,y) = 2 * (x - y) / (x + y)`


schema/deploy/computed_columns/fuel_is_carbon_taxed.sql
```diff
@@ -0,0 +1,18 @@
...
+  as $$
+    select exists(
+      select 1
+      from swrs.fuel_mapping fm
+      join swrs.fuel_carbon_tax_details fctd
+        on fm.fuel_carbon_tax_details_id = fctd.id
+        and fctd.carbon_tax_act_fuel_type_id is not null
+      where f.swrs_fuel_mapping_id = fm.id
+    );
+  $$ language sql stable;
```
matthieu-foucault, line 17: missing a `grant execute statement` and a `comment`

schema/test/unit/application_validation_functions/carbon_taxed_fuels_match_rev_zero_test.sql
```diff
@@ -0,0 +1,324 @@
...
+('[
+  {
+    "fuelRowId": '|| (select id from ggircs_portal.fuel where name = 'carbonTaxed1') || ',
+    "quantity": 10
+  },
+  {
+    "fuelRowId": '|| (select id from ggircs_portal.fuel where name = 'carbonTaxed2') || ',
+    "quantity": 100
+  }
+]')::jsonb where application_id = 1 and version_number = 0 and form_id = (select id from ggircs_portal.form_json where slug='fuel');
```
matthieu-foucault, line 92: FYI, you can make this a bit easier to read and robust with `json_build_array` and `json_build_object`

matthieu-foucault (changes requested):

matthieu-foucault (changes requested):
Fuels reported under the venting or flaring emission categories should be ignored by that validation warning.
They are not reported in SWRS but have to be reported in CIIP

PR #1667, opened 2021-05-10
Feat: analyst validation checks

pbastia:
Refactor of the analyst display for application validation, in a component that aggregates the override justification and the automatic application validation.

image

image


app/containers/Applications/ApplicationReviewValidationContainer.tsx
```diff
@@ -0,0 +1,154 @@
...
+  return (
+    <Row className="mt-3">
+      <Col>
+        <ListGroup>
+          <ListGroup.Item variant={validationVariant}>
+            <Row>
+              <Col>
+                <FontAwesomeIcon icon={faExclamationTriangle} fixedWidth />
+                &nbsp;
+                {failedValidationCount === totalValidationCount
```
kriscooke, line 61: This comparison is only ever true if all the validation checks fail, instead of pass.

app/pages/analyst/application-review.tsx
```diff
@@ -59,14 +59,14 @@ class ApplicationReview extends Component<Props, State> {
...
             versionNumber
-            overrideJustification
             isCurrentVersion
             applicationRevisionStatus {
               id
               applicationRevisionStatus
             }
             ...ApplicationDetailsContainer_applicationRevision
             ...IncentiveCalculatorContainer_applicationRevision
+            ...ApplicationReviewValidationContainer_applicationRevision
```
kriscooke, line 69: This is hard to notice (because it's just a TS error, the tests still pass), but the test query needs to be updated in `application-review.test.tsx` anytime we change the query on the page.

app/tests/unit/containers/Applications/ApplicationReviewValidationContainer.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+            }
+          },
+          {
+            node: {
+              isOk: true,
+              validationDescription: 'this is a passing validation'
+            }
+          }
+        ]
+      }
```
kriscooke, line 32: To avoid copying and pasting test queries into multiple tests, then having to carefully update every one of them if and when the fragment query changes, in exactly the right way while making sure it still tests the intended case, in all of the new [application review components I've been using a function for this](https://github.com/bcgov/cas-ciip-portal/blob/d36695f4b2517d41a62a695b6c55ccc937f2f353/app/tests/unit/containers/Admin/ApplicationReview/ApplicationReviewStepSelector.test.tsx#L7):
```
const data = getTestQuery(options);
```

For components that have many props, I've also outsourced instantiation of the test element to a function.

kriscooke (changes requested):
(🤔 Weird - I added those comments above in batch as part of a review, but they went ahead and got submitted already..)

Anyway, looks great! There's just a couple issues with displaying the case where all checks pass, and some suggestions for reducing copy pasting for test queries.

kriscooke (approved):

PR #1671, opened 2021-05-11
Chore: improved dev data

dleard:
Update the deploy-data script to handle new dev data, both on openshift and locally.
- [x] dev/load_testing data scripts added
- [x] updated deploy_data.sh to use deployDevData() and deployPgTapData() as needed
- [x] Mock function so that lastSwrsReportingYear() doesn't always return 2025 when swrs dev data is deployed
- [x] Add applications for operator names that define the state (draft operator, submitted operator, requested changes operator)
- [x] Add a cronjob to deploy the load-testing data


helm/cas-ciip-portal/templates/cronJobs/cron-load-testing-data.yaml
```diff
@@ -0,0 +1,74 @@
...
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ template "cas-ciip-portal.fullname" . }}-load-testing-data
+  labels:
+{{ include "cas-ciip-portal.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": "pre-upgrade,pre-install"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
+    "helm.sh/hook-weight": "-5"
```
matthieu-foucault, line 11: Does this need to be a hook?

helm/cas-ciip-portal/templates/cronJobs/cron-load-testing-data.yaml
```diff
@@ -0,0 +1,74 @@
...
+                requests:
+                  cpu: 100m
+                  memory: 64Mi
+              command:
+                - /usr/bin/env
+                - bash
+                - -c
+                - |
+                  set -euo pipefail;
+                  schema/data/deploy-data.sh --oc-project=$(NAMESPACE);
```
matthieu-foucault, line 72:
```suggestion
                  schema/data/deploy-data.sh --oc-project=$(NAMESPACE) --swrs-load-testing;
```
?

schema/data/dev/create_dev_applications.sql
```diff
@@ -0,0 +1,94 @@
...
+          (
+            select application_open_time
+            from ggircs_portal.reporting_year
+            where reporting_year = mocked_year
+          )
+        );
+
+        if mocked_year = 2020 then
+          alter table ggircs_portal.application_revision_status enable trigger _create_or_refresh_review_step;
+        end if;
```
matthieu-foucault, line 34: Why is 2020 a special case?

dleard: This is an attempt to create application_review_steps in the same way as they will appear in prod. So in the loop, for the years 2018 & 2019 there we create the the generic `application` review step and all years after 2020 will create review_steps from the table (currently `administrative` & `technical`) via the trigger. The trigger gets disabled at the beginning of the file & re-enabled when year=2020 in the loop (staying enabled for the rest of the following years)

helm/cas-ciip-portal/templates/cronJobs/cron-load-testing-data.yaml
dleard: This shouldn't need any flags, if we deploy the load-testing data in cas-ggircs & then redeploy the ciip data it should pull the load testing data from the cas-ggircs swrs database. `--swrs-load-testing` is for local development if we wanted to go look at it locally

helm/cas-ciip-portal/templates/cronJobs/cron-load-testing-data.yaml
dleard: Probably not? I just copied cron-deploy-data.yaml and changed a couple things

schema/data/dev/create_dev_applications.sql
pbastia: maybe let's make it `if mocked_year >= 2020` this way it's clear to the developer, and it has the same effect?

dleard: could do, it will just run the `alter table...enable trigger...` statement a few more unnecessary times that way, but if it's clearer then the impact of that is very minimal

helm/cas-ciip-portal/templates/cronJobs/cron-load-testing-data.yaml
dleard: This basically replaces the cron-deploy-data step in the dag, so I just copied it and changed the command/name

schema/data/dev/create_dev_applications.sql
dleard: I did add a comment above that line to try and be clear about what is happening

pbastia (approved):

PR #1674, opened 2021-05-12
chore: add sysdig links/monitoring to shipit

dleard:
I think the iframe will have an issue unless we make the dashboard public in sysdig?

### DIFF
--- a/shipit.dontdeployme.yml
+++ b/shipit.dontdeployme.yml
@@ -1,3 +1,17 @@
+dependencies:
+  override: []
+deploy:
+  override:
+    - echo 'Not to be deployed'
 review:
+  checklist: []
   monitoring:
     - iframe: https://app.sysdigcloud.com/#/dashboards/222088?last=3600
+      width: 400
+      height: 300
+ci:
+  hide:
+    - "ci/circleci: release/release_approval"
+links:
+  PVC_usage: https://app.sysdigcloud.com/#/dashboards/224524?last=3600
+  Health_overview: https://app.sysdigcloud.com/#/dashboards/225040?last=3600
--- a/shipit.dontdeployme.yml
+++ b/shipit.dontdeployme.yml
@@ -1,0 +1,3 @@
+review:
+  monitoring:
+    - iframe: https://app.sysdigcloud.com/#/dashboards/222088?last=3600
--- a/shipit.dontdeployme.yml
+++ b/shipit.dontdeployme.yml
@@ -7,8 +7,8 @@ review:
   checklist: []
   monitoring:
     - iframe: https://app.sysdigcloud.com/#/dashboards/222088?last=3600
-      width: 400
-      height: 300
+      width: 1200
+      height: 800
 ci:
   hide:
     - "ci/circleci: release/release_approval"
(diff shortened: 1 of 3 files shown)


matthieu-foucault:
Let's merge and see what it looks like :smile:

matthieu-foucault:
image

matthieu-foucault:
See https://cas-shipit.apps.silver.devops.gov.bc.ca/bcgov/cas-ciip-portal/dontdeployme

matthieu-foucault:
The links don't appear to be showing. I was already logged in sysdig, so the monitoring shows up. We just need to figure out how to make it bigger

matthieu-foucault:
Nevermind, the links appeared after shipit synchronized. Looks like the PVC usage dashboard wasn't share with all teams though

matthieu-foucault:
See https://github.com/Shopify/shipit-engine/blob/236ffbfefb55ffc5faaa2ec27f72bb173106b3d4/examples/shipit.yml#L17 to set the iframe width and height

matthieu-foucault:
I don't think that we need the monitoring to show up in anything else than prod. I'm not sure how the environment-specific yml files are handled though. Maybe try adding a `shipit.dontdeployme.yml`, which only has the monitoring settings, to see if it merges it with the base `shipit.yml`

dleard:
Cool thanks, I'll play with the iframe & PVC dashboard. Did you create a new shipit stack and just point it to this branch to make the 'dontdeployme'?

matthieu-foucault:
Yes I created a shipit stack, that was the easiest to test this.

dleard:
shipit.dontdeployme,yml overrides the default shipit.yml, it does not merge with it.
The iframe has been enlarged but it has some permission / data issues. I set the dashboard to be shared with all teams but if you are not logged into the CIIP namespace's sysdig team the data will show up as `No data` because the other namespaces can't find those pods

dleard:
@matthieu-foucault Per your question in stand-up today: Yes, you can log in to sysdig from the iframe if you're not already logged in

shipit.dontdeployme.yml
```diff
@@ -0,0 +1,17 @@
+dependencies:
+  override: []
+deploy:
+  override:
+    - echo 'Not to be deployed'
```
matthieu-foucault, line 5: Looks good. Let's just copy what's in the default shipit.yml and rename this to `shipit.prod.yml`

shipit.dontdeployme.yml
```diff
@@ -0,0 +1,17 @@
...
+review:
+  checklist: []
```
matthieu-foucault, line 7:
```suggestion
  checklist:
    - Is today not Friday?
```

matthieu-foucault (changes requested):

shipit.prod.yml
```diff
@@ -0,0 +1,20 @@
...
+        timeout: 2460
+review:
+  checklist:
+    - Is today not Friday?
+  monitoring:
+    - iframe: https://app.sysdigcloud.com/#/dashboards/222088?last=3600
+      width: 1200
+      height: 800
+ci:
+  hide:
```
matthieu-foucault, line 16:
```suggestion
  require:
```
the release approval is actually required for prod (and test)

matthieu-foucault (changes requested):

matthieu-foucault (approved):

PR #1679, opened 2021-05-18
Initialize and enforce mandatory products on reporter side

kriscooke:
- If a NAICS code has been set in the Administrative Data section, any mandatory products are initialized in the Production section.
- Adds custom ProductsArrayField to accomplish this by mutating the formResult on load
- "Remove product" button is overridden on a per-item basis for mandatory products

[GGIRCS-2288](https://youtrack.button.is/issue/GGIRCS-2288)

### DIFF
--- a/app/containers/Forms/Form.tsx
+++ b/app/containers/Forms/Form.tsx
@@ -14,6 +14,7 @@ import FormArrayFieldTemplate from './FormArrayFieldTemplate';
 import FuelFields from './FuelField';
 import EmissionGasFields from './EmissionGasFields';
 import EmissionSourceFields from './EmissionSourceFields';
+import ProductsArrayField from './ProductsArrayField';
 import ProductField from './ProductField';
 import EmissionField from './EmissionField';
 import ProductRowIdField from './ProductRowIdField';
@@ -41,6 +42,7 @@ const CUSTOM_FIELDS = {
   fuel: (props) => <FuelFields query={props.formContext.query} {...props} />,
   emissionSource: EmissionSourceFields,
   emissionGas: EmissionGasFields,
+  ProductsArrayField,
   product: (props) => (
     <ProductField
       naicsCode={
--- a/app/containers/Forms/Form.tsx
+++ b/app/containers/Forms/Form.tsx
@@ -44,7 +44,7 @@ const CUSTOM_FIELDS = {
   emissionGas: EmissionGasFields,
   ProductsArrayField: (props) => (
     <ProductsArrayField
-      mandatoryProducts={
+      naicsProducts={
         props.formContext.ciipFormResult.applicationByApplicationId
           .latestDraftRevision.naicsCode
       }
@@ -259,7 +259,7 @@ export default createFragmentContainer(FormComponent, {
           naicsCode {
             ...ProductRowIdField_naicsCode
             ...ProductField_naicsCode
-            ...ProductsArrayField_mandatoryProducts
+            ...ProductsArrayField_naicsProducts
           }
         }
       }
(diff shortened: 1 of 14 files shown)


app/tests/unit/containers/Forms/ProductField.test.tsx
```diff
@@ -32,233 +39,117 @@ describe('The ProductionFields Component with published product matched to a nai
...
-    formData: {
-      productRowId: 1,
-      productUnits: 'bar',
-      productEmissions: 123,
-      productAmount: 456,
-      requiresEmissionAllocation: true
-    },
-    query,
-    naicsCode
-  };
```
kriscooke, line 121: Extracted the copy-pasted test queries and props to a factory function at top so in future we only need to update queries in one place.

schema/data/prod/form_json/production.json
```diff
@@ -87,7 +90,11 @@
   "uiSchema": {
     "ui:add-text": "Add a Product",
     "ui:remove-text": "Remove Product",
+    "ui:field": "ProductsArrayField",
     "items": {
+      "ui:hasRemove": {
+        "isFalse": "isMandatory"
+      },
```
kriscooke, line 97: Hidden field helps us [override the `hasRemove` button boolean property](https://github.com/bcgov/cas-ciip-portal/pull/1679/files#diff-1c740ccc67b9a1928566d1bbe1c2c266c397b5ba4ad53f404fde5614a6a0eb5aR19-R28) for array fields on a per-item basis.

app/tests/unit/containers/Forms/ProductsArrayField.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+describe('ProductsArrayField', () => {
+  it('no NAICS code specified', () => {
+    const wrapper = shallow(getTestElement({noNaicsCode: true}));
+    expect(wrapper.find('Relay(ProductField)').exists()).toBeFalse();
+    expect(wrapper).toMatchSnapshot();
+  });
+  it('NAICS code specified, with allowable products', () => {
+    const wrapper = shallow(getTestElement({}));
+    expect(wrapper).toMatchSnapshot();
+  });
```
kriscooke, line 62: It would be nice to add more assertions... Since the component relies on there being a formResult added to the formContext prop, we'd need to modify the `createDefaultJsonSchemaFormProps` helper in json-schema-utils

app/containers/Forms/Form.tsx
```diff
@@ -235,6 +245,7 @@ export default createFragmentContainer(FormComponent, {
   `,
   ciipFormResult: graphql`
     fragment Form_ciipFormResult on FormResult {
+      id
```
matthieu-foucault, line 248: is this needed?

matthieu-foucault (changes requested):
The `ProductRowIdField` of mandatory products should be disabled/readonly, as the user is  currently able to change the product type of a mandatory product.

app/tests/unit/containers/Forms/ProductsArrayField.test.tsx
pbastia: Using `mount` instead of `shallow` we should be able to test the useEffect function, and spy on the mutation?

We could always pass in a formContext with someting like this
```javascript
const props = {...createDefaultJsonSchemaFormProps(), formContext: {ciipFormResult: {...etc...}}}
```

pbastia (commented):
Very nice use of Relay and Jsonschemaform! 👌
I'll leave @matthieu-foucault to approve since he is primary reviewer on this

app/containers/Forms/Form.tsx
kriscooke: Yes - unfortunately the dependency is a bit implicit, but it's passed in formContext as part of the formResult and the [id is used for the updateFormResultMutation there](https://github.com/bcgov/cas-ciip-portal/pull/1679/files#diff-65ea7c4e75aee861324fe97344b52cb3a194edca1bbfecd6fb835747d0d01c8eR49).

matthieu-foucault (changes requested):
On small update, and this should be good to merge

matthieu-foucault (dismissed):

matthieu-foucault (approved):

PR #1690, opened 2021-05-20
Fix: show legacy form json

dleard:
- Adds a pgTap test for `application_ordered_form_results()`
- Updates `data/prod/ciip_application_wizard` to use `on conflict` instead of `truncate..` & adds the inactive `admin` legacy form_json
- Small fixes to dev data to use the `is_active` column when creating dev form results

### DIFF
--- a/schema/data/prod/ciip_application_wizard.sql
+++ b/schema/data/prod/ciip_application_wizard.sql
@@ -1,15 +1,15 @@
 begin;
 
-truncate table ggircs_portal.ciip_application_wizard;
-
 with rows as (
 insert into ggircs_portal.ciip_application_wizard
-  (form_id, form_position)
+  (form_id, form_position, is_active)
 values
-  ((select id from ggircs_portal.form_json where slug='admin-2020'), 0),
-  ((select id from ggircs_portal.form_json where slug='emission'), 1),
-  ((select id from ggircs_portal.form_json where slug='fuel'), 2),
-  ((select id from ggircs_portal.form_json where slug='production'), 3)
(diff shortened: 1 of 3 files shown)


schema/data/prod/ciip_application_wizard.sql
```diff
@@ -1,15 +1,15 @@
 begin;
```
pbastia, line 2: Since we're not truncating anymore, should we make sure to set all the existing forms to inactive before upserting the prod ones?

```suggestion

update ggircs_portal.ciip_application_wizard set is_active=false;

```

schema/test/unit/computed_columns/application_ordered_form_results_test.sql
```diff
@@ -0,0 +1,64 @@
...
+-- Set admin form_result with application_id=2 to point to a deprecated form_json
+update ggircs_portal.form_result set form_id=1 where application_id=2 and form_id=5;
+
+select id, application_id, form_id from ggircs_portal.form_result;
+
+select id, slug from ggircs_portal.form_json;
+select * from ggircs_portal.ciip_application_wizard;
+
+with record as (select row(application.*)::ggircs_portal.application from ggircs_portal.application where id=2)
+    select form_id from ggircs_portal.application_ordered_form_results((select * from record), '1') order by form_id;
```
pbastia, line 42: this looks like debug stuff 😄

pbastia (changes requested):

pbastia (approved):

PR #1692, opened 2021-05-21
Remove defunct 'activate org access' button on localhost

kriscooke:
This has long been defunct since the implementation of row-level security, and only serves to confuse new team members.

Screen Shot 2021-05-20 at 5 49 07 PM


naomiaro (approved):
LOL! I definitely was confused! 😂

matthieu-foucault:
FYI this button only works when using the NO_AUTH flag, which I don't believe devs have been using anymore

kriscooke:
Yeah, and we should probably remove `NO_AUTH` as well. There's no real value in impersonating every user role at once when we have the other convenient flags, we got bitten by that early on.

PR #1695, opened 2021-05-21
fix: perform home page redirection on server to prevent content flashing

matthieu-foucault:


pbastia (dismissed):
Needs rebase

very nice!

app/pages/index.tsx
```diff
@@ -7,12 +7,28 @@ import DefaultLayout from 'layouts/default-layout';
...
+  if (landingRoute === '/')
+    return {
+      props: {}
+    };
+
+  return {
+    redirect: {destination: landingRoute, permanent: false}
+  };
+}
+
```
kriscooke, line 31: I'm pretty sure the routing needs to be done in `server/index.js` - even using getServerSideProps, this only runs when you are on the homepage.

So if I'm on an application page for example and click "Dashboard", I don't hit this until I first load the homepage on the client.

kriscooke (changes requested):

app/pages/index.tsx
kriscooke: (in other words, using getServerSideProps only makes the homepage load faster, and thus perform the redirect a bit faster)

matthieu-foucault: `getServerSideProps` is run by the next middleware on the server side, so this should work

kriscooke: Note to self: This solution doesn't work in development when using the `AS_REPORTER` etc. flags, which won't be the case on production.

kriscooke (approved):

PR #1696, opened 2021-05-21
Fix: filter by facility type

dleard:
- Facility Type was missing from the query params
- Changed the filter from text to enum [SFO, IF_a, IF_b, L_c]


app/components/FilterableTable/Filters/EnumFilter.tsx
```diff
@@ -48,7 +48,9 @@ export default class EnumFilter<T> extends TableFilter {
               key={this.argName + '-' + kvp.display}
               value={kvp.value as any}
             >
-              {getUserFriendlyStatusLabel(kvp.display)}
+              {this.ignoreFriendlyDisplay
+                ? kvp.display
+                : getUserFriendlyStatusLabel(kvp.display)}
             </option>
```
pbastia, line 54: For better OOP and future reuse, instead of a boolean we could allow the user to supply the implementation of the transformation function - in this case with the identity function.

And since this only affects the behaviour of the EnumFilter we probably shouldn't modify the parent `TableFilter` class ("open for extension, closed for modification")

I'd do something like that 
```javascript
export interface IEnumFilterOptionSettings<T> extends ISearchOptionSettings {
  renderEnumValue?(label: T): string;
}
export default class EnumFilter<T> extends TableFilter {
  enumValues: T[];

  constructor(
    display,
    argName,
    enumValues: T[],
    settings?: IEnumFilterOptionSettings
  ) {
    super(display, argName, settings);
    this.renderEnumValue =
      settings.renderEnumValue ?? getUserFriendlyStatusLabel;
      
      ...
```
and then instantiate with the identity
```javascript
new EnumFilter<String>(
    'Facility Type',
    'facilityType',
    ['SFO', 'IF_a', 'IF_b', 'L_c'],
    {renderEnumValue: (label) => label, sortable: false}
  )
```

pbastia (changes requested):
nice! I think we should at least move the `ignoreFriendlyDisplay` to the EnumFilter properties since it's not used in the base class.

app/components/FilterableTable/Filters/TableFilter.tsx
```diff
@@ -6,6 +6,7 @@ export interface ISearchOptionSettings {
   filterable?: boolean;
   sortable?: boolean;
   sortColumnName?: string;
+  renderEnumValue?: (x: string) => string;
```
pbastia, line 9: I'd still want to move this to the `EnumFilter` class instead, since this is specific to how enums are displayed

dleard: ah, I left it in there to fix a typescript error on the settings object. I'll add a new interface inside EnumFilter and extend IsearchOptionSettings there with renderEnumValue

pbastia (approved):
:shipit: very nice

PR #1700, opened 2021-05-25
fix: uncomment admin & analyst a11y tests

naomiaro:
https://youtrack.button.is/issue/GGIRCS-1743

Collection of a11y fixes


app/containers/Products/ProductRowItemContainer.tsx
```diff
@@ -147,6 +148,7 @@ export const ProductRowItemComponent: React.FunctionComponent<Props> = ({
       centered
       size="xl"
       show={benchmarkModalShow}
+      aria-labelledby="product-benchmark-header"
```
pbastia, line 151: We probably want to put this identifier (and the one just above) in a static export from `InnerProductBenchmarkModal`, so we don't have to worry about this changing in one or the other place

pbastia (changes requested):
Just one small thing about a magic string that lives across 2 files, looks great otherwise!

PR #1701, opened 2021-05-25
fix: excluding fuels reporting a zero quantity from validation

pbastia:
`carbon_taxed_fuels_match_rev_zero` throws a division by zero error if ciip or swrs report a zero quantity fuel.
This fixes it by removing this case from the `join` statement that combines both swrs and ciip data


dleard (dismissed):
:shipit: lgtm

dleard (approved):

PR #1703, opened 2021-05-26
Bugfixes: Prevent "Resume" button flash on Revise + show version number when viewing submitted

kriscooke:
- [GGIRCS-2325](https://youtrack.button.is/issue/GGIRCS-2325) Uses state to hold `newerDraftExists` variable when presenting the option to revise to prevent flashing "Resume latest draft" after clicking Revise
- [GGIRCS-2457](https://youtrack.button.is/issue/GGIRCS-2457)
  - Part A: Show the correct version number when viewing a submitted application
  - Part B: When viewing submitted applications that have subsequent drafts, the version number should be shown

### DIFF
--- a/app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx
+++ b/app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx
@@ -153,12 +153,15 @@ class ViewApplication extends Component<Props> {
       <div>
         Application ID: {application?.rowId}
         <br />
-        {thisVersion > 1 && (
-          <>
-            Version: {application?.latestDraftRevision.versionNumber}
-            <br />
-          </>
-        )}
+        {
+          // @ts-ignore
+          this.state.newerDraftExists && (
+            <>
+              Version: {thisVersion}
+              <br />
+            </>
+          )
+        }
         BC GHG ID: {application?.facilityByFacilityId.bcghgid}
       </div>
     );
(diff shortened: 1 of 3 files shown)


app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx
```diff
@@ -72,6 +72,19 @@ class ViewApplication extends Component<Props> {
...
+    super(props);
+    const latestSubmittedRevision =
+      props.query.application?.latestSubmittedRevision?.versionNumber;
+    const latestDraftRevision =
+      props.query.application?.latestDraftRevision?.versionNumber;
+    // Disabling in order to initialize state based on props:
+    // eslint-disable-next-line react/state-in-constructor
+    this.state = {
+      newerDraftExists: latestDraftRevision > latestSubmittedRevision
+    };
```
Comment (pbastia) [inline app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx:85]: we should be able to put the state initialization outside of the constructor (since the props are available before the constructor call from what I understand), this way we can remove the eslint warning

``` javascript
state = {
  newerDraftExists: latestDraftRevision > latestSubmittedRevision
}
constructor(props) { 
...
```

pbastia (dismissed):
A small nitpicky item that we can easily ignore, looks great!

app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx
Comment (kriscooke) [reply app/pages/reporter/application/[applicationId]/version/[versionNumber]/view.tsx:85]: > the props are available before the constructor call from what I understand

So you are right! Sometimes I get stuck in a function component mindset and forget what's available in class components as they seem so much more inflexible...

Fixing, since I'd rather not ts-ignore things.

pbastia (approved):
👌 :shipit:

PR #1706, opened 2021-05-26
fix: the display only filter doesn't render a search box

pbastia:
Fixes a bug where the `DisplayOnlyFilter` rendered an inactive search box on the products settings column

image


app/components/FilterableTable/Filters/TableFilter.tsx
```diff
@@ -56,6 +55,7 @@ export default abstract class TableFilter<T = string | number | boolean> {
           name={this.argName}
           value={(filterArgs[this.argName] ?? '') as string | number}
           aria-label={`Filter by ${this.title}`}
+          disabled={!this.isSearchEnabled}
```
pbastia, line 58: I guess this is a bit of a new feature but this should prevent confusion if the search is disabled by the dev

kriscooke (dismissed):
👌

PR #1709, opened 2021-05-26
chore: limit cronjob histories to 1 job

matthieu-foucault:


pbastia (approved):

matthieu-foucault:
This won't change anything, given that those cronjobs are created by Airflow, and the default successful history limit is 3 jobs (which obviously wasn't applied)

PR #1711, opened 2021-05-26
chore: add default approved org requests to dev data

dleard:
adds default approved org requests for test user & cypresstestreporter for all orgs (except SFO operator in case we want to look at the request process)


schema/data/dev/ciip_user_organisation.sql
```diff
@@ -13,22 +8,27 @@ begin;
...
   with rows as (
   insert into ggircs_portal.ciip_user_organisation(id, user_id, organisation_id, status)
   overriding system value
   values
-  (1, 3, 8, 'approved'),
-  (2, 3, 7, 'approved'),
-  (3, 6, 1, 'approved'),
-  (4, 6, 7, 'approved'),
-  (5, 6, 8, 'approved')
+  (1, 3, 2, 'approved'),
```
matthieu-foucault, line 16: Only `organisation_id` 1 is guaranteed to exist, as there is only one org when deploying the load testing data

matthieu-foucault: I'll update the load testing cron job so that it doesn't call this script

dleard: So no changes needed here? I was looking into conditionally adding values to the ciip_user_organisation table based on the count of organisations in the organisation table with a do block here.

matthieu-foucault: I was thinking about this too. Having a different file and calling the script with a different parameter would be cleaner than having a conditional depending on a previous script's behavior

matthieu-foucault: So no, no change needed here

matthieu-foucault (approved):
Looks good. Just needs a rebase

matthieu-foucault (approved):

PR #1717, opened 2021-05-27
fix: facilities applications are ordered by operator and facility name

matthieu-foucault:


schema/deploy/queries/facility_application_by_reporting_year.sql
```diff
@@ -25,8 +25,8 @@ create or replace function ggircs_portal.facility_application_by_reporting_year(
       on facility.id = application.facility_id
       and application.reporting_year = (select reporting_year from ggircs_portal.default_displayed_reporting_year())
     join ggircs_portal.organisation on facility.organisation_id = organisation.id
-    where _reporting_year is null
-    order by organisation.operator_name, facility.facility_name);
+    where _reporting_year is null)
+  ) sub order by operator_name, facility_name;
```
dleard, line 29: what does this 'sub' do?

dleard (dismissed):

kriscooke (commented):
Should we perhaps we should get some feedback on default ordering preference from our SME before deciding on operator and facility?

schema/deploy/queries/facility_application_by_reporting_year.sql
matthieu-foucault: it's just an alias for the subquery, which is required my postgres

dleard (approved):

PR #1725, opened 2021-06-02
Fix: imported swrs fuel fixes

dleard:
- null annual_fuel_amount values are coalesced to 0
- units are pulled from the ciip fuel table


pbastia (approved):
👌

pbastia (approved):

PR #1730, opened 2021-06-03
fix: mandatory products only populate on non-deleted associations

pbastia:
Fixes an issue where deleted product-naics-code associations still showed up if the product is mandatory


dleard (changes requested):
I'm still seeing the bug. I think the deletedAt fix also needs to be applied to the new ProductArrayField component that was added in Kristen's mandatory products PR a little while ago.

app/containers/Forms/ProductsArrayField.tsx
Kristen's PR: https://github.com/bcgov/cas-ciip-portal/pull/1679/files

dleard (approved):
:shipit:

PR #1731, opened 2021-06-04
Fix: emission form width

pbastia:
Fixes the issue where the emissions form (and any other form) width would shrink depending on its content.
Added a `<Row><Col md={12}>...` to enforce the content matching the form width.

This is most visible when the emissions form has no data reported and the "hide" button is clicked

image


matthieu-foucault (dismissed):
Just needs a rebase

matthieu-foucault (approved):

PR #1732, opened 2021-06-07
Fix: conditional benchmark settings display

dleard:
- hides benchmark option in admin product page if the product is not a ciip benchmarked product.
- updates the updateProduct mutation to return the shape of the fragment (fixes a require-refresh bug)
nobenchmark

### DIFF
--- a/app/containers/Products/ProductRowItemContainer.tsx
+++ b/app/containers/Products/ProductRowItemContainer.tsx
@@ -198,9 +198,11 @@ export const ProductRowItemComponent: React.FunctionComponent<Props> = ({
               <Dropdown.Item onClick={() => setProductModalShow(true)}>
                 Product details
               </Dropdown.Item>
-              <Dropdown.Item onClick={() => setBenchmarkModalShow(true)}>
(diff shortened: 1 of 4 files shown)


kriscooke (commented):
Looks good to me, but it made me notice another issue: 

If you edit the product details and save, it does not update the info for that product in the list until you refresh the page. So if you save a change to the details, then re-open the product details, the change is not there.

For the same reason, if I edit whether or not it's benchmarked, then the settings dropdown is not updated to hide/show the "Benchmarked" option.

dleard:
> Looks good to me, but it made me notice another issue:
> 
> If you edit the product details and save, it does not update the info for that product in the list until you refresh the page. So if you save a change to the details, then re-open the product details, the change is not there.
> 
> For the same reason, if I edit whether or not it's benchmarked, then the settings dropdown is not updated to hide/show the "Benchmarked" option.

Good eye. It's probably a quick fix in the update product mutation, I'll take a look.

kriscooke (approved):

PR #1734, opened 2021-06-07
fix: add missing product fields to mandatory update mutation

dleard:
The updateProduct mutation in the mandatory products component (ProductsArrayField) was missing the energy product boolean fields, which caused the validation to skip the energy product validations since they were always being evaluated as false.


pbastia (commented):
Looks good to me! 
We need to update those vulnerabilities first...

pbastia (approved):
:shipit:

PR #1735, opened 2021-06-07
Tech Debt: refactor network triggers (Reporter)

pbastia:
creating a higher order component to wrap mutation calls and allowing components to react to the state of that mutation (in flight or not).
This refactors:
- the ProductCreator form
- the buttons on the Reporter side, except the "Save & Continue" of the form

### DIFF
--- a/app/components/helpers/LoadingOnClickButton.tsx
+++ b/app/components/helpers/LoadingOnClickButton.tsx
@@ -12,7 +12,7 @@ interface Props extends ButtonProps {
   loadingText?: string;
 }
 
-const LoadinOnClickButton: React.FunctionComponent<Props> = ({
+const LoadingOnClickButton: React.FunctionComponent<Props> = ({
   loadingText,
   ...buttonProps
 }) => {
@@ -26,7 +26,7 @@ const LoadinOnClickButton: React.FunctionComponent<Props> = ({
 };
 
 export default withPromiseLoading(
-  LoadinOnClickButton,
+  LoadingOnClickButton,
   ASYNC_HANDLER,
   IN_FLIGHT_PROPERTY
 );
--- a/app/components/helpers/LoadingOnClickButton.tsx
+++ b/app/components/helpers/LoadingOnClickButton.tsx
@@ -1,0 +1,29 @@
+// Wraps a bootstrap Button with the `withPromiseLoading` HOC
+// This will make the button enter a disabled state until the
+// onClick handler completes
+import React from 'react';
+import withPromiseLoading from 'lib/withPromiseLoading';
+import {Button, ButtonProps} from 'react-bootstrap';
+
+const IN_FLIGHT_PROPERTY = 'disabled';
+const ASYNC_HANDLER = 'onClick';
+
+interface Props extends ButtonProps {
+  loadingText?: string;
+}
+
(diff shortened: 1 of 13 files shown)


app/components/helpers/LoadingOnClickButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+import {Button, ButtonProps} from 'react-bootstrap';
+
+const IN_FLIGHT_PROPERTY = 'disabled';
+const ASYNC_HANDLER = 'onClick';
+
+interface Props extends ButtonProps {
+  loadingText?: string;
+}
+
+const LoadinOnClickButton: React.FunctionComponent<Props> = ({
```
naomiaro, line 15: just minor, but `LoadinOnClickButton` -> `LoadingOnClickButton`

app/components/helpers/LoadingOnClickButton.tsx
```diff
@@ -0,0 +1,32 @@
+// Wraps a bootstrap Button with the `withPromiseLoading` HOC
+// This will make the button enter a disabled state until the
+// onClick handler completes
```
matthieu-foucault, line 3: Move this to a block comment above the `LoadingOnClickButton` declaration, that will make it more visible in the editor

app/lib/withPromiseLoading.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const handleAsyncEvent = async (...args) => {
+      if (!isSubscribed) return;
+      setIsLoading(true);
+
+      // We can only verify that the key exists on TProps,
+      // not the type of the property for that key
+      await (props[asyncEventHandler] as any)(...args);
+
+      if (!isSubscribed) return;
+      setIsLoading(false);
```
matthieu-foucault, line 34: We might want to wrap the `await` in a `try` block, and have the statements below in a `finally`

app/tests/unit/components/helpers/LoadingOnClickButton.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+describe('the loading on click button', () => {
+  it('renders its children in its normal state', () => {
+    const rendered = mount(
+      <LoadingOnClickButton loadingText="loading">
+        test text
+      </LoadingOnClickButton>
+    );
+
+    expect(rendered).toMatchSnapshot();
```
matthieu-foucault, line 13: can you add assertions in those tests that check the button's content, so that we don't rely on reviewers checking the snapshots

matthieu-foucault: ... and the `disabled` property too

matthieu-foucault (changes requested):

app/lib/withPromiseLoading.tsx
pbastia: Wouldn't this go against https://eslint.org/docs/rules/no-unsafe-finally ?

matthieu-foucault: interesting. Maybe just a try/catch then, and log the error in sentry in the catch block

matthieu-foucault (dismissed):

matthieu-foucault (approved):

PR #1741, opened 2021-06-11
2405

kriscooke:
Cleans up styles in DefaultLayout to get a handle on Bootstrap overrides vs. other global styles:
- Converts unique classes to ids to reflect specificity
- Removes defunct, unused styles or those that were never actually applied in practise
  * eg: `with-shadow`, `full-width`, among others
- Moves styles that were actually component-specific into that component
  * eg: `.footer` and mainly homepage-specific styles
- Moves the few global styles that were not Bootstrap overrides into the layout's local CSS, using the `:global()` selector
  * eg: `.blue`, `p`, and `.container.wide`
- Remaining global styles are all Bootstrap overrides for the purpose of contrast accessibility --> moved into a dedicated global style module, `GlobalBootstrapOverrideStyles`. All further global styles for this purpose should go here.

[GGIRCS-2405](https://youtrack.button.is/issue/GGIRCS-2405)


matthieu-foucault:
@pbastia do you want to rebase this, and I'll review it today

matthieu-foucault (approved):
The product

PR #1747, opened 2021-06-15
Fix: registration page validation

dleard:
validates the email and phone number fields on registration.


app/containers/User/UserForm.tsx
```diff
@@ -19,6 +20,21 @@ interface Props {
   onSubmit: () => any;
 }
 
+const customFormats = {
+  phoneNumber:
+    '^(\\+?\\d{1,2}[\\s,-]?)?\\(?\\d{3}\\)?[\\s.-]?\\d{3}[\\s.-]?\\d{4}$'
```
pbastia, line 25: we're assuming North-American phone numbers here, but I think that's fine :)

pbastia (dismissed):

pbastia (approved):

PR #1749, opened 2021-06-16
Fix/group perms

naomiaro:
https://youtrack.button.is/issue/GGIRCS-2481

- Fixes a bug with multi role group priority
- Writes tests for user group code

### DIFF
--- a/app/tests/unit/lib/user-groups.test.ts
+++ b/app/tests/unit/lib/user-groups.test.ts
@@ -6,7 +6,7 @@ import {
 } from 'lib/user-groups';
 
 describe('User Groups', () => {
-  it('should return all groups', async () => {
+  it('should return all groups', () => {
     const GROUPS = [
       'Guest',
       'User',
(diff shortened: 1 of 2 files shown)


app/tests/unit/lib/user-groups.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(groupPath).toBe('/');
+  });
+
+  it('should get the correct landing route', () => {
+    const userGroups = ['Incentive Administrator', 'Incentive Analyst'];
+    const groupPath = getUserGroupLandingRoute(userGroups);
+
+    expect(groupPath).toBe('/admin');
+  });
+});
```
pbastia, line 70: Nice! I would just use separate `describe(...)` for each of the functions we're testing, it'll make it easier to track where the failure is if we break something down the line

pbastia (changes requested):
looks good! just a minor change to how the tests are organised

pbastia (approved):
👌

PR #1790, opened 2021-06-22
BugFix: clicking on the icon sends the right status

pbastia:
fix for issue where clicking on the icon (not the button around) would trigger a `null` status for the application in the database

image

### DIFF
--- a/app/components/Admin/ApplicationReview/DecisionModal.tsx
+++ b/app/components/Admin/ApplicationReview/DecisionModal.tsx
@@ -23,10 +23,6 @@ export const DecisionModal: React.FunctionComponent<Props> = ({
     ? `Revert Decision`
     : 'Make a Decision or Request Changes';
 
-  const handleDecision = (e) => {
-    onDecision(e.target.value);
-  };
-
   const makeDecisionContent = (
     <>
       <Modal.Body>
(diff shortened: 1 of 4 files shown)


naomiaro (approved):

app/components/Admin/ApplicationReview/DecisionModal.tsx
```diff
@@ -40,7 +36,7 @@ export const DecisionModal: React.FunctionComponent<Props> = ({
             value="REQUESTED_CHANGES"
```
matthieu-foucault, line 36: Are the values needed?

app/components/Admin/ApplicationReview/DecisionModal.tsx
```diff
@@ -62,7 +58,7 @@ export const DecisionModal: React.FunctionComponent<Props> = ({
           value="REJECTED"
```
naomiaro, line 58: maybe don't need these values anymore

naomiaro (changes requested):
Didn't notice the buttons still had values until the end woops

naomiaro (dismissed):

matthieu-foucault (approved):

matthieu-foucault (approved):

PR #1935, opened 2021-07-22
Chore/dags live in own repo 1887

naomiaro:
Moves portal Airflow jobs to use the dag trigger job.

Consolidates the deploy_db dag depending on environment
Screenshot 2021-07-28 at 11 20 37 AM


naomiaro:
@matthieu-foucault I'm just going to read more about helm hooks now, but my current thinking for actually downloading these newly separated DAGs to Airflow before they're no triggered in this chart is to use the new chart dependency `cas-airflow-dag-trigger` two more times to call the dag `fetch_and_save_dag_from_github` with the info it needs at an appropriate time.

helm/cas-ciip-portal/Chart.yaml
```diff
@@ -12,3 +12,19 @@ dependencies:
...
+    repository: https://bcgov.github.io/cas-airflow
+    alias: download-airflow-ciip-cert-renewal
+  - name: cas-airflow-dag-trigger
+    version: 1.0.5
+    repository: https://bcgov.github.io/cas-airflow
+    alias: download-airflow-ciip-deploy-db
+  - name: cas-airflow-dag-trigger
+    version: 1.0.5
+    repository: https://bcgov.github.io/cas-airflow
+    alias: airflow-ciip-cert-renewal
```
matthieu-foucault, line 26: The certificate renewal is executed on a schedule (daily), so it does not need to be triggered when deploying, You can remove this dependency

helm/cas-ciip-portal/Chart.yaml
```diff
@@ -12,3 +12,19 @@ dependencies:
...
     repository: "https://codecentric.github.io/helm-charts/"
     condition: mailhog.enable
+  - name: cas-airflow-dag-trigger
+    version: 1.0.5
+    repository: https://bcgov.github.io/cas-airflow
+    alias: download-airflow-ciip-cert-renewal
+  - name: cas-airflow-dag-trigger
+    version: 1.0.5
+    repository: https://bcgov.github.io/cas-airflow
+    alias: download-airflow-ciip-deploy-db
```
matthieu-foucault, line 22: Ideally we'd have only one "download" DAG, either by updating it to support downloading a directory, or by having a single DAG file

Makefile
```diff
@@ -63,6 +65,10 @@ install:
 	--set ggircs.namespace=$(GGIRCS_PROJECT) \
 	--set ggircs.prefix=$(GGIRCS_NAMESPACE_PREFIX) \
 	--set ggircs.environment=$(ENVIRONMENT) \
+	--set download-airflow-ciip-cert-renewal.dagConfiguration="{\"org\": \"bcgov\", \"repo\": \"cas-ciip-portal\", \"ref\": \"$(GIT_SHA1)\", \"path\": \"dags/ggircs_cert_renewal.py\"}" \
+	--set download-airflow-ciip-deploy-db.dagConfiguration="{\"org\": \"bcgov\", \"repo\": \"cas-ciip-portal\", \"ref\": \"$(GIT_SHA1)\", \"path\": \"dags/ciip_deploy_db.py\"}" \
+	--set airflow-ciip-deploy-db.helm.hook="pre-upgrade" \
+	--set airflow-ciip-deploy-db.helm.hookWeight=0 \
```
matthieu-foucault, line 71: those could be set in the values file, as they are always the same value. The "download" dependency should have a `pre-upgrade,pre-install` hook, but the actual DAG trigger does not need to be a hook

helm/cas-ciip-portal/Chart.yaml
matthieu-foucault: We do need a manual trigger of the "issue" DAG, but only when the chart is being installed, so that should be a conditional dependency

helm/cas-ciip-portal/values.yaml
```diff
@@ -116,3 +116,27 @@ cas-postgres:
...
+  dagId: ciip_portal_acme_issue
+  helm:
+    hook: "post-install"
+
+# This job requires the cas-postgres dependency to be running,
+# and is required for the app deployment to become ready.
+# Therefore, it needs to be part of the initial install (instead of a pre/post install hook)
+# For upgrades, it needs to be a hook so that the delete policy kicks in
+# and the previous job template gets deleted (as editing job templates is not allowed)
+# It can be a hook during upgrades as the cas-postgres dependency will already be running
```
matthieu-foucault, line 137: this comment does not apply anymore

dags/cas_ciip_portal_dags.py
```diff
@@ -0,0 +1,110 @@
...
+ciip_deploy_db_args = {
+    **default_dag_args,
+    'start_date': YESTERDAY
+}
+
+SCHEDULE_INTERVAL = None
+
+default_dag = DAG('cas_ciip_portal_ciip_deploy_db', schedule_interval=SCHEDULE_INTERVAL,
+                  default_args=ciip_deploy_db_args)
+test_dag = DAG('cas_ciip_portal_ciip_deploy_db_test',
```
pbastia, line 29: I'm not sure if this DAG is still needed? If it is, it might be good to rename it something like `cas_ciip_portal_deploy_restore_from_prod` or something like that, so the user is not confused with what it does

dags/cas_ciip_portal_dags.py
```diff
@@ -0,0 +1,110 @@
...
+cron_acme_issue_task = PythonOperator(
+    python_callable=trigger_k8s_cronjob,
+    task_id='ciip_portal_acme_issue',
+    op_args=['cas-ciip-portal-acme-issue', namespace],
+    dag=acme_issue_dag)
+
+
+dag = DAG('cas_ciip_portal_acme_renewal', schedule_interval='0 8 * * *',
+          default_args=acme_renewal_args)
+
```
pbastia, line 105: since we're putting multiple DAGs in the same file, maybe we can add some comments to make finding them in the file easier

dags/cas_ciip_portal_dags.py
naomiaro: Yeah I'm also not sure if this dag is still used. @matthieu-foucault do you know if we use it?

pbastia: as @matthieu-foucault suggested we can have only one DAG that performs the "restore from prod" step if the `ENVIRONMENT` env variable is set to `test`

Makefile
```diff
@@ -47,6 +47,7 @@ install: OC_PROJECT=$(CIIP_NAMESPACE_PREFIX)-$(ENVIRONMENT)
 install: GGIRCS_PROJECT=$(GGIRCS_NAMESPACE_PREFIX)-$(ENVIRONMENT)
 install:
 	@set -euo pipefail; \
+	dagConfig=$$(echo '{"org": "bcgov", "repo": "cas-ciip-portal", "ref": "$(GIT_SHA1)", "path": "dags/cas_ciip_portal_dags.py"}' | base64 -w0); \
 	helm dep up ./helm/cas-ciip-portal; \
 	if ! helm status --namespace $(OC_PROJECT) cas-ciip-portal; then \
 		helm install --atomic --timeout 2400s --namespace $(OC_PROJECT) \
```
matthieu-foucault, line 53: this needs a `--wait-for-jobs` flag too

dags/cas_ciip_portal_dags.py
```diff
@@ -0,0 +1,110 @@
...
+        python_callable=trigger_k8s_cronjob,
+        task_id='cas-ciip-portal-prod-test-restore',
+        op_args=['cas-ciip-portal-prod-test-restore', namespace],
+        dag=dag)
+
+
+ciip_portal_init_db(default_dag) >> ciip_portal_swrs_import(default_dag) >> ciip_portal_deploy_data(
+    default_dag) >> ciip_portal_graphile_schema(default_dag) >> ciip_portal_app_user(default_dag)
+ciip_portal_init_db(test_dag) >> ciip_portal_prod_test_restore(test_dag) >> ciip_portal_deploy_data(
+    test_dag) >> ciip_portal_graphile_schema(test_dag) >> ciip_portal_app_user(test_dag)
```
matthieu-foucault, line 84: This is where you can have an conditional statement that would allow you to only have one `dag`, with different tasks depending on the environment.

matthieu-foucault (changes requested):

naomiaro:
@matthieu-foucault this runs well now

matthieu-foucault (approved):

PR #1951, opened 2021-08-05
fix: add probes port to container port

matthieu-foucault:


matthieu-foucault:
Nevermind, this is not needed. The deployment to dev just needed a bit more time to pass the liveness probes

PR #1961, opened 2021-08-17
feat: Emissions from EIOs unit change to GWh

dleard:


matthieu-foucault (approved):
Updating the product units in place here is ok because we only have one applicant who reported this product, and they reported in `GWh` (and left a comment). Otherwise we'd need to archive and recreate the product with a different ID

PR #1989, opened 2021-09-22
test: define style test schemas in makefile

dleard:
Define the schemas to test in the makefile with `--set` and use variables inside the style tests.

### DIFF
--- a/schema/Makefile
+++ b/schema/Makefile
@@ -71,7 +71,7 @@ deploy:
 .PHONY: prove_style
 prove_style:
 	# Run style-related test suite on all objects in db using pg_prove
-	@@$(PG_PROVE) --failures -d $(TEST_DB) test/style/*_test.sql --set schemas_to_test=ggircs_portal,ggircs_portal_private --set dummy_schema=test_fixture_schema --set audit_schema=audit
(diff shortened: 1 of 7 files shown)


schema/Makefile
```diff
@@ -71,7 +71,7 @@ deploy:
 .PHONY: prove_style
 prove_style:
 	# Run style-related test suite on all objects in db using pg_prove
-	@@$(PG_PROVE) --failures -d $(TEST_DB) test/style/*_test.sql
+	@@$(PG_PROVE) --failures -d $(TEST_DB) test/style/*_test.sql --set schemas_to_test=ggircs_portal,ggircs_portal_private --set dummy_schema=test_fixture_schema --set audit_schema=audit
```
matthieu-foucault, line 74: `dummy` is on the [list of terms to avoid](https://developers.google.com/style/word-list?hl=en), please find an alternative name (`placeholder_schema`,`temp_schema`,...). Should this one even be a variable?

matthieu-foucault (changes requested):

schema/Makefile
dleard: good point, dummy_schema and audit_schema do not need to be variables

matthieu-foucault (approved):

PR #1992, opened 2021-09-27
Feat: session extension modal

pbastia:
A modal dialog that pops up on any authenticated page when the session is about to expire

- [x] modal dialog
- [x] express endpoint
- [x] timer refresh logic


app/components/LogoutWarningModal.tsx
```diff
@@ -0,0 +1,53 @@
...
+      setRemainingSeconds(Math.floor((expiresOn - Date.now()) / 1000));
+    }, 1000);
+
+    return () => {
+      clearInterval(intervalId);
+    };
+  }, []);
+
+  return (
+    <Modal show size="lg" onHide={() => {}}>
```
matthieu-foucault, line 30:
```suggestion
    <Modal show size="lg">
```
Is the noop `onHide` needed?

app/components/SessionTimeoutHandler.tsx
```diff
@@ -0,0 +1,92 @@
...
+  modalDisplaySecondsBeforeLogout
+}) => {
+  const router = useRouter();
+
+  const [showModal, setShowModal] = useState(false);
+
+  // UNIX epoch (ms)
+  const [sessionExpiresOn, setSessionExpiresOn] = useState(Infinity);
+
+  const logoutOnSessionIdled = async () => {
```
matthieu-foucault, line 22: does this need to be async?

app/components/SessionTimeoutHandler.tsx
```diff
@@ -0,0 +1,92 @@
...
+        if (timeout > modalDisplaySecondsBeforeLogout) {
+          setShowModal(false);
+          timeoutId = setTimeout(() => {
+            checkSessionIdle();
+          }, (timeout - modalDisplaySecondsBeforeLogout) * 1000);
+        } else if (timeout > 0) {
+          setShowModal(true);
+          timeoutId = setTimeout(() => {
+            checkSessionIdle();
+          }, timeout * 1000);
```
matthieu-foucault, line 64: Maybe we can add the following comment to explain the different scenarios here, as it's not obvious:

```js
// We display the modal and set a timeout to check again when the session is due to expire.
// If the user has not extended their session by then we will redirect them (by invoking logoutOnSessionIdled() below)
// If they do extend their session, the `checkSessionIdle()` call will branch into the first condition above 
// and set another timeout to check the session idle when the modal is due to be displayed.
```

app/components/SessionTimeoutHandler.tsx
```diff
@@ -0,0 +1,92 @@
...
+  };
+
+  useEffect(() => {
+    let timeoutId;
+
+    const checkSessionIdle = async () => {
+      if (!pageComponent.isAccessProtected) return;
+
+      const response = await fetch('/session-idle-remaining-time');
+      if (response.ok) {
```
matthieu-foucault, line 50: should we log them out if the response is not ok?

app/components/SessionTimeoutHandler.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }, [router, pageComponent]);
+
+  return (
+    showModal && (
+      <LogoutWarningModal
+        inactivityDelaySeconds={modalDisplaySecondsBeforeLogout}
+        expiresOn={sessionExpiresOn}
+        onExtendSession={() => {
+          extendSession();
+        }}
```
matthieu-foucault, line 86:
```suggestion
        onExtendSession={extendSession}
```
unless there's a reason for the anonymous arrow function here?

app/pages/_app.tsx
```diff
@@ -64,6 +65,10 @@ export default class App extends NextApp<AppProps> {
 
     return (
       <Sentry.ErrorBoundary fallback={ErrorFallback}>
+        <SessionTimeoutHandler
+          pageComponent={Component}
+          modalDisplaySecondsBeforeLogout={60}
```
matthieu-foucault, line 70:
```suggestion
          modalDisplaySecondsBeforeLogout={120}
```

This should be 2min according to our acceptance criteria

app/pages/_app.tsx
```diff
@@ -64,6 +65,10 @@ export default class App extends NextApp<AppProps> {
 
     return (
       <Sentry.ErrorBoundary fallback={ErrorFallback}>
+        <SessionTimeoutHandler
+          pageComponent={Component}
```
matthieu-foucault, line 69: removing this prop and checking for `Component.isAccessProtected` here will make it easier to extract into its own package

matthieu-foucault (changes requested):

app/pages/_app.tsx
pbastia: oops! I had changed it for the demo and somehow committed it

app/components/LogoutWarningModal.tsx
pbastia: no!

app/components/SessionTimeoutHandler.tsx
pbastia: I think if we forcefully redirect them on a temporary either server roadbump or a short internet outage, they might lose data...the current implementation would cancel all timeouts though. Maybe we can just do a "check again in 5 minutes if there is connectivity again"?

app/components/SessionTimeoutHandler.tsx
pbastia: Just my react reflex to avoid the `bind()` call but it doesn't matter here :)

app/pages/_app.tsx
pbastia: we're relying on the component changing for the refresh condition on the `useEffect` call, we'd need a roundabout way of reloading this component if we were changing this
What I can suggest is to create an interface
```javascript
interface ISupportAccessProtection{
  isAccessProtected: boolean;
}
```
And we'd use that for the type of the component, instead of depending on the `CiipPageComponent` type.

app/components/SessionTimeoutHandler.tsx
naomiaro: If there's connectivity issues the promise is going to throw an exception here.

`and it will only reject on network failure or if anything prevented the request from completing.`
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

Then it looks like it would perhaps reach the sentry boundary?

pbastia: consensus reached: we will redirect the user to the logout page, since at that point things aren't functional

matthieu-foucault (approved):

PR #2002, opened 2021-10-05
Feat: ignore non carbon taxed emissions

dleard:
The emission_category_missing_fuel validation function should not throw for non carbon-taxed emission categories


mocks_schema/sqitch.plan
```diff
@@ -1,5 +1,6 @@
 %syntax-version=1.0.0
 %project=mocks
+%uri=mocks
```
matthieu-foucault, line 3: is sqitch complaining about having multiple projects without a URI? If so this one should probably be `https://github.com/bcgov/cas-ciip-portal/tree/develop/mocks_schema`

matthieu-foucault (changes requested):
This is only ignoring emissions from non-taxed categories, we also need to ignore fuels with non-taxed emission categories

PR #2044, opened 2021-12-08
chore: update nextjs to v11 and remove obsolete dependencies

matthieu-foucault:


dleard (approved):

app/tests/json-schema-utils.ts
```diff
@@ -1,5 +1,5 @@
 import jsf from "json-schema-faker";
-import { FormJson } from "next-env";
+import { FormJson } from "types";
```
pbastia, line 2: looks like this can also be an `import type` ?
```suggestion
import type { FormJson } from "types";
```

matthieu-foucault: I'm actually just seeing that typescript will automatically detect and erase those imports that are not used at runtime, so this may not have any effect: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

PR #2062, opened 2022-02-11
fix: get_swrs_facility_data() does not rely on deprecated view

dleard:
swrs.facility_details has been deprecated.
This PR removes the dependency from ggircs_portal.get_swrs_facility_details()


pbastia (approved):
awesome. more simplifications!

Issue #2074, opened 2022-03-21
As a CIIP analyst/admin, I want the units for Sold Electricity in Gwh instead of Mwh, so that the units are accurate


dleard:
#### Description:

Sold Electricity has the wrong units attached to the product. They are currently in Mwh & should be in Gwh. Changing the units for the existing product requires making some adjustments to any applications already submitted that have reported this product. This includes:
Total Records: **136**
Final Submitted Records: **67**
Total Records where amount > 0: **49**
Final submissions where amount > 0: **23**

These legacy records will need to have the reported amount for Sold Electricity adjusted by a factor of 1000 to be accurate after the unit change. We will need robust tests to ensure this is done accurately and no other records are affected.

#### Development Checklist:
* [x] The units for Sold Electricity in schema/data/prod/energy_product.sql are in Gwh
* [x] A sqitch change updates the json form_result column in the form_result table for all production form_results that reported this product (divide the amount by 1000)
* Meets the DOD

dleard:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-ciip-portal/2074?planning-poker) with ZenHub @BCerki @gurjmatharu @matthieu-foucault @pbastia

dleard:
I was thinking about a testing strategy for this. I think it will be a combination automated testing in the code and manual testing in the test environment.

A pgTap test can create a function that does the same thing as the migration. We can create a set of dummy data with form_results that contain sold electricity and some that don't, then run the function to assert that only the proper records are changed & the changes made were done accurately.

I would also like to retrieve all the application/form_result IDs from metabase that contain this product in the TEST environment & run the function there, with a quick edit to the function to include a check that it only changed data for those records. I'd also like to download the carbon tax / incentive data for these applications before running the function and comparing the results after running the function to see if anything odd happens to the calculations after the change.

matthieu-foucault:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-ciip-portal/2074?planning-poker) with ZenHub @mikevespi

dleard:
[...]

PR #2076, opened 2022-03-28
Feat: sold electricity units to GWh

dleard:
issue #2074


schema/deploy/database_functions/change_sold_electricity_units_to_GWh.sql
```diff
@@ -0,0 +1,57 @@
...
+            if (element::jsonb->>'productRowId')::int = 1 then
+              -- Edit the JSON data to change the productUnits to GWh and scale the productAmount by a factor of 1000.
+              new_data := '{"productUnits": "GWh", "productAmount": ' || ((element::jsonb->>'productAmount')::numeric / 1000)::real ||' }';
+              json_data := (select jsonb (element) - 'productAmount' - 'productUnits' || new_data::jsonb);
+            end if;
+            -- Append the JSON data to the product_array to recreate the array of products.
+            product_array = array_append(product_array, json_data);
+          end loop;
+          -- Update the form_result record with the edited JSON data.
+          update ggircs_portal.form_result set form_result = ('[' || array_to_string(product_array, ',') || ']')::jsonb where id = temp_row.form_result_id;
```
pbastia, line 48: I *think* this should work to avoid more string concatenation:
```suggestion
          update ggircs_portal.form_result set form_result = array_to_json(product_array)::jsonb where id = temp_row.form_result_id;
```

schema/revert/database_functions/change_sold_electricity_units_to_GWh.sql
```diff
@@ -0,0 +1,39 @@
...
+    for temp_row in
+      select application_id, id as form_result_id, form_result
+      from ggircs_portal.form_result
+      where form_id=(select id from ggircs_portal.form_json where slug='production')
+      and form_result::jsonb@>'[{"productRowId":1}]'::jsonb = true
+      loop
+        for element in select jsonb_array_elements(form_result) from ggircs_portal.form_result where id=temp_row.form_result_id
+          loop
+            if (element::jsonb->>'productRowId')::int = 1 then
+              new_data := '{"productUnits": "MWh", "productAmount": ' || ((element::jsonb->>'productAmount')::numeric * 1000)::real ||' }';
```
pbastia, line 26: same as before with `json_build_object`

schema/deploy/database_functions/change_sold_electricity_units_to_GWh.sql
```diff
@@ -0,0 +1,57 @@
...
+      from ggircs_portal.form_result
+      where form_id=(select id from ggircs_portal.form_json where slug='production')
+      and form_result::jsonb@>'[{"productRowId":1}]'::jsonb = true
+      -- Loop over each element of the form_result JSON array and edit the productUnits & productAmount if the element is the Sold Electricity product.
+      loop
+        for element in select jsonb_array_elements(form_result) from ggircs_portal.form_result where id=temp_row.form_result_id
+          loop
+            -- Grab the JSON data for the current element.
+            json_data := element;
+            if (element::jsonb->>'productRowId')::int = 1 then
```
pbastia, line 39: Should we check that the unit is MWh? Just in case some things happen out of order and we end up by accident with both MWh and GWh in the system, this would allow the migration to heal the dataset
```suggestion
            if (element::jsonb->>'productRowId')::int = 1 and (element::jsonb->>'productUnits') = 'MWh' then
```

schema/revert/database_functions/change_sold_electricity_units_to_GWh.sql
```diff
@@ -0,0 +1,39 @@
...
+    product_index := 0;
+    for temp_row in
+      select application_id, id as form_result_id, form_result
+      from ggircs_portal.form_result
+      where form_id=(select id from ggircs_portal.form_json where slug='production')
+      and form_result::jsonb@>'[{"productRowId":1}]'::jsonb = true
+      loop
+        for element in select jsonb_array_elements(form_result) from ggircs_portal.form_result where id=temp_row.form_result_id
+          loop
+            if (element::jsonb->>'productRowId')::int = 1 then
```
pbastia, line 25: same as before, only roll back the items where the unit is GWh

schema/test/unit/database_functions/change_sold_electricity_units_to_GWh_test.sql
```diff
@@ -0,0 +1,416 @@
...
+            -- Append the JSON data to the product_array to recreate the array of products.
+            product_array = array_append(product_array, json_data);
+          end loop;
+          -- Update the form_result record with the edited JSON data.
+          update ggircs_portal.form_result set form_result = ('[' || array_to_string(product_array, ',') || ']')::jsonb where id = temp_row.form_result_id;
+          -- Reset the product_array for the next iteration.
+          product_array := null;
+      end loop;
+    end
+  $body$;
```
pbastia, line 349: I wonder if we could put that in a function instead? Otherwise we're not testing the same code that is running, and changes to the original code won't be tested anymore unless this test is updated too. This breaks the regression testing aspect

pbastia (changes requested):
Good stuff, and nice to make it reversible with sqitch!

schema/test/unit/database_functions/change_sold_electricity_units_to_GWh_test.sql
dleard: It could definitely be a function, I actually have a function version of this in my .scratch folder. I'd left it as a do block since it was only ever supposed to run once, but it makes sense to be a function after the suggestion above of adding in the check for `MWh` to allow healing the dataset

pbastia (approved):
awesome

Issue #2042, opened 2021-11-29
Reporting Period Bar - applicants see above the table of facilities and how it interacts/differs from the column “Reporting Period of last SWRS report”


Sriram-pc:
#### Description:
Brief description of the work involved, including any user research context or key decisions.

#### Acceptance Criteria:
* Option - use Ghirken:
Given
When
Then
* [Other AC]
* [Other AC]
* [Other AC]
* Meets the DOD

dleard:
@LindsayMacfarlane 
To test this:
Log in to CIIP as a reporter and go to the list of facilities. There should now be a 'SWRS Report Received' column in that table where the 'Last Reporting Year of SWRS Report' column was.

LindsayMacfarlane:
Awesome! Thank you @dleard. I was able to test in dev by following your instructions above and can confirm that there is now a "SWRS report received" column that is filterable based on yes/no.

PR #2079, opened 2022-04-20
Feat: last swrs report column to boolean

dleard:
issue #2042


schema/deploy/queries/facility_application_by_reporting_year.sql
```diff
@@ -9,7 +9,7 @@ create or replace function ggircs_portal.facility_application_by_reporting_year(
   select * from (
     (select facility.id, organisation.id, application.id,
       organisation.operator_name, facility.facility_name, facility.facility_type, facility.bcghgid,
-      facility.swrs_facility_id, application.reporting_year
+      facility.swrs_facility_id, coalesce(application.reporting_year, _reporting_year) as reporting_year
```
matthieu-foucault, line 12: Do you need a `coalesce` here, or just `_reporting_year`? If `application.reporting_year` is not null, it should be equal to `_reporting_year` because of the join condition below, right? 

Is this also needed in the second half of the union below (use `(select reporting_year from ggircs_portal.default_displayed_reporting_year())` instead of `application.reporting_year`)?

matthieu-foucault (changes requested):
one small change. Looks good!

matthieu-foucault (approved):

PR #2085, opened 2022-05-02
chore: updating the wording for 2022

pbastia:


app/containers/RegistrationLoginButtons.tsx
```diff
@@ -29,10 +29,11 @@ export const RegistrationLoginButtonsComponent: React.FunctionComponent<Props> =
     </>
   ) : (
     <>
-      <p>The due date for CIIP application forms has passed.</p>
+      <p>In 2022, CIIP applications will be accepted from May 24 to June 30.</p>
```
matthieu-foucault, line 32: Do we want this to be displayed based on the configured reporting periods?

pbastia: That's a great idea, we can suggest that as a feature for the business and implement it in a separate PR to make sure we get this shipped this week

matthieu-foucault (approved):

PR #2140, opened 2023-01-17
Chore: security fixes

pbastia:
And cleaning up old resolution that are not unnecessary


Sepehr-Sobhani (approved):

dleard (dismissed):
Nice! Awesome that so many of those forced resolutions got removed.

dleard (approved):

PR #2147, opened 2023-01-20
feat: enabling bceid login

pbastia:


app/components/Layout/Header.tsx
```diff
@@ -134,12 +134,19 @@ const HeaderLayout: React.FunctionComponent<Props> = ({
...
+                  <LoginButton idpHint="idir">
                     <button className="nav-button" type="submit">
                       Login (IDIR)
                     </button>
                   </LoginButton>
                 </li>
+                <li>
+                  <LoginButton idpHint="bceidboth">
+                    <button className="nav-button" type="submit">
+                      Login (BCeID)
```
pbastia, line 146: We might want to rename this to something like "Industrial Operator Login (BCeID)" and "Program Administrator Login (IDIR)" to avoid confusion

dleard (approved):

PR #2162, opened 2023-02-15
chore: update ciip_user rls policy

dleard:


schema/deploy/policies/ciip_user_policies_003_restrict_permissions.sql
```diff
@@ -0,0 +1,45 @@
+-- Deploy ggircs-portal:policies/ciip_user_policies_003_restrict_permissions to pg
+-- requires: policies/ciip_user_policies_002_recreate_after_type_change
```
BCerki, line 2: Should I always be requiring the previous version of the function in the new one?

schema/deploy/policies/ciip_user_policies_003_restrict_permissions.sql
```diff
@@ -0,0 +1,45 @@
...
+  select id from ggircs_portal.ciip_user where uuid ilike '%@idir';
+$fn$ language sql strict stable security definer;
+
+grant execute on function ggircs_portal_private.get_valid_users to ciip_administrator, ciip_analyst, ciip_industry_user;
+
+do
+$policy$
+declare industry_user_statement text;
+begin
+industry_user_statement := 'id in (select ggircs_portal_private.get_valid_users()) or uuid=(select sub from ggircs_portal.session())';
```
BCerki, line 37: Can/should the check for `uuid=(select sub from ggircs_portal.session())` be part of the `get_valid_users` function?  I'd expect a function called valid users to give me all the valid users including whoever's logged in

schema/test/unit/tables/ciip_user_test.sql
```diff
@@ -69,16 +79,42 @@ select throws_like(
     'Administrator cannot delete rows from table_ciip_user'
 );
 
+select user_id from ggircs_portal.ciip_user_organisation where organisation_id=7;
+select * from ggircs_portal.ciip_user_organisation;
```
BCerki, line 83: Leftover?

schema/test/unit/tables/ciip_user_test.sql
```diff
@@ -69,16 +79,42 @@ select throws_like(
...
+
 -- CIIP_INDUSTRY_USER
+set jwt.claims.sub to '00000000-0000-0000-0000-000000000000';
 set role ciip_industry_user;
 select concat('current user is: ', (select current_user));
 
-select results_eq(
+select id from ggircs_portal.ciip_user;
```
BCerki, line 90: leftover?

schema/test/unit/tables/ciip_user_test.sql
```diff
@@ -69,16 +79,42 @@ select throws_like(
...
-    'Industry user can view all data from ciip_user'
+    'Industry user can view their own data in the ciip_user table'
+);
+
+select isnt_empty(
+  $$
+    select * from ggircs_portal.ciip_user where id = 6;
+  $$,
+    'Industry user can view data from ciip_user table for users within the same approved organisation'
+);
```
BCerki, line 104: Not sure I'm totally following these `select *` tests, but they seem to check that users are able to see the things they should. I don't see they check that they can't see things they shouldn't (except for the `is_empty` one below).

BCerki (commented):
Thanks for the walkthrough on this yesterday!

schema/deploy/policies/ciip_user_policies_003_restrict_permissions.sql
dleard: very good point. I'll add it another union to the function to do this instead of slipping it into the using/with check statment

schema/deploy/policies/ciip_user_policies_003_restrict_permissions.sql
dleard: For these numbered migrations I always require the previous one, it may not _actually_ depend on the previous change, but it does have a dependency somewhere along the chain and if we make all subsequent numbered migrations depend on the previous numbered migration then we will always have a transitive dependency to wherever it does have a dependency in the chain.

schema/test/unit/tables/ciip_user_test.sql
dleard: I've set up users 2 & 6 in a way that the current user (user 3) should be able to see that data, so isnt_empty asserts that something is returned when querying those user ids. I've set up user 1 in a way that user 3 should not be able to read that data, so when querying the table for user 1's data, it comes back empty (user 3 cannot see it).

dleard: policies differ from grants where if a user queries for data that they're not allowed to see, no exception is thrown. The data just silently is not returned from the query.

dleard:
closing in favour of #2166

PR #2167, opened 2023-02-21
fix: removing notify admin on new submissions

gurjmatharu:


app/cypress/integration/email.spec.js
```diff
@@ -133,44 +133,6 @@ describe("Confirmation emails", () => {
   beforeEach(() => {
     cy.mockLogin("reporter");
   });
-
```
BCerki, line 136: This is the only test in this describe block so I think we can drop the whole thing

BCerki (commented):
Looking good!

BCerki (approved):

BCerki (approved):

Issue #2145, opened 2023-01-19
As a CIIP Admin, I want fuels to be associated to emission categories, so that there is less margin for error when reporters are reporting fuels


dleard:
#### Description:
Applicants should report flaring and venting using the Natural Gas fuel. The guidance states this, but it was ignored by many and we received many applications with various other fuels reporting under flaring and venting.

This is a replacement card for #1985. The previous card outlined the creation of a UI for CIIP admins to be able to associate fuels to emission categories themselves, but given timelines we have instead decided to do this manually as a dev team.

We should get the list of fuels and their associated emission categories from CAS, then create and populate the necessary tables ourselves.

#### Acceptance Criteria:
- A fuel_emission_category table exists. This table has a the data structure {id, fuel_id, emission_category_id} plus the timestamp columns generated by the upsert_timestamp_columns function.
- The fuel_emission_category table has select, insert and update permissions for admins, and select for reporters and analysts.
- The fuel_emission_category table is populated with the fuel-emission category relationship data as provided by CAS
- Tests

LindsayMacfarlane:
@dleard - can this ticket be estimated prior to us getting the list?

dleard:
yes, similar to the CTA incremental tax work this should be a minimal amount of data entry work

dleard:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-ciip-portal/2145?planning-poker) with Zenhub @BCerki @gurjmatharu @mikevespi @pbastia @sam-warren @Sepehr-Sobhani @tmastrom

LindsayMacfarlane:
Program area has provided fuels_emissions category. File can be found in Scrum Team>Files>CIIP. Direct link: https://bcgov.sharepoint.com/:x:/t/00608-ScrumTeam/EfgHZU1baH5GriNNwlZl64cBUs02NOUn_FWS0Fx69zX_ww?e=zoNYB7

Dianadec:
@gurjmatharu @tmastrom please add your estimates if you would like to.

Dianadec:
@LindsayMacfarlane can the business area tag be removed now that we have that file?

LindsayMacfarlane:
Hi @tmastrom - please add your estimate :)

Dianadec:
Opportunity for new CIIP people to try this card.

BCerki:
[...]

PR #2169, opened 2023-02-22
Create fuel_emission_category table

BCerki:
Card: https://github.com/bcgov/cas-ciip-portal/issues/2145

To check I'd correctly entered the data:
- checked that the number of inserted records (216) matched the sum of records in the spreadsheet (added up all the 1s)
- made sure that the breakdown of each category (eg 8 flaring) matched the amount of times find/replace found that category

### DIFF
--- a/schema/deploy/tables/fuel_emission_category.sql
+++ b/schema/deploy/tables/fuel_emission_category.sql
@@ -1,0 +1,55 @@
+-- Deploy ggircs-portal:tables/fuel_emission_category to pg
+
+begin;
+create table ggircs_portal.fuel_emission_category (
+  id integer primary key generated always as identity,
+  fuel_id integer not null references ggircs_portal.fuel(id), 
+  emission_category_id integer not null references ggircs_portal.emission_category(id)
+);
+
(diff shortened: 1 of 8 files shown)


schema/deploy/tables/fuel_emission_category.sql
```diff
@@ -0,0 +1,54 @@
...
+-- Grant ciip_industry_user permissions
+perform ggircs_portal_private.grant_permissions('select', 'fuel_emission_category', 'ciip_industry_user');
+
+
+
+end
+$grant$;
+
+-- Enable row-level security
+-- alter table ggircs_portal.incremental_fuel_charge_baseline enable row level security;
```
BCerki, line 41: The card doesn't ask for this but it was on the other tables so we might need it here too? (but with the correct table name)

dleard: For this table we don't need any granular RLS policies. The grants you've defined here cover it

dleard (approved):
This looks great. Awesome work :100: 
I manually checked the values in the table against the data we received from CAS & all values are correct.
This branch just needs a rebase

dleard (approved):
This looks great. Awesome work. 
I manually checked the values in the table against the data we received from CAS & all values are correct.

PR #2170, opened 2023-02-23
fix: parse comments from json array

dleard:
The comments were incorrectly being pulled from the root of the json object. They live inside each fuel object in the array, so they needed to be parsed from those individual array objects rather than the root.


mikevespi (approved):
👌 Looks good to me!

PR #2178, opened 2023-03-10
fix: move logout button outside profile dropdown

Sepehr-Sobhani:
[ZH CARD 2149](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-ciip-portal/2149)


pbastia (changes requested):
Amazing! So much code deleted :)
We might need just a couple more test on `Header.tsx` with `isRegistered` and `isLoggedIn` set to make sure the right things are rendered

pbastia (approved):
amazing

PR #2180, opened 2023-03-15
Feat: display bceid business name

dleard:
Screenshot from 2023-03-15 15-37-30
show bceid_business_name on request for access page if user logged in with business bceid


app/cypress/integration/search-tables.spec.js
```diff
@@ -113,12 +113,12 @@ describe("When filtering organisation requests", () => {
     cy.useMockedTime(new Date("June 10, 2020 09:00:00"));
   });
 
-  it("The application can be filtered by multiple search fields", () => {
+  it.only("The requests can be filtered by multiple search fields", () => {
```
pbastia, line 116: `only` left behind

schema/deploy/tables/ciip_user_organisation_001_bceid.sql
```diff
@@ -0,0 +1,9 @@
+-- Deploy ggircs-portal:tables/ciip_user_organisation_001_bceid to pg
+-- requires: tables/ciip_user_organisation
+
+begin;
+
+alter table ggircs_portal.ciip_user_organisation add column bceid_business_name varchar(1000);
```
pbastia, line 6: There are scenarios (company merge, business bceid org reset) where an existing BCeID can get a new business name. We'd probably need to invalidate something if that was the case? Or figure how to handle that case with the business area.

pbastia (changes requested):
amazing. just an `only` left behind

pbastia (approved):

Issue #2181, opened 2023-03-28
As an Analyst, I want to download attachments to an application, so that I can access verification statements


pbastia:
#### Description:

Analysts need to download the verification statement(s) from an application version, when it has been attached

#### Acceptance Criteria:

Given I am a CIIP analyst
When I view an application from a reporter
And a verification statement was uploaded
Then I can download the verification statement

Given I am a CIIP analyst
When I view an application from a reporter
And no verification statement was uploaded
Then I can see a message "no verification statement was uploaded by the applicant"

Given I am a CIIP Industrial reporter
When I have uploaded a document
Then I can click on the title to download it

#### Dev checklist
- [ ] see [cif download middleware](https://github.com/bcgov/cas-cif/blob/develop/app/server/middleware/attachmentDownloadRouter.ts)

#### Wireframes
![image.png](https://images.zenhubusercontent.com/5faafee70e671ac198e88bc7/4bb59430-ebce-4a20-86f8-7bd817d594d1)

pbastia:
Fantastic work @tmastrom ! I can see the verification statements to download both on the application summary for the external applicant, and for the analyst.

PR #2185, opened 2023-04-03
Feat: 2126 Upload verification files

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-ciip-portal/2126


app/server/postgraphile/uploadFieldPlugin.js
```diff
@@ -0,0 +1,147 @@
+/* eslint-disable */
```
BCerki, line 1: This is copy-pasted from CIF with no linting errors, so I'm assuming it's fine to disable here

helm/cas-ciip-portal/templates/bucket-provisioning-job.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.db.restore.enabled }}
```
BCerki, line 1: Note to self: delete this whole file, already handled in another PR

app/containers/Forms/VerificationStatement.tsx
```diff
@@ -0,0 +1,178 @@
...
+            return (
+              <>
+                <a href="#" className="attachment-link" key={node.id}>
+                  {node.fileName}
+                </a>{" "}
+                <FontAwesomeIcon
+                  icon={faTrash}
+                  onClick={() => {
+                    const { environment } = relay;
+                    return deleteAttachmentMutation(environment, {
```
dleard, line 102: **To be done in a separate PR:** 
This deletes the attachment from the database, but not the GCS bucket. We will need to use https://www.npmjs.com/package/@google-cloud/storage to delete the file from GCS & then remove it from the database if the GCS delete was successful.

app/containers/Forms/VerificationStatement.tsx
```diff
@@ -0,0 +1,178 @@
...
+        },
+      },
+    };
+
+    await createAttachmentMutation(environment, variables)
+      .catch(() => {
+        setIsUploading(false);
+      })
+      .then(() => setIsUploading(false));
+  };
```
dleard, line 59: **To be done in a separate PR:**
We should put in some protections on file size / type. Let's reach out to CAS & determine if they know what file types they expect these verification statements to be formatted as.

app/mutations/application/deleteAttachment.ts
```diff
@@ -0,0 +1,33 @@
+import { graphql } from "react-relay";
+import RelayModernEnvironment from "relay-runtime/lib/store/RelayModernEnvironment";
+import {
+  deleteAttachmentMutationVariables,
+  deleteAttachmentMutation as deleteAttachmentMutationType,
+} from "deleteAttachmentMutation.graphql";
+import BaseMutation from "mutations/BaseMutation";
+
+const mutation = graphql`
+  mutation deleteAttachmentMutation(
```
dleard, line 10: **To be done in a separate PR:**
We will want to handle delete differently for previous versions (hard delete if attachment is from current version, soft delete it attachment was added in a previous version).
We can either define two mutations here and conditionally pass them to the performMutation() call below, or we can override the behaviour with a custom SQL delete mutation.

schema/deploy/tables/attachment.sql
```diff
@@ -0,0 +1,69 @@
...
+create index attachment_file on ggircs_portal.attachment(file);
+create index application_id on ggircs_portal.attachment(application_id);
+create index ggircs_portal_attachment_foreign_key on ggircs_portal.attachment(application_id, version_number);
+
+select ggircs_portal_private.upsert_timestamp_columns(
+  table_schema_name := 'ggircs_portal',
+  table_name := 'attachment',
+  add_create := true,
+  add_update := true,
+  add_delete := false);
```
dleard, line 28: **To be done in a separate PR:**
We will need to make `add_delete` true here to enable soft delete

dleard: (And add comments on the new `delete` columns)

dleard (approved):
Great work on this! 
There are a few things that we need to address around upload/delete but I am approving this PR as-is since it unblocks the download card #2181 & allows us to do more work in parallel. 
I'll write up a new ticket in zenhub to capture the changes I mentioned in the comments on this PR.

PR #2191, opened 2023-04-11
Feat/fuels by emission category

mikevespi:
Filter fuels by emissions category to prevent erroneous reporting.

### DIFF
--- a/app/containers/Forms/FuelField.tsx
+++ b/app/containers/Forms/FuelField.tsx
@@ -15,11 +15,14 @@ export const FuelFieldComponent: React.FunctionComponent<Props> = (props) => {
     [formData]
   );
 
-  const emissionCategoryFuelIds = useMemo(() => {
+  const getFuelIdsByEmissionCategoryRowId = (EmissionCategoryRowId) => {
     const categoryObject = query.fuelIdsByEmissionCategoryId.edges.find(
-      (edge) => edge.node.emissionCategoryId === selectedEmissionCategoryRowId
+      (edge) => edge.node.emissionCategoryId === EmissionCategoryRowId
     );
     return categoryObject?.node?.fuelIds;
+  };
+  const emissionCategoryFuelIds = useMemo(() => {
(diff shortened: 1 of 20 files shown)


app/containers/Forms/FuelField.tsx
```diff
@@ -9,7 +9,66 @@ interface Props extends FieldProps {
...
+    () => ({
+      activeRowIds: props.query.activeFuels.edges
+        .filter(({ node }) => emissionCategoryFuelIds?.includes(node.rowId))
+        .map(({ node }) => node.rowId),
+      activeNames: props.query.activeFuels?.edges
+        ?.filter(({ node }) => emissionCategoryFuelIds?.includes(node.rowId))
+        .map(({ node }) => node.name),
+    }),
+    [props.query, emissionCategoryFuelIds]
+  );
```
dleard, line 38: This whole area is a little busy & inter-related. Could probably use some comments describing what's happening in each function.

app/containers/Forms/FuelField.tsx
```diff
@@ -29,13 +88,37 @@ export const FuelFieldComponent: React.FunctionComponent<Props> = (props) => {
...
+    if (newFuelIds && newFuelIds.includes(fuelRowId)) onChange(fuel);
+    else
+      onChange({
+        ...fuel,
+        fuelRowId: undefined,
+        quantity: "",
+        fuelUnits: undefined,
+      });
+  };
+
```
dleard, line 106: Same thing here, Just a comment describing the flow here where this function clears fuel data if the emission category is changed to one that does not include that fuel

dleard (changes requested):
This is awesome! I tested it manually and everything works great. 
I think a few comments describing what's happening in some of the more complicated function interactions would be helpful, but otherwise this looks good to go.

dleard (dismissed):
:shipit:

dleard (approved):
:100:

PR #2193, opened 2023-04-11
Fix: 2192 ciip production product name

dleard:
The CIIP Production view was returning empty values for Product Name in many cases.
This PR will fix the view so that it always returns a value


mikevespi (approved):
:shipit: nice!

PR #2205, opened 2023-04-24
2196 Attachments part 3

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-ciip-portal/2196

Note:
- `graphqluploadexpress` has a `maxFileSize` setting but there's not much info about it and people are having trouble with it, so I set the errors manually. Details on graphql upload:
-- docs: https://yarnpkg.com/package/graphql-upload#:~:text=Added%20new%20maxFieldSize%2C%20maxFileSize%20and%20maxFiles%20middleware%20options.


schema/deploy/tables/attachment.sql
```diff
@@ -12,8 +12,7 @@ create table ggircs_portal.attachment
   file_size varchar(100),
   application_id integer not null references ggircs_portal.application(id),
   version_number int,
-  foreign key (application_id, version_number) references ggircs_portal.application_revision(application_id, version_number),
-  unique (application_id, version_number)
+  foreign key (application_id, version_number) references ggircs_portal.application_revision(application_id, version_number)
```
BCerki, line 15: This prevented people from uploading more than one attachment, ok to drop? It comes with the side effect that people can upload attachments with the same filename

app/containers/Forms/VerificationStatement.tsx
```diff
@@ -87,38 +121,63 @@ export const VerificationStatementComponent: React.FunctionComponent<Props> = ({
...
+            justifyContent: "space-around",
+            alignItems: "center",
+          }}
+        >
+          {application.attachmentsByApplicationId.edges.length > 0 && (
+            <table className="bc-table">
+              <tr>
+                <th>File</th>
+                <th>Created Date</th>
+                <th>Version Number</th>
```
dleard, line 145: I wonder if "Version Number" could be confusing here. It could be perceived as the file's version number, but what we're showing here is the Application's version number. Should we be explicit here and make the header "Application Version Number"?

schema/deploy/tables/attachment.sql
dleard: yes, this should be fine

dleard (changes requested):
Great Work!
I do think we need to see all the attachments when viewing a submitted application as either a reporter or analyst/admin, otherwise we're going to lose the ability to see the files.
Here's the flow I followed:
I uploaded 2 files & submitted the application.
I looked at the application as a reporter and saw my 2 files
I looked at the application as an admin and saw my 2 files
I created a new revision of the application
I uploaded 1 new file
I looked at the application as a reporter and only saw my 1 new file
I looked at the application as an admin and only saw my 1 new file
I created a 3rd revision of the application
I uploaded no files
I looked at the application as a reporter and it says no verification file uploaded
I looked at the application as an admin and it says no verification file uploaded

This means if the reporter needs to make a change other than re-uploading a verification statement (ie: fix a fuel/emission value) and they don't re-upload all files. it will show the analyst/admin that no verification files were uploaded.

The issue is in `ApplicationDetailsContainer`, where the attachments are being retrieved with the `attachmentsByApplicationIdAndVersionNumber` query.

app/containers/Forms/VerificationStatement.tsx
```diff
@@ -36,28 +33,62 @@ export const VerificationStatementComponent: React.FunctionComponent<Props> = ({
 }) => {
   const [isUploading, setIsUploading] = useState(false);
   const saveAttachment = async (e) => {
-    setIsUploading(true);
+    try {
+      setIsUploading(true);
+      const { environment } = relay;
+      const file = e.target.files[0];
+
+      if (file.size > 5000000) {
```
dleard, line 41: Need one more '0' here. This is currently restricting to files less than 5mb

dleard (changes requested):
I'm getting errors when uploading files that should fit the size and type requirements. Can you check the upload manually and let me know if you're able to upload any files?

dleard (approved):
:shipit: Nice Work!
Upload works as expected and restricts by file type/size properly
I can see all files uploaded across versions

PR #2206, opened 2023-04-24
Feat/2181 download attachments

tmastrom:
Card: https://github.com/bcgov/cas-ciip-portal/issues/2181


app/containers/Applications/ApplicationDetailsContainer.tsx
```diff
@@ -39,6 +48,9 @@ export const ApplicationDetailsComponent: React.FunctionComponent<Props> = ({
   const diffFromResults = review
     ? diffQuery?.old?.orderedFormResults?.edges
     : undefined;
+  const attachments = applicationRevision.attachmentsByApplicationIdAndVersionNumber
+    ? applicationRevision.attachmentsByApplicationIdAndVersionNumber.edges
+    : undefined;
```
BCerki, line 53: Could use optional chaining to shorten this, something like `const attachments = applicationRevision?.attachmentsByApplicationIdAndVersionNumber?.edges`

And then instead of checking if `attachments` is false for conditional rendering, check if its length is 0

app/containers/Applications/ApplicationDetailsContainer.tsx
```diff
@@ -156,7 +170,72 @@ export const ApplicationDetailsComponent: React.FunctionComponent<Props> = ({
...
+              >
+                {isOpen ? "+" : "-"}
+              </Button>
+            </Col>
+          </Row>
+        </Card.Header>
+        <Collapse in={!isOpen}>
+          <Card.Body className="card-body">
+            {attachments ? (
+              attachments.map((opt) => (
```
BCerki, line 197: could use `{node}` as the map argument and then you don't have to repeat `opt.` everywhere

app/tests/unit/server/middleware/attachmentDownloadRouter.test.tsx
```diff
@@ -0,0 +1,114 @@
+/**
+ * @jest-environment node
+ */
+import {
+  attachmentDownloadRouter,
+  handleDownload,
+} from "server/middleware/attachmentDownloadRouter";
+
+const performQuery = require("../../../../server/postgraphile/graphql");
```
BCerki, line 9: I ~* think you might be able to shorten the require path to just `server/postgraphile/graphql`?

app/tests/unit/server/middleware/attachmentDownloadRouter.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+const Storage = require("@google-cloud/storage");
+
+jest.mock("../../../../server/postgraphile/graphql");
```
BCerki, line 12: same here if it works

app/tests/unit/server/middleware/attachmentDownloadRouter.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const routerUnderTest = attachmentDownloadRouter;
+
+    expect(routerUnderTest.stack).toHaveLength(1);
+    expect(routerUnderTest.stack[0].route.path).toBe("/download/:attachmentId");
+    expect(attachmentDownloadRouter.stack[0].route.stack[0].handle).toBe(
+      handleDownload
+    );
+  });
+
+  it("pipes the storage client's file", async () => {
```
BCerki, line 30: This looks great! Thanks so much for figuring out the old jest test setup

app/containers/Applications/ApplicationDetailsContainer.tsx
```diff
@@ -156,7 +170,72 @@ export const ApplicationDetailsComponent: React.FunctionComponent<Props> = ({
           />
         ))}
       </div>
+      <Card
```
BCerki, line 173: This looks super nice in the app

BCerki (commented):
This looks great, review comments are all nits. I think we could additionally use a test for `VerificationStatement` to check that clicking the filename calls the router or has the right href, whatever makes more sense. There's the beginnings of an example of simulating a click in old jest here: https://github.com/bcgov/cas-ciip-portal/pull/2205/files#diff-bbd02c55400eb98b04e21e90329fba88eb8c6c035b0753fcb48e0dc503534c6cR107-R124

app/tests/unit/containers/Applications/VerificationStatement.test.tsx
```diff
@@ -21,32 +25,53 @@ describe("The Attachment Upload Component", () => {
             createdAt: "date",
           },
         },
-        {
```
tmastrom, line 24: there should only be 1 attachment

BCerki (dismissed):

app/tests/unit/containers/Applications/VerificationStatement.test.tsx
tmastrom: Thanks for letting me know, I've updated the testing to reflect this

PR #2208, opened 2023-05-01
feat: added verification statement alert

JoshLarouche:
[zenhub 2203](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-ciip-portal/2203)

### DIFF
--- a/app/containers/Forms/VerificationStatement.tsx
+++ b/app/containers/Forms/VerificationStatement.tsx
@@ -6,7 +6,7 @@ import { dateTimeFormat } from "functions/formatDates";
 import createAttachmentMutation from "mutations/application/createAttachmentMutation";
 import deleteAttachmentMutation from "mutations/application/deleteAttachment";
 import React, { useState } from "react";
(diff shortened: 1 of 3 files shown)


BCerki (commented):
Looking good! A couple general notes:
- the card description talks about a tooltip--I assume it's meant to be an Alert since that's what the dev checklist says, but you could check with the card writer/wireframe if we have one
- could add an assertion to `VerificationStatement.test.tsx` "renders the component" to check that the Alert component shows up

JoshLarouche:
> Looking good! A couple general notes:
> 
> * the card description talks about a tooltip--I assume it's meant to be an Alert since that's what the dev checklist says, but you could check with the card writer/wireframe if we have one
> * could add an assertion to `VerificationStatement.test.tsx` "renders the component" to check that the Alert component shows up

Sounds good! I'll work on these thanks!

BCerki (approved):
🤸

PR #2224, opened 2023-05-26
Feat/update gwp to ar5

dleard:


schema/revert/tables/gas_002_ar5_gwp.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert ggircs-portal:tables/gas_002_ar5_gwp from pg
+
+BEGIN;
+
+-- XXX Add DDLs here.
```
mikevespi, line 5: Makes sense to me, just want to make sure we're intentionally leaving this as a one way mutation

schema/verify/tables/gas_002_ar5_gwp.sql
```diff
@@ -0,0 +1,7 @@
+-- Verify ggircs-portal:tables/gas_002_ar5_gwp on pg
+
+BEGIN;
+
+-- XXX Add verifications here.
```
mikevespi, line 5: Intentionally empty or did it just slip through the cracks?

schema/revert/tables/gas_002_ar5_gwp.sql
dleard: Nop, not intentional. Thanks!

mikevespi (approved):
Awesome! Went in and tested for 2022 and pre 2022, everything looks right to me, calculations are using the correct values. Nice fix!

PR #2246, opened 2023-07-13
Fix: gwp values correction

dleard:


schema/deploy/database_functions/correct_gwp_values.sql
```diff
@@ -0,0 +1,113 @@
...
+  set form_result = replace(form_result::text, '"gwp": 25', '"gwp": 28')::jsonb
+  where application_id in (
+    select application_id
+      from ggircs_portal.application_revision ar
+      join ggircs_portal.application a
+      on a.id = ar.application_id
+      and a.reporting_year = 2022
+      and ar.created_at < '2023-05-30' and ar.created_at > '2023-04-01'
+  )
+  and form_id=2;
```
andrea-williams, line 25: what's `form_id`?

dleard: form_id is a foreign key to the form_json table. The form_json table contains the json schemas for each form. In this case, form_id=2 is the emissions schema. So this condition ensures we're only touching emission form data (and not fuel, production or admin data)

schema/deploy/database_functions/correct_gwp_values.sql
```diff
@@ -0,0 +1,113 @@
...
+-- CH2HF5
```
pbastia, line 86: should be `C2HF5` ?

schema/revert/data/correct_gwp_values.sql
```diff
@@ -0,0 +1,82 @@
...
+
+alter table ggircs_portal.form_result disable trigger _100_timestamps, disable trigger _immutable_form_result;
+
+  -- CH4
+update ggircs_portal.form_result
+  set form_result = replace(form_result::text, '"gwp": 28', '"gwp": 25')::jsonb
+  where application_id in (
+    select application_id
+    from ggircs_portal.application_revision
+    where created_at < '2023-05-30' and created_at > '2023-04-01'
```
pbastia, line 13: we probably want that selector on application year on the revert file (or just not implement revert?)

pbastia (changes requested):

schema/test/unit/database_functions/correct_gwp_values_test.sql
```diff
@@ -0,0 +1,195 @@
...
+          from source_types
+        )
+    select distinct((gases ->> 'gwp')::numeric)            as gwp
+    from gases
+    where application_id = 2 and version_number=1
+  $$,
+  $$
+  select distinct gwp from ggircs_portal.gwp where reporting_year_start = 2022 or gwp = 1
+  $$,
+  'Application with ID 2 was updated because it was created between before April 1st 2023 and May 30 2023'
```
andrea-williams, line 87:
```suggestion
  'Application with ID 2 was updated because it was created between April 1st 2023 and May 30 2023'
```

andrea-williams (approved):
One small suggested change re: confusing wording in test output, but not a big deal. Otherwise code LGTM

pbastia (approved):
Looks good! Thanks for doing this so quickly Dylan

PR #2249, opened 2023-08-11
Fix: annual co2e values correction

Sepehr-Sobhani:


dleard (changes requested):

dleard:
@Sepehr-Sobhani closing this PR as the issue was handled via the app & doesn't require a database migration

PR #2280, opened 2024-02-17
Use terraform chart

pbastia:


joshgamache (approved):
Fantastic! Fun to see the new chart in use!

PR #2281, opened 2024-02-21
CVEs

pbastia:
Fixing CVEs and adding action able to ignore certain advisories


joshgamache (commented):
Everything ran fine locally, and looks like all of the checks pass now!

joshgamache (approved):
Everything ran fine locally, and looks like all of the checks pass now!

